PR #99, opened 2015-01-16
Added gulp task

schorsch3000:
#### What's this PR do?
it adds the taskGulpRun(), which runs gulp-tasks.
Yes, it looks like taskrunnerception, but there are projects split in frontend and backend and grunt / gulp are quite nice for frontend-task.
#### How should this be manually tested?
with something like:
```
public function gulptest($task='default'){

        $this->taskGulpRun();
        $this->taskGulpRun($task)->color()->simple()->run();
    }
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 5.4
   - 5.5
+  - 5.6
 
 before_script:
   - composer install -n --prefer-source
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 5.4
   - 5.5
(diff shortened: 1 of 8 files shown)


schorsch3000:
the travis-ci build breaks up on the q in 
"codeception/aspect-mock": "0.5q.*"
in composer.json which i fixed locally but did not commit

DavertMik:
@schorsch3000 thanks! Especially for handling new task format!

boedah:
In Run->run() you call Base->getCommand() with a parameter, but the method has none and uses $this->task directly.

DavertMik:
@boedah thanks, fixed it

PR #101, opened 2015-01-23
Update Ssh.php

Butochnikov:
ERROR: Class 'Robo\Task\Remote\Result' not found
in /home/user/.composer/vendor/codegyre/robo/src/Task/Remote/Ssh.php:138


DavertMik:
Thanks!

PR #102, opened 2015-01-23
Update Remote.md

Butochnikov:


docs/tasks/Remote.md
```diff
@@ -96,7 +96,7 @@ $gitTask = $this->taskGitStack()
     ->checkout('master')
     ->pull();
 
-$this->taskSsh('remote.example.com')
+$this->taskSshExec('remote.example.com')
```
boedah: @DavertMik I just asked me, if we should name it `taskSsh` ;)

DavertMik:
This doc was generated from docblock of original Ssh class. Please update it instead https://github.com/Codegyre/Robo/blob/master/src/Task/Remote/Ssh.php#L16

Butochnikov:
then you need to replace here https://github.com/Codegyre/Robo/blob/master/src/Task/Remote/loadTasks.php#L19

taskSshExec -> taskSsh

PR #104, opened 2015-01-23
use CopyDir in _copyDir shortcut

### LINKED ISSUE #103: CleanDir is used in _copyDir shortcut

boedah:
Fixes #103 (now without handy-dandy GitHub function on the correct base...)


DavertMik:
Ups... Thanks!

PR #108, opened 2015-01-25
Adding default value option to the Configuration trait

TamasBarta:
With this modification the users of the `Robo\Common\Configuration` trait can now specify a default value to use, if none is set.


DavertMik:
Thanks! Looks useful

PR #110, opened 2015-01-27
ssh doc correction

Butochnikov:


boedah:
@DavertMik: just saw this and thinking `taskSsh()->exec()` looks nicer than `taskSshExec()->exec()`.

What do you think about renaming the method?
Especially since 0.5 since traits are named loadTasks and the task itself is called `Ssh`?

PR #112, opened 2015-01-28
Add property remoteDir to Ssh Task

boedah:
Much cleaner than prefixing `exec('cd /some-dir')`.
Can also be set via static configuration for all future calls:
`\Robo\Task\Remote\Ssh::configure('remoteDir', '/some-dir');`.

### DIFF
--- a/src/Task/Remote/Ssh.php
+++ b/src/Task/Remote/Ssh.php
@@ -11,11 +11,11 @@ use Robo\Task\BaseTask;
  * Runs multiple commands on a remote server.
  * Per default, commands are combined with &&, unless stopOnFail is false.
  *
- * ``` php
+ * ```php
  * <?php
  *
  * $this->taskSshExec('remote.example.com', 'user')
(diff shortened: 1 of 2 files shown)


src/Task/Remote/Ssh.php
```diff
@@ -117,6 +134,12 @@ public function getCommand()
         foreach ($this->exec as $command) {
             $commands[] = $this->receiveCommand($command);
         }
+
+        $remoteDir = $this->remoteDir ? $this->remoteDir : $this->getConfigValue('remoteDir');
+        if (!empty($remoteDir)) {
+            $this->printTaskInfo(sprintf('Working directory "%s"', $remoteDir));
```
DavertMik: `getCommand` is not supposed to print something to output, as it can be used as parameter to other task and not executed directly.

boedah:
@DavertMik: seems reasonable. Removed the `printTaskInfo`.

DavertMik:
Thanks!

PR #113, opened 2015-01-28
Bugfix/same configuration key overwrites other class

boedah:
In Configuration trait, `__CLASS__` is used to prefix the key. But:

> When used in a trait method, `__CLASS__` is the name of the class the trait is used in.

This results in `__CLASS__` always being  `Task\BaseTask`, regardless of the actual task.
This PR fixes this (and adds a test ;).


DavertMik:
I missed this point and you covered my back. Thank you!

PR #116, opened 2015-01-30
Fix _copyDir / _mirrorDir shortcuts

### LINKED ISSUE #115: Shortcuts _copyDir and _mirrorDir throw an error
They use the wrong constructor arguments for the Robo tasks.
Needs to be an array instead of two params.

Will push a PR soon.

boedah:
* commited the Cest first so you can see the old behaviour.
* the mirrorDir test is not that helpful, as it only tests copying (and not updating), but as you do not have a test for the MirrorDir task, I simply copied the copyDir test (too much copying indeed! ;)
* needed to alias `_copyDir as public shortcutCopyDir`, because else the `build` did not pick it up. Can you elaborate why (underscore)?

Fixes #115


boedah:
Hmm, the more I think about it, the shortcuts should check for an array of `$src` and use that directly for the constructors or else feed an array `$src => $dst`.

Ideally, `CopyDir` would be able to do this, but it uses the `BaseDir` constructor (which could of course also be overridden).

Any thoughts about this?

boedah:
ping @DavertMik

DavertMik:
Thanks for the fix. I didn't have time to take a look at it earlier. 
I think we can leave current syntax for shortcuts (without array support). 
Probably If user needs to do multiple copying/mirroring it is a good idea to write a task call instead of shortcut,

PR #117, opened 2015-01-30
Phpdoc / imports cleanup

boedah:
Initially only wanted to fix the return doc in `taskFilesystemStack`, but then ran the PhpStorm Inspections and cleaned the obvious (I hope so) ones.


DavertMik:
Thanks!

PR #119, opened 2015-02-07
Fail with exception so the task run exits with > 0

raphaelstolt:
Had the case where a non writeable .semver file didn't cause the whole task run to exit with an error > 0.


DavertMik:
Thanks!

PR #120, opened 2015-02-13
New methods replace() and regexReplace()

asterixcapri:
I added docblocks and refactored place() to use replace() internally.

### DIFF
--- a/src/Task/File/Write.php
+++ b/src/Task/File/Write.php
@@ -31,33 +31,97 @@ class Write extends BaseTask
         $this->filename = $filename;
     }
 
+    /**
+     * add a line
+     *
+     * @param string $line
+     *
+     * @return Write The current instance
+     */
     public function line($line)
     {
         $this->body .= $line . "\n";
         return $this;
     }
 
-    public function lines($lines)
+    /**
+     * add more lines
+     *
+     * @param array $lines
+     *
+     * @return Write The current instance
+     */
+    public function lines(array $lines)
     {
         $this->body .= implode("\n", $lines) . "\n";
         return $this;
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


src/Task/File/Write.php
```diff
@@ -31,33 +31,97 @@ public function __construct($filename)
...
+     *
+     * @param string $name
+     * @param string $val
+     *
+     * @return Write The current instance
+     */
     public function place($name, $val)
     {
-        $this->body = str_replace('{{' . $name . '}}', $val, $this->body);
+        $this->replace("{" . $name . "}", $val);
```
DavertMik: originally this was {{name}} (same as in handlebars)

DavertMik:
Thanks, I'm merging this. I will restore {{ }} pattern in `place` by myself

asterixcapri:
Thanks for merging. Regarding the braces, in revision bb57f68066580faba79ecf766ee1b6d9d36479f9 ```str_replace("{{$name}}", $val, $this->body)``` was replaced with ```str_replace("{{$name}}", $val, $this->body)```. This is wrong since the string "{{$name}}" is evaluated as "{value}".

asterixcapri:
asterix@linuxdev:~$ php -r '$name = "value"; echo "{{$name}}\n";'
{value}

boedah:
Yeah, that's why I wanted to clean up this statement with string concatenation (apart from PhpStorm choking on it...). 

Isn't there a test for that? 

My preference also is {{}}.
So this is a BC break.

boedah:
Two other cool (tiny) features I would love:

* allow arbitrary amount of spaces around the variable name 
* allow placing an array of variables

PR #121, opened 2015-02-15
Allow to set custom name of coverage file raw name.

raistlin:
As is noted on Codeception, --coverage option is allowed to set custom
file name.

I looked at source code of Codeception project, and this should work if not other --coverage-xxx is included.


DavertMik:
thanks

PR #123, opened 2015-02-19
Update robo

Narno:
Cleaner method to detect if Robo is running from Phar file or not.

### DIFF
--- a/robo
+++ b/robo
@@ -1,19 +1,16 @@
 #!/usr/bin/env php
 <?php
 /**
- * if we're running from phar load the phar autoload, else let the script 'robo' search for the autoloader
+ * if we're running from phar load the phar autoload,
+ * else let the script 'robo' search for the autoloader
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks

DavertMik:
But unfortunately it does not work this way :(

```
robo
PHP Fatal error:  Class 'Robo\Runner' not found in /usr/bin/robo on line 18
PHP Stack trace:
PHP   1. {main}() /usr/bin/robo:0

```

I reverted back this change

DavertMik:
Unfortunately this does not work this way:

```
robo
PHP Fatal error:  Class 'Robo\Runner' not found in /usr/bin/robo on line 18
PHP Stack trace:
PHP   1. {main}() /usr/bin/robo:0

```

Narno:
Arg, I'm doing more tests.

PR #126, opened 2015-02-28
Bugfix/CopyDir recursive destination check

### LINKED ISSUE #125: CopyDir destination directory issues
I have some issues with "CopyDir" task (mkdir in line 35):

1. the destination is not created recursively, so we ended up doing this before the copy dir call, which leads to the next issue
1. even though the @-operator is used, I get an `ERROR: mkdir(): File exists
in vendor/codegyre/robo/src/Task/FileSystem/CopyDir.php:35` using it in a project.
This breaks the Robo run and causes our CI-System to fail the build :/

Already added some tests and will create a PR in a moment.

Looking at the code however, I do not understand the initial need for a `CopyDir` task?
I can use the `MirrorDir` task (which deletes files on the destination) or use `FilesystemStack` mirror function directly.

Can you bring some light into this?

boedah:
Fixes #125 (see there for explanation).

Also added a test to check that files are overwritten.

### DIFF
--- a/src/Task/FileSystem/CopyDir.php
+++ b/src/Task/FileSystem/CopyDir.php
@@ -33,7 +33,7 @@ class CopyDir extends BaseDir
             throw new TaskException($this, "Cannot open source directory '" . $src . "'");
         }
         if (!is_dir($dst)) {
-            mkdir($dst);
+            mkdir($dst, 0777, true);
         }
         while (false !== ($file = readdir($dir))) {
             if (($file !== '.') && ($file !== '..')) {
--- a/src/Task/FileSystem/CopyDir.php
+++ b/src/Task/FileSystem/CopyDir.php
@@ -32,7 +32,9 @@ class CopyDir extends BaseDir
         if (false === $dir) {
             throw new TaskException($this, "Cannot open source directory '" . $src . "'");
         }
-        @mkdir($dst);
+        if (!is_dir($dst)) {
+            mkdir($dst);
+        }
         while (false !== ($file = readdir($dir))) {
             if (($file !== '.') && ($file !== '..')) {
                 $srcFile = $src . '/' . $file;
--- a/tests/_data/claypit/some/deeply/existing_file
+++ b/tests/_data/claypit/some/deeply/existing_file
@@ -1,0 +1,1 @@
+some existing file
--- a/tests/_data/claypit/some/deeply/nested/structu.re
+++ b/tests/_data/claypit/some/deeply/nested/structu.re
@@ -1,0 +1,1 @@
+Just a file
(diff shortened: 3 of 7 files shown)


src/Task/FileSystem/CopyDir.php
```diff
@@ -32,7 +32,9 @@ protected function copyDir($src, $dst)
         if (false === $dir) {
             throw new TaskException($this, "Cannot open source directory '" . $src . "'");
         }
-        @mkdir($dst);
+        if (!is_dir($dst)) {
+            mkdir($dst, 0777, true);
```
burzum: Isn't 777 a little too much? I think 775 or 755 would be better.

boedah: I did not want to change existing behavior, and 0777 is the default (http://php.net/manual/de/function.mkdir.php).

burzum: It's not like that standard php is secure at all. I think especially in the case of dealing with assets that are usually going into a *public* webroot it is critical to **not** set them to 777. This will make it very easy to add malicious files and code into files.

DavertMik: `permissions` option is not very descriptive, could you propose something better?
I agree to provide better option for that, and maybe 755 is a good idea for making it standard.

burzum: dirPermissions(), folderPermissions()? chmod()?

burzum:
I've added the permission feature in #145 :) @DavertMik let me know if you agree to the method name and I'll merge it ASAP.

boedah:
Wow, to much updates in the last 24 hours...

This PR was solely about the recursive bug in #125 ( @DavertMik : can you comment on my question there as to why the CopyDir task is even needed)?

I did not want to break compatibility with existing code.

The whole permissions thing is another one (from my point of view) and should be an own PR (so it is clearly visible).

If you look at Symfonys Filesystem `mkdir`, the default mode is also 0777, and this method is used without a chmod argument in all `copy()`, `mirror()`, etc. methods. Those methods are in turn are used by Robos FilesystemStack.

So if you really want to clean up the permissions thing, you had to make PRs to Symfony for providing the mode or implement those functions in Robo directly (which seems not good, but the CopyDir task already does it, see above).

EDIT:
Just saw on http://php.net/manual/en/function.mkdir.php:

> The mode is also modified by the current umask, which you can change using umask().

So the permissions should be alright, just did a quick test:

```bash
$ umask
0022
$ php -r 'mkdir("test1/test2", 0770, true);' && ls -la test1
drwxr-x---  2 pgasser pgasser 4096 Mar 13 10:10 test2
```

So maybe just merge this and decide on the permission cleanup in a new PR?

PR #128, opened 2015-03-03
test Task ReplaceInFile for multiple strings in file

pscheit:
I noticed that the current implementation allows to supply an array of arguments for to and from, so that IO access to files can be minimized. I think this is a good feature, but I noticed that there is no test for that.
So I created a little test and a sentence in the documentation for this behaviour.

Best regards Philipp

### DIFF
--- a/src/Task/File/Replace.php
+++ b/src/Task/File/Replace.php
@@ -23,6 +23,11 @@ use Robo\Task\BaseTask;
  *  ->regex('~^service:~')
  *  ->to('services:')
  *  ->run();
+ *
+ * $this->taskReplaceInFile('box/robo.txt')
+ *  ->from(array('##dbname##', 'robo'))
+ *  ->to(array('##dbhost## ', 'localhost'))
+ *  ->run();
  * ?>
  * ```
  *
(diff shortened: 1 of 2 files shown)


docs/tasks/File.md
```diff
@@ -39,14 +39,19 @@ $this->taskReplaceInFile('config.yml')
  ->regex('~^service:~')
  ->to('services:')
  ->run();
+
+ $this->taskReplaceInFile('box/robo.txt')
```
DavertMik: please move this example to dockblock of the class.
this file is generated from it

pscheit:
Okay, moved the example to the php-file. Unfortunately I cannot re-generate the docs (all md files are emptied)

burzum:
@pscheit is there still any kind of issue with this? What exactly do you mean by

> Unfortunately I cannot re-generate the docs

Is this broken for everyone now or just for you?

I references the wrong PR in my commit by the way. :(

burzum:
According to @DavertMik it's OK to merge this, doing it now.

pscheit:
you are right. Whats the best way to fix? seperate PR? Or should I commit into this branch?

@burzum when I generate all documentation files are emptied

Issue #135, opened 2015-03-05
Make JSqueeze configurable in \Robo\Task\Assets\Minify


burzum:
I've used the Minify task and seen that it kept the comments. I was looking for a way to configure that but couldn't find any. So I ended up overloading the method just to change the 2nd constructor arg. It would be nice if this would be configurable.

```php
class ExtendedMinify extends \Robo\Task\Assets\Minify {
	protected function getMinifiedText() {
		switch ($this->type) {
			case 'css':
				return \CssMin::minify($this->text);
				break;
			case 'js':
				$jsqueeze = new \JSqueeze();
				return $jsqueeze->squeeze(
					$this->text,
					true,  // $singleLine
					false, // $keepImportantComments
					false  // $specialVarRx
				);
				break;
		}
		return false;
	}
}
```

How about changing the Minify task this way?

```php
$this->taskExtMinify($this->webroot . 'js/vendors.js')
    ->singleLine(true)
    ->keepImportantComments(false)
    ->specialVarRx(false)
    ->run();
```

PR #136, opened 2015-03-05
Making it possible to pass options to the JS minification.

burzum:
See my proposal in #135

### DIFF
--- a/src/Task/Assets/Minify.php
+++ b/src/Task/Assets/Minify.php
@@ -35,6 +35,13 @@ class Minify extends BaseTask
     /** @var string $type css|js */
     protected $type;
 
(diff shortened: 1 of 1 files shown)


src/Task/Assets/Minify.php
```diff
@@ -157,6 +169,36 @@ protected function getMinifiedText()
     }
 
     /**
+     * Single line option for the JS minimisation.
+     *
+     * @return $this;
+     */
+    public function singleLine($singleLine)
```
DavertMik: should return `$this` for chained config. 
Same form methods below

burzum: Meh, forgot them, fixed.

PR #138, opened 2015-03-06
Update php server eamples

stevewest:
I noticed that in the docs and on the website the code examples for the php server task refer to a method "$this->taskPHPServer" that does not exist, I believe that these references should be updated to "taskServer" and that is reflected in this pull request.


disem:
Same issue on the main page of http://robo.li/

PR #139, opened 2015-03-06
phpunit: replace log-xml with log-junit 

vkunz:
Hi,

--log-xml has been replaced  with --log-junit in 3.4 and removed in 3.5.

vkunz


DavertMik:
thanks! looks like I refered to old docs

Issue #131, opened 2015-03-03
Any way to stop a background process?


chrisnharvey:
Just wondering if there is a way to force a background process to stop?

I see there is a protected stop method, but the only way I can call it publicly is by directly calling __destruct, which is a bit hacky.

Is there any way to force a background process to stop?

PR #143, opened 2015-03-10
Adding the ability to stop background processes

chrisnharvey:
This PR makes the stop() method on processes public. It also unsets the instance from the $instances static property when a process is stopped.

This addresses #131

### DIFF
--- a/src/Task/Base/Exec.php
+++ b/src/Task/Base/Exec.php
@@ -93,10 +93,14 @@ class Exec extends BaseTask implements CommandInterface, PrintedInterface
         $this->stop();
     }
 
-    protected function stop()
+    public function stop()
     {
         if ($this->background && $this->process->isRunning()) {
             $this->process->stop();
+
+            $key = array_search($this, static::$instances);
+            unset(self::$instances[$key]);
+
             $this->printTaskInfo("stopped <info>".$this->getCommand()."</info>");
         }
     }


burzum:
:+1:

DavertMik:
I think this should be part of `Result` object and not a task.
Your fix is nice and simple, however not a very consistent. I hope to come with a better solution soon

chrisnharvey:
Great! I'm happy to implement any solution if you give me an idea of how you'd like it to work.

Would we not still need to have the stop method public so it can be called from the Result object?

DavertMik:
@chrisnharvey probably yes, stop could be done public. But there should be created a new Result object only for Exec task. It will contain method `stopBackgroundProcess`.

burzum:
@chrisnharvey are you going to move that to the result object?

chrisnharvey:
Yes, I'm going to work on this sometime this week. My MacBook Pro has died and I only got a new one today. So after I get everything setup, I'll get to work on this.

chrisnharvey:
@DavertMik Should I create a new Result class just for the Exec Task? I.e. ExecResult.php?

Or should I just add a ```stopBackgroundProcess``` method to the current Result class?

DavertMik:
> Should I create a new Result class just for the Exec Task? I.e. ExecResult.php?

Yes

tikolakin:
Hi there
when i will be released?
Thanks.

tikolakin:
Will reference to this 
https://github.com/Codegyre/Robo/issues/274

greg-1-anderson:
Closed due to lack of activity.

PR #145, opened 2015-03-12
Bugfix/copy dir recursive destination check

burzum:
This should finish https://github.com/Codegyre/Robo/pull/126

### DIFF
--- a/src/Task/FileSystem/CopyDir.php
+++ b/src/Task/FileSystem/CopyDir.php
@@ -58,7 +58,6 @@ class CopyDir extends BaseDir
         if (false === $dir) {
             throw new TaskException($this, "Cannot open source directory '" . $src . "'");
         }
-        @mkdir($dst, $this->chmod);
         if (!is_dir($dst)) {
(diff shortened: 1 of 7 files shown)


src/Task/FileSystem/CopyDir.php
```diff
@@ -26,13 +29,39 @@ public function run()
...
+     * @return void
+     */
     protected function copyDir($src, $dst)
     {
         $dir = @opendir($src);
         if (false === $dir) {
             throw new TaskException($this, "Cannot open source directory '" . $src . "'");
         }
-        @mkdir($dst);
+        @mkdir($dst, $this->chmod);
```
DavertMik: this should be removed

burzum: Why? It will set the default permission there as well.

boedah: This line was the cause of the error in #125 ...

Please see https://github.com/Codegyre/Robo/pull/126#issuecomment-78880801

Issue #127, opened 2015-03-02
JSqueeze 2.0


nesl247:
JSqueeze has been updated to 2.0.

DavertMik:
Does it have api changes? 
If not it can be updated in `composer.json`, othwise we need to update the Minify task for it

nesl247:
I have nothing to do with the project so I am not 100% sure. I did try changing my composer.json, and found out the namespace changed.

burzum:
@nesl247 would you mind to test it? :)

burzum:
Fixed.

PR #146, opened 2015-03-12
#127 BC fix to support Jsqueeze 1.x and 2.x

burzum:


DavertMik:
I wonder if there are no API changes other than class names between versions?

PR #149, opened 2015-03-16
fix PR #128 document replaceInFile with better example

pscheit:
I switched the arguments
and I added the array type to the arguments

I regenerated the docu there is a "non related" change in the bottom, is this correct, too?

### DIFF
--- a/docs/tasks/File.md
+++ b/docs/tasks/File.md
@@ -39,12 +39,17 @@ $this->taskReplaceInFile('config.yml')
  ->regex('~^service:~')
  ->to('services:')
  ->run();
+
+$this->taskReplaceInFile('box/robo.txt')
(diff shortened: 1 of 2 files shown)


docs/tasks/File.md
```diff
@@ -68,7 +73,7 @@ $this->taskWriteToFile('blogpost.md')
 * `lines(array $lines)`  add more lines
 * `text($text)`  add a text
 * `textFromFile($filename)`  add a text from a file
-* `place($name, $val)`  substitute a placeholder with value, placeholder must be enclosed by {{}}
+* `place($name, $val)`  substitute a placeholder with value, placeholder must be enclosed by `{}`
```
pscheit: this came in, when i regenerated the docs, is this correct?

DavertMik:
Thanks!

PR #152, opened 2015-03-23
Add a line space command and a changed the format of the question

andrewtarry:
There was some odd formatting when we tried to start using robo in our project. The question "Should I create a RoboFile" was pushing it's blue background over 2 lines and putting a question mark at the start. I don't know if that was intentional but it does appear a little strange when you first pick up the project. This pull request will change the format so the blue is only on the question and the question mark is only at the end.

In addition there is a new space command that will create an empty line. It was just something we found useful to make it easier to read the output.


burzum:
>In addition there is a new space command that will create an empty line. It was just something we found useful to make it easier to read the output.

I'm OK with the change but why do you call it `space()`, which is just wrong, instead of `newLine()`? Also I think it could be nice if you could pass an optional argument of how many lines you want to print: `newLine($lines = 1);`

DavertMik:
Could you attach screenshots of how it was and how it become, because I don't see an issue here. 
Yes, the "?" was intentional from start...

burzum:
@andrewtarry any response on this please? :)

greg-1-anderson:
Closed due to lack of activity.

PR #154, opened 2015-03-23
use custom error handler to correctly handle errors that were suppressed...

### LINKED ISSUE #151: Robo reports error even if suppressed
If php encounters an error that is suppressed via the @ operator, Robo still reports it as an error after the task execution.

```php
{
    public function testError()
    {
        @unlink('/not/existing/file');
    }
}
```

I don't know if there is another solution than checking the triggering line for an @ and not report the error if this is the case.
It may not be a severe issue (i just encountered it in a doctrine command), but maybe it's worth thinking about.

aaukt:
... with the @-operator or by disabling error reporting

fixes #151

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -49,6 +49,7 @@ class Runner
     public function execute($input = null)
     {
         register_shutdown_function(array($this, 'shutdown'));
+        set_error_handler(array($this, 'handleError'));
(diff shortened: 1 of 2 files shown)


burzum:
:+1:

DavertMik:
thanks!

PR #155, opened 2015-03-26
Fix error in array name at documentation

javigomez:
Fixes a wrong name of a variable in the documentation. Is wrong because in previous line is called `$opts` and in the present line is called `$opt` instead


burzum:
Thank you!

PR #156, opened 2015-03-28
Adding a less task that features two less compilers.

burzum:


src/Task/Assets/Less.php
```diff
@@ -0,0 +1,179 @@
...
+ * inject the name there.
+ */
+class Less extends BaseTask
+{
+    /**
+     * The less compiler to use.
+     *
+     * @var string
+     */
+    protected $lessCompiler = 'less';
```
DavertMik: Could you match available compiler by `class_exists`? That would simplify life

burzum: I think a mapping by name is better because it allows you to extend the less task and implement your own parser method. For example you want to do something special with one of the compilers. Right now we just use the basics of them but I've seen they provide more functionality. If you only use the class name it wont give you the ability to implement any custom logic around your compiler.

Issue #129, opened 2015-03-03
Testsuite does not run without failures on windows


pscheit:
Hi, 

thank you for robot.li. This is by far the best task-library I have used.

I noticed when submitting another PR that the tests do not run without failures completely on a windows system. There are 3 failing tests and I think they are all easy to fix, when one might use a condition if the test is running on windows.
I can fix those tests or skip those tests for windows, Are you willing to accept a PR for that?

Best regards
Philipp

DavertMik:
Sorry for taking some time
@pscheit thanks for your proposal. That would be awesome if you could get those tests fixed on windows!
Please send a PR

burzum:
That's the output I get when running the tests:

```
Codeception PHP Testing Framework v2.0.9
Powered by PHPUnit 4.4.2 by Sebastian Bergmann.

[1mCli Tests (13) [22m--------------------------------------------------------------------------------------------------------------------
Trying to [35;1mto exec ls command[39;22m (ExecCest::toExecLsCommand)                                                                      Ok
Trying to [35;1mto create dir[39;22m (FileSystemStackCest::toCreateDir)                                                                    Ok
Trying to [35;1mto delete file[39;22m (FileSystemStackCest::toDeleteFile)                                                                  Ok
Trying to [35;1mcopy dir with _copyDir shortcut[39;22m (ShortcutsCest::useTheCopyDirShortcut)                                              Ok
Trying to [35;1mmirror dir with _mirrorDir shortcut[39;22m (ShortcutsCest::useTheMirrorDirShortcut)                                        Ok
Trying to [35;1mwrite lines with WriteToFile task[39;22m (WriteFileCest::writeFewLines)                                                    Ok
Trying to [35;1mappend to file[39;22m (WriteFileCest::appendToFile)                                                                        Ok
Trying to [35;1minsert file[39;22m (WriteFileCest::insertFile)                                                                             Ok
Trying to [35;1mreplace in file[39;22m (WriteFileCest::replaceInFile)                                                                      Ok
Trying to [35;1mclean dir with DeleteDirTask[39;22m (CleanDirCept)                                                                         Ok
Trying to [35;1mconcat files using Concat Task[39;22m (ConcatCept)                                                                         Ok
Trying to [35;1mcopy dir with CopyDir task[39;22m (CopyDirCept)                                                                            Ok
Trying to [35;1mdelete dir with DeleteDirTask[39;22m (DeleteDirCept)                                                                       Ok
[...]

PR #157, opened 2015-03-31
Fix unit-tests on windows

pscheit:
This should fix issue #129.

I found out that one testcase in the gulp test was testing a part where escapeshellarg from php is used for escaping the name of the task. In windows it's a bit complicated to escape double quotes on the command line (been there ..) So I used the symfony utility which behaves correctly on windows and unix escaping double quotes in commands.

The other fixes were just prepending "call" to batch files and changing from single-quotes to double-quotes

best regards
Philipp

### DIFF
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -71,6 +71,6 @@ trait ExecCommand
         }
         $this->stopTimer();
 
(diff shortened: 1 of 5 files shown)


DavertMik:
thanks!

PR #158, opened 2015-04-08
PhpServer: Wrap `php` command with `exec`

fabfuel:
Wrap `php` command with `exec` to prevent detached PHP server process on e.g. Debian after stopping the background task.

On Debian, the PHP server process currently does not get killed.
The server gets startet correctly:
```
vagrant   4233  pts/0    Ss   09:05   0:00  |       \_ -bash
vagrant   5936  pts/0    S+   10:43   0:00  |           \_ php bin/robo server
vagrant   5937  pts/0    S+   10:43   0:00  |               \_ sh -c php -S 0.0.0.0:8000 -t public
vagrant   5938  pts/0    S+   10:43   0:00  |                   \_ php -S 0.0.0.0:8000 -t public
```

But after finishing the task and stopping the process, the actual PHP server process is still running:
```
vagrant   5938  pts/0    S    10:43   0:00 php -S 0.0.0.0:8000 -t public
```

In this example, only the PID 5937 was killed, not the server itself 5938.
Wrapping the `php` call with `exec` fixes that issue.

Best
Fabian

### DIFF
--- a/src/Task/Development/PhpServer.php
+++ b/src/Task/Development/PhpServer.php
@@ -29,7 +29,7 @@ class PhpServer extends Exec
 {
     protected $port;
     protected $host = '127.0.0.1';
-    protected $command = 'php -S %s:%d ';
+    protected $command = 'exec php -S %s:%d ';
 
     public function __construct($port)
     {
--- a/src/Task/Development/PhpServer.php
+++ b/src/Task/Development/PhpServer.php
@@ -29,11 +29,15 @@ class PhpServer extends Exec
 {
     protected $port;
     protected $host = '127.0.0.1';
-    protected $command = 'exec php -S %s:%d ';
+    protected $command = 'php -S %s:%d ';
 
     public function __construct($port)
     {
         $this->port = $port;
+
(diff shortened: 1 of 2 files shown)


DavertMik:
Will this work on other systems than Debian?

burzum:
@fabfuel this change will break the PhpServer on windows.

fabfuel:
You're right, I forgot about Windows... `exec` works on all Unix based operating systems, including OS X. I did not test Robo on Windows, but if it currently works correct without the `exec` wrapper, we can skip it.

I now check PHP_OS and only add `exec` on "linux" environments.

By the way, Symfony does it the same way. As far as I see, they are not checking the OS, but I'm not familiar with Symfony, so maybe there's a magic way how they resolve that:
https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Command/ServerStartCommand.php#L220

Best
Fabian

fabfuel:
On OS X it also works without the `exec` wrapper, that's why I check especially for "linux" environments.

DavertMik:
Great. Thank you!

PR #166, opened 2015-04-24
Add option for show codeception output in compact style

miramir:


DavertMik:
I'm confused either...
I think `--report` option is not as widely used to include it. You can always pass it as 

```
 taskCodecept()
  ->...
 ->opt('--report')
```

PR #168, opened 2015-05-08
allow wildcard path on taskMinify

kodok-buncit:
this PR will allow doing task like this.

```
$this->taskMinify( 'public/css/*.css' )
     	->run();
```

### DIFF
--- a/src/Task/Assets/Minify.php
+++ b/src/Task/Assets/Minify.php
@@ -58,7 +58,7 @@ class Minify extends BaseTask
             $this->files[] = $input;
             return $this;
         }
-        else
+        elseif( False !== strpos($input, '*') )
         {
             // by wildcard
             $finder = new Finder();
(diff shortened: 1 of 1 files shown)


src/Task/Assets/Minify.php
```diff
@@ -49,8 +53,42 @@ class Minify extends BaseTask
...
+            }
+            
+            foreach ($iterator as $file) {
+                $this->files[] = $file->getRealpath();
+            }
+
+            return $this;
         }
 
         return $this->fromText($input);
```
boedah: this is never reached and breaks existing behavior...

kodok-buncit:
@boedah  how about new commit?

burzum:
I'm not against the feature but it seems useless to me because usually the order of the files matters. I ran into the same issue when I tried to process an angular application that has a lot sub-folders.

boedah:
agree.
Also seems the Result object of the `_run()` method is never used...

kodok-buncit:
allright it seems the feature only usefull for me.
closed

burzum:
@kodok-buncit you can get this done much more easy with http://php.net/manual/en/function.glob.php than adding all of that code. Check the examples and comments there.

PR #169, opened 2015-05-14
Fix formatting of taskExec() methods

bcremer:


DavertMik:
This doc was generated, probably issue should be fixed in Robofile :(

PR #173, opened 2015-06-01
added "git tag" task

SebSept:
I need it. Maybe someone else :)

### DIFF
--- a/src/Task/Vcs/GitStack.php
+++ b/src/Task/Vcs/GitStack.php
@@ -12,6 +12,7 @@ use Robo\Task\CommandStack;
  *  ->stopOnFail()
  *  ->add('-A')
  *  ->commit('adding everything')
+ *  ->tag('0.6.0')
  *  ->push('origin','master')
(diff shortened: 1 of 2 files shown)


DavertMik:
@boedah looks like this should be merged. No objections?

DavertMik:
Btw, here is the use case. Recenlty I implemented a script that mirrors Codeception into another repo with reducing some dependencies. 

https://github.com/Codeception/Codeception/blob/master/RoboFile.php#L632

boedah:
Yes, but it would be great, if you could add support for (optional) annotated tags.
This should be easy to do, meaning:

* adding  `-a -m %s`, where the message is escaped
** using http://api.symfony.com/2.7/Symfony/Component/Process/ProcessUtils.html#method_escapeArgument
** this could be done for many other tasks see https://github.com/Mihailoff/Robo/commit/44d2f8fe601a812d06c94f76b6c7ac444d2ade72#commitcomment-10860178
** also see #157

PR #178, opened 2015-06-16
Fix wrong option to skip groups

javigomez:
There was a mistake in the skip group option:

screen shot 2015-06-16 at 17 08 06

As you see, the real name of the option is:

screen shot 2015-06-16 at 17 09 31

This pull fixes the issue


javigomez:
Thanks ^_^

PR #184, opened 2015-07-06
Fixing deletion of all files other than .gitignore and .gitkeep

gitarnaud:


DavertMik:
I'm can't get why we should keep `.gitattributes` or any other git-related file...

gitarnaud:
I agree with you that a task named CleanDir should not keep any file in the directory. In this pull request, I just extended the current behavior, making it more generic.

Maybe the emptyDir($path) function should take an optional second argument which would be a regexp array of files to exclude from deletion.

greg-1-anderson:
Behaves as implemented.

PR #185, opened 2015-07-08
continue on fail for command stack

jaminmelville:
Fix for the following so COMMAND2 runs if COMMAND1 fails.

$this->taskExecStack()
  ->stopOnFail(FALSE)
  ->exec("COMMAND1")
  ->exec("COMMAND2")
  ->run();

currently executes "COMMAND1 && COMMAND2" which if COMMAND1 fails COMMAND2 is never executed.

This pull request changes it so executes "COMMAND1 ; COMMAND2" so COMMAND2 is executed even if COMMAND1 fails.

### DIFF
--- a/src/Task/Base/ExecStack.php
+++ b/src/Task/Base/ExecStack.php
@@ -24,10 +24,4 @@ use Robo\Task\Base;
  */
 class ExecStack extends CommandStack
 {
-
-  public function getCommand()
-  {
-    return implode($this->stopOnFail ? ' && ' : ' ; ', $this->exec);
-  }
-
 }
--- a/src/Task/Base/ExecStack.php
+++ b/src/Task/Base/ExecStack.php
@@ -24,4 +24,10 @@ use Robo\Task\Base;
  */
 class ExecStack extends CommandStack
 {
+
+  public function getCommand()
+  {
+    return implode($this->stopOnFail ? ' && ' : ' ; ', $this->exec);
+  }
+
 }
--- a/src/Task/CommandStack.php
+++ b/src/Task/CommandStack.php
@@ -21,7 +21,7 @@ abstract class CommandStack extends BaseTask implements CommandInterface, Printe
 
     public function getCommand()
     {
-        return implode(' && ', $this->exec);
+      return implode($this->stopOnFail ? ' && ' : ' ; ', $this->exec);
     }
 
     public function exec($command)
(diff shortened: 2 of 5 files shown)


boedah:
(Sorry, this got long ;) but am also interested in your thoughts @DavertMik)

Ah yes, also stumbled upon the various implications of `stopOnFail`.
Now trying to sum them up:

**CommandStack (which is the abstract class of ExecStack)**
  * `stopOnFail === true` does execute each command (without using `getCommand()`), and returns the failed Result instead of executing the remaining commands (https://github.com/Codegyre/Robo/blob/master/src/Task/CommandStack.php#L61)
  * `stopOnFail === false` (default) calls `getCommand()` (which currently joins via `&&`) and therefore also does not execute the remaining commands, of one fails (https://github.com/Codegyre/Robo/blob/master/src/Task/CommandStack.php#L53)

Both return a

**Result object**

Static `stopOnFail` (https://github.com/Codegyre/Robo/blob/master/src/Tasks.php#L42) is used to _exit the PHP process_ if the return code of a command indicates an error (https://github.com/Codegyre/Robo/blob/master/src/Result.php#L114).

**Ssh**

I made this task respect the stopOnFail like in this PR (https://github.com/Codegyre/Robo/blob/master/src/Task/Remote/Ssh.php#L142)

**ParallelExec**

Also does a `implode(' && ', $this->processes);`, however `getCommand()` does not seem to be used because (of course) it starts own processes in the background (https://github.com/Codegyre/Robo/blob/master/src/Task/Base/ParallelExec.php#L65)

Maybe it should not implement `CommandInterface` at all, since the only method in there is `getCommand()`.

**General Thoughts**

Does it even make sense to use `getCommand()` in `run()`?
As looping over each command and returning the failing `Result` leads to better understanding, *which* command actually fails.
Else I get one `Result` object, which tells me something like: `command "command1 && command2 && command3 && command4"` failed...

**This PR**

Regarding this PR, some thoughts:

* Windows break: `&&` is the same but a `;` would be a single `&` (https://technet.microsoft.com/en-us/library/bb490954.aspx#EEAA)
* breaks existing behaviour (`stopOnFail` should be set to `true`)
* why not do it directly in `CommandStack`?
  * the only ones extending it are `ExecStack`, `GitStack` and `SvnStack`

Woo, think that is all.

greg-1-anderson:
Re-open if you want to return to this issue. Underlying code may have changed.

Issue #181, opened 2015-06-24
Include symfony/yaml package to allow yaml parsing


caseyfw:
It would be nice if the Robo phar included the symfony/yaml package so tasks could parse a yaml file for config items.

This is useful for build/deployment tasks that require a bunch of different configuration that can be transported in a yaml file.

caseyfw:
We're using Robo to build a Drupal project, and part of our build reads a bunch of config from a yaml file and sets the corresponding value in Drupal using `drush cset`.

Our temporary solution until the symfony/yaml package is included in the robo phar relies on a `composer install` having already been done in the project directory. This way, those extra packages can be included for autoloading:

```php
public function buildConfigure() {
  // Assuming deps have been pulled, load them.
  if (file_exists(__DIR__ . '/vendor/autoload.php')) {
    require_once __DIR__ . '/vendor/autoload.php';
  }
  if (file_exists($this->config_file)) {
    $raw = file_get_contents($this->config_file);
    $yaml = Yaml::parse($raw, TRUE, FALSE);
    $task_stack = $this->taskExecStack();
    foreach ($yaml['drupal_config'] as $drupal_config_name => $drupal_config_items) {
      foreach ($drupal_config_items as $key => $value) {
        $task_stack->exec($this->drush_cmd . ' cset ' . $drupal_config_name .
          ' ' . $key . ' "' . $value . '" -y');
      }
    }
    $task_stack->printed(FALSE)->stopOnFail(TRUE)->run();
    $this->say('Site configuration imported');
  }
  else {
    $this->say($this->config_file . ' file not found');
  }
}
```

boedah:
From my opinion projects should include additional dependencies themselves as Robo should not include dependencies it is not using itself.

caseyfw:
What do you think about a getConfig() task or something along those lines? We could even get more extreme and use the symfony/config package...

boedah:
Talked about this with @DavertMik, and I think he is not a big fan of this (see comments from https://github.com/Codegyre/Robo/issues/41#issuecomment-46954307 on).
We also reverted to using PHP instead of YAML for config files.

Nevertheless, I would find a config system similar to ant/phing, where you can use config properties of a global config file in local config files easily appealing...

caseyfw:
I can see the logic in this. YAML does add needless overhead when we can achieve what we want with plain old php. I guess I was perhaps a bit too "all aboard the YAML train".

c33s:
[...]

PR #186, opened 2015-07-09
Add symfony/yaml ~2.5 to composer packages

caseyfw:
This enables the parsing of YAML files in Robo - super handy for handling complex configuration during builds. See issue #181.

---

We're using Robo to build a Drupal project, and part of our build reads a bunch of config from a yaml file and sets the corresponding value in Drupal using `drush cset`.

Our temporary solution until the symfony/yaml package is included in the robo phar relies on a `composer install` having already been done in the project directory. This way, those extra packages can be included for autoloading:

```php
use \Symfony\Component\Yaml\Yaml;

// ...

public function buildConfigure() {
  // Assuming deps have been pulled, load them.
  if (file_exists(__DIR__ . '/vendor/autoload.php')) {
    require_once __DIR__ . '/vendor/autoload.php';
  }
  if (file_exists($this->config_file)) {
    $raw = file_get_contents($this->config_file);
    $yaml = Yaml::parse($raw, TRUE, FALSE);
    $task_stack = $this->taskExecStack();
    foreach ($yaml['drupal_config'] as $drupal_config_name => $drupal_config_items) {
      foreach ($drupal_config_items as $key => $value) {
        $task_stack->exec($this->drush_cmd . ' cset ' . $drupal_config_name .
          ' ' . $key . ' "' . $value . '" -y');
      }
    }
    $task_stack->printed(FALSE)->stopOnFail(TRUE)->run();
    $this->say('Site configuration imported');
  }
  else {
    $this->say($this->config_file . ' file not found');
  }
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "symfony/console": "~2.5",
         "symfony/process": "~2.5",
         "symfony/filesystem": "~2.5",
+        "symfony/yaml": "~2.5",
         "henrikbjorn/lurker": "1.0.*@dev"
     },
     "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "51655a3db511e0655263959527ede2d7",
+    "hash": "0a8092ad5acb2738af74cf38a252287b",
     "packages": [
         {
             "name": "henrikbjorn/lurker",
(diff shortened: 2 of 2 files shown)


DavertMik:
Hi, I'd rather not include Yaml as it is not used by Robo itself. But you can add yaml to your fork updated and rebuild `robo.phar` by yourself. 

```
robo phar:build
```

caseyfw:
I can see the logic in this - we'll move our config to a plain .php file as @boedah suggested. Putting some thought into it, I can't actually see a reason why YAML is preferable to config in PHP files, and adding yaml ability to Robo adds 100kB to the phar.

PR #188, opened 2015-07-10
Fixed capitalisation of task info output

caseyfw:
The task info outputs couldn't seem to make up their mind if they should be capitalised or not. Symfony's console often uses the paradigm where every first-column item is capitalised, and indented items can be whatever makes sense.


DavertMik:
Yeah, this is a good suggestion. Thanks!

PR #193, opened 2015-07-31
Less Task Bugfixes

### LINKED ISSUE #192: Less Task: Passing parser options to less.php compiler
In <code>Robo\Task\Assets\Less::less()</code>: <code>$this->compilerOptions</code> should be passed to <code>Less_Parser</code> Constructor.

pr0nbaer:
Fixes #192 and #194

### DIFF
--- a/src/Task/Assets/Less.php
+++ b/src/Task/Assets/Less.php
@@ -112,7 +112,7 @@ class Less extends BaseTask
     protected function compile($file)
     {
         if (is_callable($this->lessCompiler)) {
-            return $this->lessCompiler($file, $this->compilerOptions);
+            return call_user_func($this->lessCompiler, $file, $this->compilerOptions);
         }
         if (method_exists($this, $this->lessCompiler)) {
             return $this->{$this->lessCompiler}($file);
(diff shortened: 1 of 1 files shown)


src/Task/Assets/Less.php
```diff
@@ -112,7 +112,7 @@ public function compiler($compiler, array $options = [])
     protected function compile($file)
     {
         if (is_callable($this->lessCompiler)) {
-            return $this->lessCompiler($file, $this->compilerOptions);
+            return call_user_func($this->lessCompiler, $file, $this->compilerOptions);
         }
         if (method_exists($this, $this->lessCompiler)) {
             return $this->{$this->lessCompiler}($file);
```
MAXakaWIZARD: $this->compilerOptions also should be passed here (in case with `lessphp`):
```php
return $this->{$this->lessCompiler}($file, $this->compilerOptions);
```

logbon72:
Hi,

Any update on when this PR will be merged into the main repo ?

DavertMik:
@logbon72 what would you say about @MAXakaWIZARD comment? Do you think it should be taken into account before merge?

logbon72:
The comment looks like an enhancement, this PR is a bug fix and the issue is causing stuffs to break, I think users can write callbacks as a workaround for @MAXakaWIZARD comment. 
I noticed import statements were not working because the file contents were passed to the compilers in place of the file path, which was why I wrote a callback in the first place.

DavertMik:
Ok, thanks. Merging it

PR #196, opened 2015-08-06
Add missing packages error reporting to Minify and Less tasks

MAXakaWIZARD:
If required package is not installed, detailed detailed error message will be shown:
```bash
Class lessc not found. Please install leafo/lessphp Composer package
```

### DIFF
--- a/src/Task/Assets/Less.php
+++ b/src/Task/Assets/Less.php
@@ -128,6 +128,10 @@ class Less extends BaseTask
      */
     protected function lessphp($file)
     {
+        if (!class_exists('\lessc')) {
+            return Result::errorMissingPackage($this, 'lessc', 'leafo/lessphp');
+        }
+
         $lessCode = file_get_contents($file);
         $less = new \lessc();
         return $less->compile($lessCode);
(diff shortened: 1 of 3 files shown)


src/Task/Assets/Less.php
```diff
@@ -141,6 +145,10 @@ protected function lessphp($file)
      */
     protected function less($file)
     {
+        if (!class_exists('\Less_Parser')) {
+            return Result::error($this, 'Less_Parser not found. Please install oyejorge/less.php composer package');
```
DavertMik: why not `Result::errorMissingPackage` then?

src/Task/Assets/Less.php
```diff
@@ -128,6 +128,10 @@ protected function compile($file)
      */
     protected function lessphp($file)
     {
+        if (!class_exists('\lessc')) {
+            return Result::error($this, 'lessc not found. Please install leafo/lessphp composer package');
```
DavertMik: why not Result::errorMissingPackage then?

MAXakaWIZARD: Ah, sorry, my bad. Forgot to add it there (it was there in my test project)

MAXakaWIZARD:
@DavertMik fixed

DavertMik:
Thanks!

PR #203, opened 2015-08-28
Fix Replace task always writing to file

gabor-udvari:
The Replace task is always writing to file regardless whether any changes were made or not. This can bring the taskrunner into an inifinite loop if a replaced file is being watched.


DavertMik:
thanks

PR #206, opened 2015-08-30
Add ability to set import dir for less compilers

MAXakaWIZARD:
Given:
```
//styles.less
@import "mixins.less"
```
```php
//RoboFile.php
$this->taskLess(['css/styles.less' => 'css/styles.css'])
    ->compiler('lessphp')
    ->run();
```
This will fail since `lessphp` is unable to find imported file even if it is in same dir as `styles.css`.
This PR adds ability to specify `importDir` option for `lessphp` (string or array):
```php
$this->taskLess(['css/styles.less' => 'css/styles.css'])
    ->compiler('lessphp', ['importDir' => 'css'])
    ->run();
```
or
```php
$this->taskLess(['css/styles.less' => 'css/styles.css'])
    ->compiler('lessphp', ['importDir' => ['css', 'less']])
    ->run();
```

### DIFF
--- a/src/Task/Assets/Less.php
+++ b/src/Task/Assets/Less.php
@@ -23,6 +23,18 @@ use Robo\Task\BaseTask;
  * "oyejorge/less.php": "~1.5"
  * ```
  *
+ * Specify directory (string or array) for less imports lookup:
+ * ```php
+ * <?php
+ * $this->taskLess([
+ *     'less/default.less' => 'css/default.css'
+ * ])
+ * ->importDir('less')
+ * ->compiler('lessphp')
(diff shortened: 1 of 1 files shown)


DavertMik:
Is there any reason it can't be done in this manner?

```php
<?php
$this->taskLess(['css/styles.less' => 'css/styles.css'])
    ->importDir('style')
    ->compiler('lessphp')
    ->run();

```

MAXakaWIZARD:
@DavertMik I've rewritten it, added `importDir` method and also implemented support for second less compiler also.

DavertMik:
thanks

PR #207, opened 2015-09-02
Refactoring in Less and Scss tasks

### LINKED ISSUE #205: Create common abstract class for Less task and SCSS task
I would like to open a discussion about creating a common abstract class for the Less task and SCSS task and the methods the abstract class should have.

When I wrote the SCSS task (merged in #200) I copied the Less task so there is a lot of repeated code. This makes a double effort with bugfixes (#193 is now only applied to the Less task) and inconsistency with features (see [MAXakaWIZARD's comment](https://github.com/Codegyre/Robo/pull/193#discussion_r36404723) for implementing importDir)

The project [phpgearbox/asset](https://github.com/phpgearbox/asset) does the same thing, both Less and Scss extending the CSS class, but I would like more public methods, like addImportPath, setImportPaths and setFormatter which [leafo/scssphp](https://github.com/leafo/scssphp) has.

MAXakaWIZARD:
Extract common functionality from Scss and Less classes to CssPreprocessor abstract class.
Fixes #205.

### DIFF
--- a/src/Task/Assets/CssPreprocessor.php
+++ b/src/Task/Assets/CssPreprocessor.php
@@ -42,11 +42,6 @@ abstract class CssPreprocessor extends BaseTask
     public function __construct(array $input)
     {
         $this->files = $input;
-        foreach ($this->files as $in => $out) {
-            if (!file_exists($in)) {
-                return Result::error($this, 'File %s not found.', $in);
-            }
-        }
 
         $this->setDefaultCompiler();
     }
@@ -128,17 +123,6 @@ abstract class CssPreprocessor extends BaseTask
      */
     public function compiler($compiler, array $options = [])
     {
-        if (!in_array($compiler, $this->compilers) && !is_callable($compiler)) {
-            $this->say(
-                sprintf(
-                    'Invalid ' . static::FORMAT_NAME . ' compiler %s!',
-                    $compiler
-                )
-            );
-
-            return false;
-        }
-
         $this->compiler = $compiler;
         $this->compilerOptions = array_merge($this->compilerOptions, $options);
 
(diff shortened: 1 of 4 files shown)


src/Task/Assets/CssPreprocessor.php
```diff
@@ -0,0 +1,205 @@
...
+     * Constructor. Accepts array of file paths.
+     *
+     * @param array $input
+     */
+    public function __construct(array $input)
+    {
+        $this->files = $input;
+        foreach ($this->files as $in => $out) {
+            if (!file_exists($in)) {
+                return Result::error($this, 'File %s not found.', $in);
```
DavertMik: it is better to throw errors only on `run`. This would make sense in case we add options to append files into configuration

src/Task/Assets/CssPreprocessor.php
```diff
@@ -0,0 +1,205 @@
...
+    {
+        if (!in_array($compiler, $this->compilers) && !is_callable($compiler)) {
+            $this->say(
+                sprintf(
+                    'Invalid ' . static::FORMAT_NAME . ' compiler %s!',
+                    $compiler
+                )
+            );
+
+            return false;
```
DavertMik: ok, this breaks chained call and goes to fatal error. 
as it was mentioned above, any errors should be thrown at `run` method.
so compiler check should be there as well

src/Task/Assets/CssPreprocessor.php
```diff
@@ -0,0 +1,205 @@
...
+    protected function setDefaultCompiler()
+    {
+        if (isset($this->compilers[0])) {
+            //set first compiler as default
+            $this->compiler = $this->compilers[0];
+        }
+    }
+
+    /**
+     * Sets import directories
```
DavertMik: should be mentioned that it is an alias for `setImportPaths`

MAXakaWIZARD:
@DavertMik I've fixed all you remarks, please review once more.

DavertMik:
I like that. I think we can merge it. @gabor-udvari what you think?

gabor-udvari:
I think it can be merged too, great job!

DavertMik:
thanks

PR #212, opened 2015-09-09
[Proposal] Exclude packages tests from phar build

Benoth:
It seems that the packages tests are not required to run Robo, right ? We may so exlude them (reducing final phar size ~40%).


boedah:
:+1:

DavertMik:
:+1:

simondubois:
:+1:

PR #215, opened 2015-09-13
Add new task FlattenDir

gabor-udvari:
The task searches for files in a nested directory structure and copies them to a target directory with or without the parent directories. The task was inspired by [gulp-flatten](https://www.npmjs.com/package/gulp-flatten).

Tests and documentation included.


DavertMik:
Thanks! That's really nice

PR #218, opened 2015-09-23
Make Robo not die if no RoboFile available.

### LINKED ISSUE #130: robo --version not possible if RoboFile isnt existing
It would be cool, to show the version even if no RoboFile.php is present in the current working directory.

holyspecter:
Also I'd propose to make separate command for initializing RoboFile to avoid annoying prompt if one doesn't exist. Would it be useful?
This PR fixes #130.


DavertMik:
Thank you!

PR #226, opened 2015-10-03
Fix an example in documentation.

IRuslan:
Hi!

I've noted that example in docs is not relevant, there is no such trait like use Robo\Task\Exec and seems use Robo\Task\Base\loadShortcuts is enough.


DavertMik:
thnks

Issue #216, opened 2015-09-14
Improved and consistent glob support


dlancea:
Tasks seem to be updated piecemeal to include php's glob() function.

It would be helpful to:
1. Use a better glob function such as: https://github.com/webmozart/glob . This library supports an ant-like syntax, similar to what's used in gulp/grunt.
2. Implement glob functionality in every task that accepts files as input, for a more consistent API.

My suggestion would be to include a globbing function in Robo\Task\BaseTask or create a wrapper Glob class in the Robo\Common namespace (or maybe a larger FileSystem class with more common file tasks).

I think these changes would put Robo in better competition with other task runners.

boedah:
:+1:

dlancea:
In the mean time, to satisfy 1., you can use webmozart's glob in your robofile:

```php
use Webmozart\Glob\Glob;
use Webmozart\PathUtil\Path;

class RoboFile extends \Robo\Tasks{
	function assets(){
		// Get all .js files from the app javascript paths.
		$app_js = Glob::glob(Path::makeAbsolute('js/model/**/*.js', getcwd()));
		// concat app js files
		$this->taskConcat($app_js)
			->to(self::WEBROOT_PATH.'cache/js/models.js')
			->run();
	}
}
```

It would be nice if this was integrated, though.

burzum:
:+1:

gabor-udvari:
:+1: I relied on the globbing of Symfony 2 Finder for the [FlattenDir task](https://github.com/Codegyre/Robo/blob/adb287892c123e89e4354bb7bae7dec6853898da/src/Task/FileSystem/FlattenDir.php#L181), but I can see the the advantage of the Ant syntax.

DavertMik:
Closed in favor of @233

PR #228, opened 2015-10-13
Add ImageMinify task

gabor-udvari:
Minifies images. When the required minifier is not installed on the
system the task will try to download it from the [imagemin](https://github.com/imagemin) repository.

The task supports the following minifiers:
- optipng
- pngquant
- advpng
- pngout
- zopflipng
- pngcrush
- gifsicle
- jpegoptim
- jpeg-recompress
- jpegtran
- svgo (only minification, no downloading)

__Note__ before merging. The task works as intended, generated docs are included, the coding standard was checked. However the class is 603 lines of code and there are a few refactoring possibilities which you might want to consider:

- I am not sure if the `Robo\Task\Assets` namespace is the right one, maybe a new `Robo\Task\Images` namespace would be more future proof.
- The whole `findFiles` method is an exact copy from the FlattenDir task. There is already a [discussion about adding glob support](https://github.com/Codegyre/Robo/issues/216) for all the tasks. When that happens the findFiles method can be replaced.
- The task contains an `installFromImagemin` method which basically takes an URL, downloads the content and puts it into a file. Such a general download method could be created in the `Robo\Task\Base` namespace, maybe other tasks would benefit from it.

### DIFF
--- a/docs/tasks/Assets.md
+++ b/docs/tasks/Assets.md
@@ -1,4 +1,66 @@
 # Assets Tasks
+
+## ImageMinify
+
+
+Minifies images. When the required minifier is not installed on the system
+the task will try to download it from the [imagemin](https://github.com/imagemin) repository.
+
+When the task is run without any specified minifier it will compress the images
+based on the extension.
+
+```php
+$this->taskImageMinify('assets/images/*')
+    ->to('dist/images/')
+    ->run();
+```
+
+This will use the following minifiers:
+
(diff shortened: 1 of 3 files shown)


DavertMik:
wow, that's great! Thank you, reviewing...

DavertMik:
Thanks, for now I will merge it as it is, but will keep on mind the refactoring ideas you proposed

Issue #111, opened 2015-01-27
Using a custom Robo Class


brad-jones:
In the change log it mentions: __RoboFile can be changed to any provided class.__

This is not immediately obvious to me.

Is the idea that I simply create my own custom cut down version of the ```Robo\Runner``` class.
Removing all the logic around loading the ```RoboFile.php```.

Or is there something I am missing here???

boedah:
Maybe have a look at #95.

brad-jones:
If I use the example from #95 I get:
Call to undefined method Symfony\Component\Console\Application::execute()

If I change ```execute()``` to ```run()``` I get a working Symfony Console Application, however no output.

I believe the correct usage would be:

```php
\Robo\Config::setOutput(new Symfony\Component\Console\Output\ConsoleOutput());
$runner = new \Robo\Runner();
$app = $runner->createApplication('ClassNameWithTasks');
$app->run($optionalInputGoesHere);
```

In the end I created my own runner.
see: https://github.com/brad-jones/robo-tasks/blob/master/src/Runner.php

boedah:
Couldn't we add an `-f <file>` option to use a different file?

c33s:
it would also be good having an environment variable, to store the relative path of the `RoboFile.php` which should be used 
`export ROBO_FILE=src/Robo`

greg-1-anderson:
This has been implemented. You can use the --load-from option, or see [framework.md](https://github.com/consolidation/Robo/blob/master/docs/framework.md)

Issue #95, opened 2014-12-29
Allow use Robo with a custom class


oscarotero:
Hi.
I'd love to be able to execute Robo with a custom class instead define a robofile with a Robofile class inside. Even define a custom input. Something like:

```php
use Robo\Runner;
use Symfony\Component\Console\Input\ArgvInput;

$runner = new Runner();
$myTasks = new MyTasks();
$input = new ArgvInput($argv);

$runner->execute($myTasks, $input);
```

DavertMik:
Yes, this is allowed in 0.5.0 which is in current master branch:

```php
<?php
        $runner = new \Robo\Runner();
        $app = $runner->createApplication(MyTasks::class);
        $app->execute($argumentsAsString);
?>
```

See Runner.php for details

oscarotero:
Fantastic. Thank you :smile:

PR #232, opened 2015-10-22
using class variables instead of constants

rdeutz:
# Executive summary
These change allows to make myRobo packages easier. 

# Backwards compatibility
Full b/c, no problems are expected, anyway I changed the version to 0.6.0 because this is a new feature

# Translation impact
none

# Why
In further versions of Robo I had to copy code from the runner.php to make it act and work with a custom RoboClass and custom RoboFile-Name. Now I can create my package just with setting the Robo package as dependency and create my own runner.php with extending \Robo\Runner. The 2nd step is to create a Task.php and extend \Robo\Tasks and add my tasks. So I have all what Robo has plus my tasks easily as a package available. 

I hope that does makes sense for others too. Anyone can check how I did it here (is WIP ofc): https://github.com/rdeutz/jrobo/tree/development

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -14,22 +14,50 @@ class Runner
 {
     use IO;
 
-    const VERSION = '0.5.4';
+    const VERSION = '0.6.0';
     const ROBOCLASS = 'RoboFile';
     const ROBOFILE = 'RoboFile.php';
-
+    
+    /**
+     * @var CurrentDir
+     */
     protected $currentDir = '.';
+    
+    /**
+     * @var PassThoughArgs
+     */
     protected $passThroughArgs = null;
 
     /**
      * @var ConsoleOutput
      */
     protected static $printer;
-
+    
+    /**
+     * @var RoboClass
+     */
+    protected $roboClass;
+    
+    /**
+     * @var RoboFile
+     */
+    protected $roboFile;
+    
+    /**
+     * Class Constructor
+     */
+    public function __construct()
+    {
+        // set the const as class properties to allow overwriting in child classes
+        $this->roboClass = self::ROBOCLASS;
+        $this->roboFile  = self::ROBOFILE;
+    }
+    
+    
     protected function loadRoboFile()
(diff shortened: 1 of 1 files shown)


src/Runner.php
```diff
@@ -14,22 +14,50 @@ class Runner
 {
     use IO;
 
-    const VERSION = '0.5.4';
+    const VERSION = '0.6.0';
```
boedah: The version should not be changed in a PR.

src/Runner.php
```diff
@@ -14,22 +14,50 @@ class Runner
...
     const ROBOCLASS = 'RoboFile';
     const ROBOFILE = 'RoboFile.php';
-
+    
+    /**
+     * @var CurrentDir
```
boedah: There certainly is no class named `CurrentDir`, right?
Same for the other doc-block changes...

src/Runner.php
rdeutz: find by me

boedah:
Nice.
How does this relate to #95 and #111 ?

src/Runner.php
rdeutz: I just added doc blocks didn't check it this variables are used. You are right this one and $printer aren't used at all. I will remove them and update the PR

rdeutz:
I have seen #95 I think it is to complicated and had this idea, for me it is important to package all. We are planning to use it for the Joomla project and doing it this way allows us to support developers with some Joomla specific tasks

src/Runner.php
boedah: I just actually meant the `@var` declaration and not the usage, but you seem right. Those seem to be left-overs...

boedah:
I would say your method (where you need to create a new `Runner`) seems to be more complicated than the 3 lines of #95.
Nevertheless I think this PR is a good addition since it offers more ways to use Robo.

DavertMik:
Sure, this makes sense, merging it then

Issue #231, opened 2015-10-18
Check if files exist before using them


burzum:
taskConcat() and taskMinify() do not seem to care if the file you try to use exists or not.

What about adding a generic method (could be implemented in a trait as well) that can be used in all tasks to check a string or an array (iterate over it) if it's a path or file and throw a hard error if it does not exist?

DavertMik:
Yes, it should be added. Please do it )

burzum:
Do you want to throw an exception or use one of the methods to print out an error? If yes, which one please. :)

burzum:
@DavertMik whould this be good enough for you?

```php
<?php
namespace Robo\Common;

trait FileAndFolderCheck
{
	use IO;

	/**
	 * Checks if the given input is a file or folder.
	 *
	 * @param string|array $resources
	 * @param string $type "file", "dir", "both"
	 * @return void
	 */
	protected function checkResources($resources, $type)
	{
		if (is_string($resources)) {
			$resources = [$resources];
		}
		foreach ($resources as $resource) {
			switch ($type) {
				case 'file':
					if (!$this->isFile($resource)) {
						$this->printTaskError(sprintf('File "%s" does not exist!'), $this);
					}
				case 'dir':
					if (!$this->isDir($resource)) {
						$this->printTaskError(sprintf('Directory "%s" does not exist!'), $this);
					}
				case 'both':
					if (!($this->isDir($resource) || $this->isFile($resource))) {
						$this->printTaskError(sprintf('File or directory "%s" does not exist!'), $this);
					}
			}
		}
	}

	/**
	 * Convenience method to check the often uses "source => target" file / folder arrays.
	 *
	 * @param string|array $resources
	 * @return void
	 */
	protected function checkSourceAndTargetResource($resources) {
		if (is_string($resources)) {
			$resources = [$resources];
		}
		$sources = [];
		$targets = [];
		foreach ($resources as $source => $target) {
			$sources[] = $source;
			$target[] = $target;
		}
		$this->checkResources($sources);
		$this->checkResources($targets);
	}

	protected function isDir($directory)
	{
		return is_dir($directory);
	}

	protected function isFile($file)
	{
		return file_exists($file);
	}
}
```

DavertMik:
```php
    protected function checkResources($resources, $type)
    {
        if (!in_array($type, ['file', 'dir', 'both'])) {
            throw new \InvalidArgumentException(sprintf('Invalid resource check of  type "%s" used!', $type));
        }
[...]

PR #238, opened 2015-11-16
Adding a trait to check for files and folders

burzum:
This is a trait that should be used by every task that uses files and folders as input to check if they really exists.


src/Task/FileSystem/CleanDir.php
```diff
@@ -16,8 +17,13 @@
  */
 class CleanDir extends BaseDir
 {
+    use FileAndFolderCheck;
+
     public function run()
     {
+        if (!$this->checkResources($this->dirs, 'dir')) {
+            return Result::error($this, 'Resources are missing!');
```
DavertMik: It is hard to get in current context what `resources` are.

src/Task/FileSystem/DeleteDir.php
```diff
@@ -16,8 +17,13 @@
  */
 class DeleteDir extends BaseDir
 {
+    use FileAndFolderCheck;
+
     public function run()
     {
+        if (!$this->checkResources($this->dirs, 'dir')) {
+            return Result::error($this, 'Resources are missing!');
```
DavertMik: It is hard to get in current context what `resources` are.

burzum: Suggest a better name?

src/Task/FileSystem/CleanDir.php
burzum: Suggest a better name?

src/Common/FileAndFolderCheck.php
```diff
@@ -0,0 +1,89 @@
+<?php
+namespace Robo\Common;
+
+trait FileAndFolderCheck
```
DavertMik: strange that it is named `FileAndFolder` while the parameter used is `fileAndDir`. I'd rename it to `FileExistenceChecker` or `ResourceExistenceChecker`

DavertMik:
Thanks, I've got some comments

src/Task/FileSystem/CleanDir.php
DavertMik: For this specific case the error should state like "Source directories are missing"

DavertMik:
Wow, great work! Especially with tests and improving current tasks. 
Is that ready, or some other tasks should be refactored to use it?

burzum:
It's ready. I think there are no other tasks left.

boedah:
Just linking #231

DavertMik:
The day has come!

PR #240, opened 2015-11-21
Added the InputInterface filled with all arguments and options to Robo\Config

thomas-spigel:
I needed the full InputInterface, and found out that it's not saved into the Config object. 

With this patch, you can get the input with `$this->getInput()` or `\Robo\Config::get('input')` both directly from the RoboFile.


DavertMik:
Thanks!

PR #241, opened 2015-11-24
Cast relative path to absolute path

webflo:
I have the use-case to invoke robo with a relative path. The relativ path passes the first check in <code>:: loadRoboFile</code>, not sue why it did not work for the second file_exists check.


DavertMik:
Makes sense. Thank you

PR #242, opened 2015-11-27
Allow RoboClass without setting RoboFile.

derhasi:
When using robo within a script or an autoloaded environment, it is not necessary to provide the exact location of the RoboClass. So it would be nice to make that RoboFile optional.

This way we can do something like this.
```php
$robo = new \Robo\Runner('\MyProject\MyRoboFile');
$robo->execute(array('robo', 'something'));
```

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -49,6 +49,11 @@ class Runner
 
     protected function loadRoboFile()
     {
+        // In the case the RoboClass is already autoloaded, we do not need any further checks.
+        if (class_exists($this->roboClass)) {
+            return true;
+        }
+        
         if (!file_exists($this->dir)) {
             $this->yell("Path in `{$this->dir}` is invalid, please provide valid absolute path to load Robofile", 40, 'red');
             return false;


boedah:
Robo itself has a RoboFile. Does it get autoloaded if installing Robo via composer?

derhasi:
No, it doesn't, as only ./src is added to the autoloader

boedah:
Ah, of course ;)
Got confused because one needs to exclude Robos RoboFile in an IDE...

greg-1-anderson:
One of the main motivations for doing this is to run Robo directly from a Composer custom installer. Unfortunately, Composer only loads classes for custom installer's dependencies, not files. This makes components such as Guzzle, which depend on the inclusion of files, unusable in this context. See: composer/composer#4764

This feature may also be desirable in other scripting contexts, for environments where a script wishes to use Robo functionality without entering through the main `robo` file (e.g. a tool that has its own cli parser). However, in this context it might be better to support direct use of Robo tasks without the runner. That approach also requires a bit of work, and should be the subject of a separate PR.

This should stay open in case a use-case not involving autoload files presents itself. It probably wouldn't hurt to just merge it, although tests would be nice.

greg-1-anderson:
So, drupal-scaffold is now running Robo like this, except it is using a [custom Runner sublcass](https://github.com/drupal-composer/drupal-scaffold/blob/master/src/RoboRunner.php#L20) to make it work.

We could remove this class if this PR was merged; I'd recommend putting it in.

brad-jones:
My interim solution, if anyone else is interested.
https://github.com/brad-jones/robo-runner

greg-1-anderson:
The change from this PR is also included in #285.

DavertMik:
So I'm closing in favor of #285

PR #243, opened 2015-12-01
Allow Rsync::exclude to accept a list of patterns to exclude.

greg-1-anderson:
This PR allows `taskRsync::exclude()` to take an array of patterns, in addition to its traditional parameter value of a single string.

Motivation: sometimes, you might want to call taskRsync with a variable list of locations to exclude.

Before this PR:
```
    $rsync = $this->taskRsync()
      ->fromPath("$tmpDir/drupal-8/")
      ->toPath($webroot)
      ->args('-a', '-v', '-z')
      ->args('--delete');
    foreach ($excludes as $exclude) {
      $rsync->exclude($exclude);
    }
    $rsync->run();
```
After this PR:
```
    $this->taskRsync()
      ->fromPath("$tmpDir/drupal-8/")
      ->toPath($webroot)
      ->args('-a', '-v', '-z')
      ->args('--delete')
      ->exclude($excludes)
      ->run();
```

### DIFF
--- a/src/Task/Remote/Rsync.php
+++ b/src/Task/Remote/Rsync.php
@@ -229,16 +229,7 @@ class Rsync extends BaseTask implements CommandInterface
 
     public function exclude($pattern)
     {
-        if (is_array($pattern)) {
-            foreach ($pattern as $item) {
-                $this->exclude($item);
-            }
-        }
-        else {
-            $this->option('exclude', escapeshellarg($pattern));
-        }
-
-        return $this;
+        return $this->optionList(__FUNCTION__, $pattern);
     }
 
     public function excludeFrom($file)
@@ -248,6 +239,16 @@ class Rsync extends BaseTask implements CommandInterface
         }
 
         return $this->option('exclude-from', $file);
+    }
+
+    public function includeFilter($pattern)
+    {
+        return $this->optionList('include', $pattern);
+    }
+
+    public function filter($pattern)
+    {
+        return $this->optionList(__FUNCTION__, $pattern);
     }
 
     public function filesFrom($file)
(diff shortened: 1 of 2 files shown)


src/Task/Remote/Rsync.php
```diff
@@ -224,16 +224,21 @@ public function itemizeChanges()
...
-        return $this;
+        return $this->exclude(['.git/', '.svn/', '.hg/']);
     }
 
     public function exclude($pattern)
     {
-        return $this->option('exclude', escapeshellarg($pattern));
+        if (is_array($pattern)) {
+            foreach ($pattern as $item) {
+                $this->option('exclude', escapeshellarg($item));
```
boedah: Please call `$this->exclude` here to avoid duplication.

boedah:
Useful PR! 
Please see my comment.

greg-1-anderson:
Sure thing!  Thank you for the feedback.

boedah:
:+1:

greg-1-anderson:
At the risk of complicating this issue, I pushed another commit for consideration, adding 'includeFilter' and 'filter' options as well.

To avoid repeating the same pattern over and over again, I factored the option list implementation into an optionLis() method in CommandArguments.  The complication here is that, in order to be useful as a shorthand method, optionList() must call escapeshellarg on the values passed to it. This, however, makes optionList() inconsistent with option(), which does not escape.

Changing option() is not a good option (no pun intend), because existing code expects that escapeshellarg is necessary, and we do not want to double-escape. Forcing callers of optionList to use an awkward array_map to escape their array values would be possible, but this just seemed awkward.

In the end, I merely documented the difference, but could do it the other way if desired.

greg-1-anderson:
@boedah: Any feedback here?

greg-1-anderson:
Bump. This would be good to get in, but please consider the consistency-vs-backwards-compatibility issue above.

DavertMik:
Ok, looks good.

PR #189, opened 2015-07-16
PackZip

roland-d:
This adds zip support to Robo.li. This is the initial version of PackZip 
and is modeled after PackPhar for consistency.

Please test and review if it works OK for you. I have included a PHPDoc 
that explains how to use the PackZip.

Looking forward to your feedback.

PR #244, opened 2015-12-04
Add taskArchive and taskExtract, to archive and extract zip and tar archives.

greg-1-anderson:
Here is a useful task to extract archives.


boedah:
Nice!
Tests would be great!

greg-1-anderson:
Certainly. This would be most easily done in concert with #189.

DavertMik:
> This would be most easily done in concert with #189

Well, in #189 packing is done via Zip extension, in your case you use command line utils. 
Personally I prefer doing that through extensions as we can guarantee that unpacking will work withour 3rd-party utils even on Windows.

greg-1-anderson:
I don't have time to work on this this weekend, but I'd be happy to help integrate these two approaches next week.  I'll see if I can find a good `tar` library; if anyone has any recommendations, please post them here.

greg-1-anderson:
It appears that PHP 5.3 has built-in functions to [create a tar or zip archive](http://php.net/manual/en/phardata.buildfromdirectory.php), and [extract them again](http://php.net/manual/en/phardata.decompress.php). PECL required, though.

Maybe [Pear Archive_Tar](https://github.com/pear/Archive_Tar) would be better (more universally available) than PECL?  It's [available via Packagist](pear/archive_tar).

greg-1-anderson:
One step closer: using Archive_Tar and ZipArchive classes instead of exec now.

Still needs tests.

greg-1-anderson:
This is working well and is now ready for review.  The unit tests are working for me locally, but I get an error "Unable to create archive" on Travis.  I suspected this might be due to the missing ext-zip library, but when I tried installing it in  b7ba1f5, I got an error stating it was already installed.  The error sort of feels like some sort of a filesystem permissions error, and yet the recursive copy tests write folders in the sandbox just as my tests do.

Can anyone help diagnose what's going wrong on Travis?

greg-1-anderson:
I tried adding assets that the zippy installs in its .travis.yml file, but that did not work either.  I also rebased with master and merged in #251 for good measure.

It's still a head-scratcher why zippy is working fine locally, but not on Travis.

greg-1-anderson:
Switching to alchemy/zippy was a big mistake; although it has a nice object-oriented API, underneath it calls through to the executable binaries.

The latest commit I pushed here switches back to ZipArchive and Archive_Tar; PEAR and PECL required, but otherwise should be cross-platform.  All archives supported during extract; only zip supported during archive. More coming.

greg-1-anderson:
This is all implemented, tested, and ready for review.

greg-1-anderson:
### Naming things ###

I added an `extend` method to Result. The use-case here is to have sub-methods for a task that each return a Result; the main task would like to extend the provided result to include more data, e.g. the total execution time for the task.  There are other ways this could be done; the method could be called clone or copy, it could be a static method that took a Result parameter, or it could be a method that modified the $data field in place. Result is currently immutable, though, and I didn't want to make it mutable. Change to suit.

greg-1-anderson:
I decided I didn't like `extend`, because it sounds mutable. I renamed this method to `copy`. I picked `copy` over `clone` because the later implies that the object state does not change, whereas here it does.

composer.json
```diff
@@ -20,6 +20,7 @@
         "symfony/console": "~2.5",
         "symfony/process": "~2.5",
         "symfony/filesystem": "~2.5",
+        "pear/archive_tar": "~1.0",
```
DavertMik: should not be a strict dependency. Please move it to `require-dev`

src/Common/PHP.php
```diff
@@ -0,0 +1,20 @@
+<?php
+namespace Robo\Common;
+
+use Robo\Result;
+
+trait PHP
```
DavertMik: I'd rename it to smth like PHPStatus

src/Task/Archive/Archive.php
```diff
@@ -0,0 +1,215 @@
...
+ * $this->taskArchive(
+ * <archiveFile>)
+ * ->add('README')                         // Puts file 'README' in archive at the root
+ * ->add('project')                        // Puts entire contents of directory 'project' in archinve inside 'project'
+ * ->addFile('dir/file.txt', 'file.txt')   // Takes 'file.txt' from cwd and puts it in archive inside 'dir'.
+ * ->run();
+ * ?>
+ * ```
+ */
+class Archive extends BaseTask implements PrintedInterface
```
DavertMik: I'd rename it to pack

src/Task/Archive/Archive.php
```diff
@@ -0,0 +1,215 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * Add an item to the archive.
+     *
+     * Like file_exists(), the parameter may be a file or a directory.
+     *
+     * @var string
```
DavertMik: strange linebreak here. I'm not sure documentation generator will handle this )

src/Task/Archive/Archive.php
```diff
@@ -0,0 +1,215 @@
...
+     *   path in the archive, so absolute paths should not be used here.
+     *   If given an array, the key of each item should be the path to store
+     *   in the archive, and the value should be the filesystem path to the
+     *   item to store.
+     */
+    public function add($item) {
+        if (is_array($item)) {
+            $this->items = array_merge($this->items, $item);
+        }
+        else {
```
DavertMik: PSR-formatting!

src/Task/Archive/Extract.php
```diff
@@ -0,0 +1,192 @@
+<?php
+namespace Robo\Task\Archive;
+
+use Robo\Result;
+use Robo\Task\BaseTask;
+use Alchemy\Zippy\Zippy;
```
DavertMik: you actually not using it, right?

DavertMik:
Awesome work. Thank you! Looks like its almost ready to be merged.
Please merge with master and do changes in my comments.

greg-1-anderson:
Made suggested changes here; tests still failing on PHP 5.4 as previously mentioned. Other tests passing.

src/Task/Archive/Extract.php
```diff
@@ -0,0 +1,196 @@
...
+        @mkdir($extractLocation);
+        @mkdir(dirname($this->to));
+
+        $this->startTimer();
+
+        $this->printTaskInfo("Extracting <info>{$this->filename}</info>");
+        // Perform the extraction of a zip file.
+        if (($mimetype == 'application/zip') || ($mimetype == 'application/x-zip')) {
+            $result = $this->extractZip($extractLocation);
+        }
```
greg-1-anderson: Not sure what Nitpick-CI is trying to say here; a space should not come after a closing brace, newline is correct. Also, this line is the same as all of the other similar occurrences of this pattern.

greg-1-anderson:
Merged in #257 here so that tests would pass; also hand-fixed remaining Nitpick-CI complaints.

greg-1-anderson:
Removed #257 and rebased with master. Would be good to get this in.

PR #249, opened 2015-12-24
[WIP] Added task for downloading files with progressbar.

Koc:
Work in progress

### DIFF
--- a/src/Task/Remote/DownloadFileTask.php
+++ b/src/Task/Remote/DownloadFileTask.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Task;
+
+use Robo\Common\TaskIO;
+use Robo\Contract\TaskInterface;
+use Robo\Result;
+use Symfony\Component\Console\Helper\ProgressBar;
+
+class DownloadFileTask implements TaskInterface
+{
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Please hold this thought for post-#273 work. I'd like to avoid tasks writing to stdout by default. I'd like to see Robo introduce some way that a progress bar could be "attached" to a task when appropriate.

Note also that when #273 goes in, `$this->getOutput()` will disappear -- but you can get it back simply by adding `use IO;` in your task, which is allowed, but deprecated. Hopefully by supporting attachable progress indicators we can also avoid tasks having a reference to getOutput().

greg-1-anderson:
#300 allows tasks access to a progress bar; see ParallelExec and PackPhar for example usage.

You'll have to enhance the interface if you'd like to set the format of the progress bar, though, as at the moment that API is not exposed.

Koc:
@greg-1-anderson will rebase and change code after #300 will be merged

greg-1-anderson:
#300 has been merged.

greg-1-anderson:
Re-open if you want to return to this issue.

PR #250, opened 2015-12-25
Update the Lurker library dependency to ~1.1@dev

szymach:
The Lurker library in version ~1.0 has an issue, where sometimes it is impossible to watch directories. I propose updating it to version ~1.1 where I did not encounter this problem.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "symfony/console": "~2.5",
         "symfony/process": "~2.5",
         "symfony/filesystem": "~2.5",
-        "henrikbjorn/lurker": "1.0.*@dev"
+        "henrikbjorn/lurker": "~1.1@dev"
     },
     "require-dev": {
         "patchwork/jsqueeze": "~1.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,11 +4,12 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "4c146f082ebbabf4f5d4e797db4d52d3",
+    "hash": "f3c3d5e97422b992cfe5313b8ae06136",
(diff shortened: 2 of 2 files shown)


jjok:
:+1: I've been investigating this issue all afternoon. Just worked out where the issue is.

jjok:
Do we know why Robo uses a fork of Lurker? flint/Lurker seems to be more up to date. It has a 1.1 tag, whereas henrikbjorn/Lurker doesn't.

szymach:
Don't really know, I had to fork the Robo lib and use it in my project to make things work.

composer.json
```diff
@@ -20,7 +20,7 @@
         "symfony/console": "~2.5",
         "symfony/process": "~2.5",
         "symfony/filesystem": "~2.5",
-        "henrikbjorn/lurker": "1.0.*@dev"
+        "henrikbjorn/lurker": "~1.1@dev"
```
jjok: I would change it to `"flint/lurker": "~1.0"`. That should get you the 1.1 branch, and I think that version is more actively developed.

DavertMik:
> Do we know why Robo uses a fork of Lurker? flint/Lurker seems to be more up to date. It has a 1.1 tag, whereas henrikbjorn/Lurker doesn't.

In the days it was initially born, this version was the most stable one. Not sure, maybe its time to switch to other.

composer.json
DavertMik: There is no flint/lurker library on packagist:
```
https://github.com/flint/Lurker/blob/master/composer.json#L2
```

szymach: Right, it seems the repository is flint/lurker, but on packagist it is henrikbjorn/lurker. https://github.com/flint/Lurker So it is not a fork, simply the dependency version needs to be updated.

DavertMik:
Done. Also updated symfony version to match 3.x as well

PR #251, opened 2015-12-30
Make testErrorIsHandled unit test pass.

greg-1-anderson:
This test was failing for me locally; I am on php 5.6, but from reading the PHP docs, it seems this should have always been failing.  In any event, the version here works for me; let's see what Travis has to say.


DavertMik:
Thank you!

PR #252, opened 2015-12-31
Add task collections to Robo to support rollbacks and transient objects.

greg-1-anderson:
## Motivation ##

Robo provides a very useful way to create quick scripts that are easy to write, and easy to read. Exception handling is avoided, so the script flow is generally linear and free from unnecessary conditionals.

What if you wanted to ensure that a script always cleaned up its transient objects, and did not continue execution after one phase failed?  Robo has a useful stop-on-fail mechanism, supported in the `Result` class (hard exit after any task fails), and some stacks (taskFileSystemStack) also have a stop-on-fail switch. This does not always give the opportunity to clean up easily. More complicated solutions run the risk of adding a lot of conditionals to scripts. We want to keep things simple and linear, even as scripts become more complex.

## Task Collections ##

This PR expands on the idea of a task stack (which currently only works for operations that are all of the same kind) to also allow the creation of task collections.  Multiple tasks of different types can be queued up in a single collection, and then executed.  If one task fails, the rest of the tasks in the chain are not called.  Rollback and completion tasks may also be provided to help clean up.

## Example ##
```
$collection = $this->taskCollection();
$this->taskFileSystemStack()
      ->mkdir('logs')
      ->touch('logs/.gitignore')
      ->chgrp('logs', 'www-data')
      ->symlink('/var/log/nginx/error.log', 'logs/error.log')
      ->collect($collection, $this->taskDeleteDir('logs'));
/// ... collect other tasks
$collection->run();
```
The idea here is not to try to record every last `chmod`, and roll them back individually, but rather to collect operation results together in a target directory that can be cleaned up if the tasks do not finish, or left intact if all goes well.  In the example above, `collect()` replaces the `run()` method; it serves the same purpose, except that execution is deferred, and a second task is provided to clean up in the event of a rollback. (n.b. Scripts that do not care about rollback can continue to use direct execution via the `run()` method.)

`collect()` is provided via the Collectable trait, which is used by the BaseTask class, so all Robo tasks are collectable.  If the collected task implements the RollbackInterface, then it will be called on any rollback. If the collected task implements the CompletionInterface, then it will be called if all tasks in a collection complete without errors, or if there is a rollback.  Rollback and completion tasks are only called if the task itself was executed (or attempted to be executed).

This PR also provides for transient objects, such as temporary directories, which are always cleaned up at the completion phase of a task collection. Simple single-threaded scripts may create temporary directories without using a collection, and these will be removed when the program terminates.

## Naming Things ##

I put the collections in the Robo\TaskCollection namespace, although Robo\Task\Collection might have done as well.  A collection of tasks technically isa task, and task collections can be nested; all the same, perhaps `$collection = $this->taskCollection();` should just be `$collection = new Collection();`.  Instead of `collect()`, I might have used `defer()`; other names might also do here.  As it stands, everything fits in nicely with existing Robo code, which can continue to work as it currently does.

Let me know if you have any suggestions or feedback here.


greg-1-anderson:
I added another commit that used the concept of a `CompletionInterface` rather than a `ContainerAwareTask` to implement the TmpDir task.  I think we could just go with this, and remove `ContainerAwareTask`, `runInCollection()` and so on. It's simpler this way, and I don't think that there is anything that cannot be done simply by implementing these methods.

n.b. Tests are failing here, but only because I separated #251 into its own PR.

greg-1-anderson:
I implemented the change I suggested above, removing `ContainerAwareTask` & c., and updated the original description above to match the new state of the code -- which is much cleaner than previously.  I also merged #251 into this branch, so that the tests could be green.

greg-1-anderson:
I simplified the implementation of TmpDir by introducing a Transient trait, and TransientInterface.  I added a TmpFile implementation as well.

greg-1-anderson:
Trying out some new naming conventions.  I renamed `Collectable::collect()` to `Collectable::runLater()`, and added a method `Collection::runNow()` that simply calls through to `Collection::run()`.  We can't just rename `run()` to `runNow()` because Collection isa TaskInterface. Having two methods that perform exactly the same function is usually discouraged, but what I was aiming at here was better comprehensibility for new users. The relationship between `runLater()` and `runNow()` is much clearer than the relationship between `collect()` and `run()`.  I also considered renaming `collect()` to `enqueue()`, which is an improvement, but suffers from the same problem. We can go with one of the other naming alternatives if you prefer.

I also added another test demonstrating an alternate way of building collections via `Collection::add()`. This might be preferable in instances where all of the tasks can easily be represented on a single line. When many of the tasks in the collection are built up with multiple lines of chained setters, then I think that the `runLater()`-style syntax is more readable.

greg-1-anderson:
If folks do decide to build task collections via multiple calls to `Collection::add()`, it is possible that they might inadvertently introduce a spurious call to `runNow()` in the middle (e.g. via cut-copy-paste error). The latest commit will catch this sort of error.

greg-1-anderson:
I improved the usability of the _tmpdir shortcut by making it return the path to the directory created instead of a Result object. Test for an empty string to see if the directory creation failed, or just use stopOnFailure().

Naming question: should the shortcut be named `_tmpdir` like `_mkdir`, or `tmpDir` like `copyDir`?

DavertMik:
Oh my, that sounds like promise-based PHP :)
This really awesome idea and I like the implementation. I will write something more sane when I read everything )

greg-1-anderson:
One other thing I want to do here is integrate this feature with Result::stopOnFail().  If stop on fail is `true`, then we should ensure that any completion or rollback function of the current collection is executed, should execution of the script be terminated.

greg-1-anderson:
randomString should be factored out somewhere; not sure where.

DavertMik:
Once again thank you for a great idea. I see that collections has wide potential. In future they can even use events so maybe you can add hooks before and after tasks (as it is done in rake). 

I'd like to do some syntax proposals. Currently collection is a set of tasks and corresponding rollback actions. They are stored in a list like this:

* task1
* task2
* rollback1+2
* task3
* rollback3

I.e. it is possible to combine them with collection by applying

```php
$collection->add($this->taskXXX1());
$collection->add($this->taskXXX2());
$collection->rollback($this->taskYYY());
$collection->add($this->taskXXX3());
$collection->rollback($this->taskYYY3());
$collection->run();
```

Instead of using `getPath` and adding more getters to tasks I'd recommend to create a class `DeferedResult` which is returned by `runLater` call:

```php
 $deferedResult = $this->taskTmpDir()
        ->runLater($collection)
$collection->rollback($this->_remove($deferedResult->getData()['dirname']));
```

Current syntax is ugly, so I'd improve Result to have `ArrayAccess` interface, so this can be written as:

```php
$deferedResult = $this->taskTmpDir()
        ->runLater($collection)
$collection->rollback($this->_remove($deferedResult['dirname']));
```
To summarize it, my point is we should use Result class to pass data between tasks. In case of data in collection we can use `DeferedResult` class which will allow to access data by promise. This improves your example, as we can take the result of a task execution.

What do you think?

greg-1-anderson:
Yes, I like the idea of adding `ArrayAccess` to the Result class.  Doing this makes it easier to return a uniform result from `runLater()`, and it is doubly useful if it is the same result type that is returned from `run()`.

DavertMik:
I think Collections can be used to have `before` and `after` hooks. For instance, we can have a fully-functioning deploy script packaged and distributed as collection of tasks (like it is done in Capistrano)
From our side we should provide a way to hook into this collection. 

So tasks added to collection should have a name:

```
$collection->add('migrate', $this->taskExec('migrate')); 
$collection->before('migrate', $this->taskExec('...')); // hook
```

greg-1-anderson:
I think that named tasks are a good idea, but not every task necessarily needs to have a name in every use-case, so perhaps these should be optional:
```
$collection->add(['migrate' => $this->taskExec('migrate')]); 
```
or
```
$collection->add([$this->taskExec('migrate')]); 
```

DavertMik:
I think that array syntax is not necessary here. But I agree, both calls should be valid:

```php
$collection->add('migrate', $this->taskExec('migrate')); 
$collection->add($this->taskExec('migrate')); 
```

greg-1-anderson:
That works for me.

DavertMik:
Also I think that we should somehow pass defered results to the hooks. For instance, `after` hook should rely on results of task execution:

```php
$collection->after('migrate', function($result) {
    return $this->taskDoSmth($result['data']);
});
```
alternatively without callbacks this can look like

```php
$result = $collection->get('migrate') // returns task
  ->runLater() // returns deferred result;
$collection->after('migrate', $this->taskDoSmth($result['data']));
```

DavertMik:
Result is now ArrayAccessible https://github.com/Codegyre/Robo/commit/c897342c98fc68b179b7fcf8ef5f367d478cab9d

greg-1-anderson:
The code would be simpler if we could make the `before` / `after` hooks work uniformly with those added via `add` (or some variant thereof).  It is also important that `run()` and `runLater()` operate in the same way.

What do you think of changing the `TaskInterface::run()` prototype to `TaskInterface::run(Result $incrementalResults = null)`?  The convention here is that collections would pass incremental results to run, whereas folks would continue to use the same convention (parameterless `run`) for immediate-execution tasks.

Most tasks will not be incremental-results aware, and will just ignore the parameter passed to them. Tasks that `must` be used in a before / after context could `throw` if someone tried to `run` them instead of `runLater` them.

If we did this, then `before` and `after` could be just like `add`, except they would insert the item into the execution list somewhere in the middle instead of at the end.

DavertMik:
I think I like the idea, but could you provide a code sample, so I could better visualize it in my head?

greg-1-anderson:
Sure; working on it.

greg-1-anderson:
Changing the prototype of TaskInterface::run() to include a parameter did not work out. Providing a default value helped existing code that called run(), but PHP still expects any derived class that overrides the run() method to conform exactly to its prototype. Therefore, if we added a Result object as a parameter to run(), then it would have to be added to every run() method everywhere. Even if we are okay with breaking b/c, that's just too messy, as most standard tasks won't care about incremental results.

So, instead of this, I introduced a new interface, FilterTaskInterface, that filter tasks may implement. The run() method of FilterTaskInterface takes a Result parameter, and filter tasks may modify this.  I modified the collection's task stack to require all tasks in the stack to be filter stacks, and I allow clients to provide regular tasks to Collection::add() by wrapping them in a TaskWrapper, which isa FilterTaskInterface that hasa TaskInterface (the task being executed).  This same task wrapper is also responsible for keeping the before and after task lists. Keeping the before and after lists in the wrapper class like this keeps things organizes, so they do not get separated, and any Collection::rollback() will happen after all of the after filters, regardless of what order they are added in.

greg-1-anderson:
Oh, another question: can we adjust NitPick CI to allow embedded classes inside unit test files?  It's possible to tell composer to set up a separate autoload directory for unit tests ([drush does it](https://github.com/drush-ops/drush/blob/master/composer.json#L52)), but for simple tests, it's more convenient to just have small test classes in the same file as the unit tests. Not sure if NitPick CI is configurable like this; we can set up a require-dev autoload folder if you prefer.

greg-1-anderson:
Oh, I merged the defer-write branch in here, but I think that was unnecessary.

DavertMik:
> can we adjust NitPick CI to allow embedded classes inside unit test files?

NitPick has no configuration ) 
But you can disable coding standards instead

```php
// @codingStandardsIgnoreStart
// @codingStandardsIgnoreEnd

// or

// @codingStandardsIgnoreFile

```

DavertMik:
Actually I didn't get why it is called Filter. Also I'd like all task to share one TaskInterface with single method `run`.

I think we can get back to the idea of `runLater` method. Tasks should be simple to use no matter they are used in collection or not.

greg-1-anderson:
It was called filter because it allows you to filter or adjust the results of some other task.  Maybe `alter` would be a much better name here.  Note that a task only needs to implement this interface if it wants to receive the intermediate results as a parameter.  In most cases, `before` and `after` tasks can be standard tasks that implement TaskInterface, implement the `run` method, and return additional result data in a new unique key. The alternate interface is only needed for an 'after' task that needs to know the results of a previous task.  I think that having a separate interface for this operation follows the design pattern that common things (ordinary tasks) should be simple, and uncommon things should be possible. Forcing all tasks to take an unused 'intermediate results' parameter feels wrong.

I'm not sure what you mean in your comment about `runLater`. Do you mean that this method is unnecessary? If that is your comment, I agree in principal, as everything can be done with `Collection::add`.  However, if you have a bunch of stack-based tasks, each with a large number of chained methods, then `runLater` is a nice alternative.

To illustrate:

Code choice 1:
```
$c = new Collection();
$taskA = $this->taskFrob()
   ->a()
   ->b()
   ->c();
$taskB = $this->taskMaul()
  ->e()
  ->f()
  ->g();
$c->add([$taskA, $taskB]);
$c->run();
```
That's a lot of local variables; we can get rid of them via:
```
$c = new Collection();
$c->add(
  [
    $this->taskFrob()
       ->a()
       ->b()
       ->c(),
    $this->taskMaul()
      ->e()
      ->f()
      ->g(),
  ]
);
$c->run();
```
If you don't mind the indentation, then `runLater` is not necessary. However, the nice thing about `runLater` is that code that uses it reads very similarly to the same non-collection-based function. The maintenance cost of `runLater` is very low; just ensure it is part of the base task, and it is available for use wherever needed.

On a related note, the TmpDir task does, I think, still need a `getPath` accessor method. Simply putting the path into the result object does not help, as code that use temporary directories will need to know the path to the directory to be created, so that other tasks can be created (e.g. place a file inside the temporary directory, etc.). The result isn't available until `run`, so we need an accessor.  Making results usable as arrays is very useful for other use-cases, though.

DavertMik:
Actually I was wrong, you can ignore Nitpick files https://gist.github.com/adamwathan/bf382f05c575db4efb6c

greg-1-anderson:
I renamed `runLater()` to simply `add()`, and removed the `runLater()` method.  Compare the following syntax to the example above:
```
$c = new Collection();
$this->taskFrob()
   ->a()
   ->b()
   ->c()
   ->add($c);
$this->taskMaul()
  ->e()
  ->f()
  ->g()
  ->add($c);
$c->run();
```
I think this is similar enough as to not cause confusion, but also provides a useful enough alternative that it's worth keeping around.

I also renamed the 'filter' interface for better clarity.

DavertMik:
I'd replace `->add($c);` to `->addToCollection()` or `addTo()` or `->collect($c)`. Add sounds too general for me

DavertMik:
>  The result isn't available until run, so we need an accessor. Making results usable as arrays is very useful for other use-cases, though.

Yes, so we should have such values to work as promises. So there is a proxy class DeferedResult, which is ArrayAcess as well. But when you access its data it returns an instance of a `Deferred` class. 
Deferred can be passed to param of any task but on execution real value must be unpacked.

greg-1-anderson:
Okay, take as an example the code for the current functional test for the tmpDir task:
```
        $tmpPath = $I->taskTmpDir()
            ->addToCollection($collection)
            ->getPath();
        $result = $collection->add(
            [
                $I->taskFileSystemStack()->mkdir("$tmpPath/log")->touch("$tmpPath/log/error.txt"),
                $I->taskCopyDir([$tmpPath => 'copied3']),
            ]
        )->run();
```
The advantages here are:

- We can get $tmpPath easily with the accessor (prior to the creation of the tmp dir), and use it when adding out other tasks to the collection.
- The tmp dir is automatically cleaned up on rollback (when any task fails) or complete (if `run()` finishes successfully)

Are you thinking that we might get rid of the `getPath()` accessor using a promise using some code similar to this (below)?
```
        $tmpDirResult = $I->taskTmpDir()
            ->addToCollection($collection); // n.b. change return type of addToCollection to DeferredResult
        $tmpDirPath = $tmpDirResult['path'];
        $result = $collection->add(
            [
                $I->taskFileSystemStack()->mkdir("$tmpPath/log")->touch("$tmpPath/log/error.txt"),
                $I->taskCopyDir([$tmpPath => 'copied3']),
            ]
        )->run();
```

This code looks fairly similar to the first example, but there are some stark differences.

- The first example needed no explicit error checking prior to `run()`; if the tmp dir cannot be created, then the task rolls back, as usual. In the case of the second example, error checking is required at `$tmpDirPath = $tmpDirResult['path'];` to ensure that the tmp directory really was created.
- In the first example, all tasks will always execute in the order they were added to the collection. In the second example, some tasks may run sooner, e.g. accessing `$tmpDirResult['path']` will create the temporary directory. In theory this does not matter, but in practice it introduces complexity.

I tend to think that the concept of promises should be reserved for code that absolutely needs async behavior (e.g. via a `runAsync()` method), and leave accessors for the simpler use cases.

If you had another idea, please let me know what you had in mind.

greg-1-anderson:
Okay, I tried some things, and I have an opinion on how to proceed here.

1) My [opinion on promises has not changed](https://github.com/Codegyre/Robo/pull/252#issuecomment-172399942). We should avoid them in favor of accessors until we have a real need for async code.

2) As you said, there should only be one TaskInterface. Passing the Result to `run()` is not the right solution, because most tasks don't need this value, and doing this is a huge burden on simple use cases that want to call tasks without collections. The simple use-case should not be impacted by the existence of collections.

3) `after` tasks that need the intermediate results can wrap their tasks with a class that has a reference to the collection, just as the rollback registration class currently does. The collection can keep the intermediate results in a field, and make it accessible via an accessor. After tasks then use this accessor if they need to examine or modify the intermediate results.

This will be clean + functional.  What do you think?

greg-1-anderson:
The commit I just pushed pretty much implements my suggestion above. Some minor changes are still needed; for example, I think it would be a good idea to accumulate result data under the name of the task.  For example, if you have a collection that runs `taskFoo` twice, named 'a' and 'b' with different parameters, and `taskFoo` returns a result 'bar', then $result->get('a/bar') and $result->get('b/bar') should be the way to access the results of the two calls to taskFoo.  If a task is called multiple times but is not given names, then the results will overwrite ($result->get('bar')), as they do today.

greg-1-anderson:
Actually, the niftier thing to do would be to store the entire Result object from each task in the data of the top-level Result.  Then you would be able to just access `$result['a']['bar']`.

greg-1-anderson:
I added a method `runPreservingTransients()`, which allows you to run all of the tasks on a collection so far, and get a result back, and then continue adding tasks to the collection, and run those as well.  The transients (e.g. tmp folders) from the first set are not deleted until the `run()` method completes.

Not sure if I like it or not, but it is an option.

greg-1-anderson:
By giving the FunctionWrapper task class a reference to the collection, it is possible to provide an `after` interface that uses anonymous functions as @DavertMik [suggested in an earlier comment](https://github.com/Codegyre/Robo/pull/252#issuecomment-170897178). This gives us easy access to intermediate results in after hooks without requiring deferred results, and without requiring us to pass Result objects into `run()` methods.

It's pretty much working now, except I still have some bugs in my result-merging functions. I'll add another commit here once that is worked out.

greg-1-anderson:
@DavertMik: There are two things I would like you to consider.

1) Access to results of previous tasks from within a task adds a lot of complexity to the code.

It would be possible for me to continue to work out the issues with merging task result data such that `after` tasks can look up and modify the results from previous tasks; however, the code to make this work right is kind of complex and arbitrary. Rake, in comparison, seems to make little provision for adjusting the results of one task in another task. There may be some techniques for doing it, but it is not provided by any of the Rake APIs, and there is little discussion of this technique in the Rake discourse, making me think it is uncommonly done. More often, folks do things in two stages: decide what needs to be done, and set up variables, and then declare tasks that perform actions, passing needed information in as parameters. Sometimes, tasks act on the current state of objects on the filesystem, which is sort of a de-facto way to pass information, but even this, when done, is to get the filesystem into a desired state, and not to pass values.

An alternative to passing / making available `$result` in an after task would be to have folks manage their shared variables themselves.  For example:
```
class MyTaskA implements Taskinterface
{
    public function __construct(MySharedObject $shared) { ... }
}
```
If `MyTaskB` is declared similarly, the user could then instantiate an instance of `MyTaskA` and `MyTaskB`, and pass them both the same shared object. If `MyTaskB` wants information from `MyTaskA`, then `MyTaskA` can provide it by whatever mechanism the two tasks agree upon. An even more straightforward way to do this would be to put the two tasks into the same class:
```
class MySharedController
{
    public function myTaskA() { ... }
    public function myTaskB() { ... }
}
```
The two methods can now share information through fields of their class.  To run these methods with Robo, the user would simply wrap MySharedController with one TaskInterface that called `$controller->myTaskA`, and another TaskInterface that called `$controller->myTaskB`.  We could even provide such a wrapper class (a run method that calls the provided method of a provided object) in Robo, but beyond that, Robo would be uninvolved in the sharing of information between tasks, as this would only be done cooperatively (i.e. never by patching someone else's task).

If that sounds good to you, then I'll just remove result-sharing from task collections.  Collection::run() will still return the unioned results from each task that was executed, but we can get rid of the dissatisfying results accumulator inside of Collection, and the accessor methods to the same.

2) Maybe the Collection, not the Robo\Tasks, should be the place where task traits are `use`d.

Moving all of the loadTasks from Robo\Tasks into the task collection would allow things such as:
```
$collection->taskFileSystemStack()
            ->mkdir("somewhere/log")
            ->touch("somewhere/log/error.txt");
```
... and so on, for as many tasks as you wished to add to the collection.  If you want to run a task early, just add a `run()` at the end of the task chain, and it will run, and its rollbacks / completion tasks will queue up. The rest will execute when you call `$collection->run()`.

Here's the real kicker, though. If we make Robo\Tasks extend Collection, then existing Robo files will continue to run the same way they always have.  The main difference is that now, tasks **always** have a reference to their collection (if they want it -- `loadTasks` just needs to pass `$this` in to the constructor of the task when it instantiates one).  The collection for Robo\Tasks is the global Collection, whose transients are cleaned up when the script terminates.

I'm definitely going to move forward with 2), as it is going to clean up a lot with transients, and will also make it convenient for using Robo tasks in contexts other than a class that extends Robo\Tasks.

I should have a new commit for 2) ready soon; please let me know what direction you think we should go vis-a-vis 1).

greg-1-anderson:
2) above turned out to not be very convenient. The way transients work today is, they are wrapped in a CollectionTask that belongs to the global (cleaned up at termination time). Internally to the collection class, all tasks added are also wrapped in a CollectionTask. If a transient task is added to a collection, it is unwrapped from the CollectionTask it is in.  If we extended this model to **all** tasks, then everything returned from any `$this->fooTask()` would be wrapped in a CollectionTask.  Note that CollectionTask adds a task's rollbacks and completions to its collection at `run()` time, and calls through to its wrapped task's methods via __call(). This seems fine for transients, but I'm not sure it is the right solution in general for all tasks. Perhaps what we need is a better mechanism than CollectionTask for managing transients.

As for 1) above, I removed the classes CompletionTask and RollbackTask, and instead now use FunctionPointerTask to wrap a task to call its complete() and rollback() methods, as needed. This same technique can be used to implement something like the `MySharedController` example shown above. I think this is the best way for tasks to cooperatively share results.  Note that you could also wrap some other task in a controller, call through to the other task's `run()` method in your controller's run method, and cache its $result until it is needed in a follow-on task. In general, though, I think that result-sharing inside tasks can usually just be avoided entirely.

Let me know if this all makes sense to you; if you agree, I'll just back out result aggregation. At that point, I think this will be pretty much ready to go.

src/Result.php
```diff
@@ -153,6 +153,28 @@ protected function printSuccess()
...
+        $this->data = $data + $this->data;
+        return $this;
+    }
+
+    /**
+     * Merge another result into this result.  Data already
+     * existing in this result takes precedence over the
+     * data in the Result being merged.
+     */
+    public function merge(Result $result)
```
DavertMik: I think this should replace `overlay` method. Just pass additional `$replace` param into it

greg-1-anderson: I don't like mode-altering parameters; however, I think that `overlay` can probably go away.

src/Contract/RollbackInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ * be called when the task collection it is added to
+ * fails, and runs its rollback operation.
+ *
+ * Rollback operations can also be managed as transient
+ * tasks, just as in the CompletionInterface. However,
+ * in this instance, the rollback operation will only
+ * be done if TransientManager::fail() is explicitly called.
+ *
+ * Tasks that implement RollbackInterface should consider
+ * implementing TransientInterface instead.
```
DavertMik: "instead"?

src/TaskCollection/TransientManager.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Since the results could be undefined if cleanup functions were called
+ * at arbitrary times during a program's execution, cleanup should only
+ * be done immeidately prior to program termination, when there is no
+ * danger of cleaning up after some unrelated task.
+ *
+ * An application need never use the TransientManager directly, save to
+ * call TransientManager::complete() immediately prior to terminating.
+ */
+class TransientManager
```
DavertMik: why not to call it smth like `GlobalCollection` ?

src/Result.php
DavertMik: I don't like them either but 2 methods of same kind is too verbose

src/TaskCollection/FunctionPointerTask.php
```diff
@@ -0,0 +1,64 @@
...
+ * also works with methods of objects.
+ *
+ * Example usage:
+ *
+ *      $collection->add(new FunctionPointerTask([$myobject, 'mymethod']));
+ *
+ * The collection will call $myobject->mymethod() when $collection->add() is
+ * called.
+ */
+class FunctionPointerTask implements TaskInterface, CollectedInterface
```
DavertMik: I think this can be renamed to `CallableTask`

greg-1-anderson:
This code has been much simplified now, and the names of things are also much simpler.

One other thing I could still do is move classes that the client does not need to use directly into a Robo\Collection\Internal namespace.  The classes that could be internal are:

 - CallableTask
 - Element
 - TaskWrapper

Let me know if you think that would be an improvement, or if you have any other feedback.

DavertMik:
I had it reviewed and I really like it!

greg-1-anderson:
Should I move those last three classes into an "internal" namespace?

PR #256, opened 2016-01-08
Wrapper task

greg-1-anderson:
This PR introduces a new base task, WrapperTask, to make it easier to create Robo tasks from existing PHP class libraries.  FilesystemStack is ported to use it as an example.  See comments in the file for more details on usage. 

Merged in #251 again, so that tests will pass.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -12,8 +12,11 @@ class RoboFile extends \Robo\Tasks
         $leadingCommentChars = " * ";
         $methodDescriptions = [];
         $methodImplementations = [];
+        $immediateMethods = [];
         foreach ($delegate->getMethods(ReflectionMethod::IS_PUBLIC) as $method) {
             $methodName = $method->getName();
+            $getter = preg_match('/^(get|has|is)/', $methodName);
+            $setter = preg_match('/^(set|unset)/', $methodName);
             $argPrototypeList = [];
             $argNameList = [];
             $needsImplementation = false;
@@ -34,8 +37,13 @@ class RoboFile extends \Robo\Tasks
             if ($methodName[0] != '_') {
                 $methodDescriptions[] = "@method $methodName($argPrototypeString)";
 
+                if ($getter) {
+                    $immediateMethods[] = "    public function $methodName($argPrototypeString)\n    {\n        return \$this->delegate->$methodName($argNameListString);\n    }";
+                } elseif($setter) {
+                    $immediateMethods[] = "    public function $methodName($argPrototypeString)\n    {\n        \$this->delegate->$methodName($argNameListString);\n        return \$this;\n    }";
+                }
                 // Include an implementation for the wrapper method if necessary
-                if ($needsImplementation) {
+                elseif ($needsImplementation) {
                     $methodImplementations[] = "    protected function _$methodName($argPrototypeString)\n    {\n        \$this->delegate->$methodName($argNameListString);\n    }";
                 }
             }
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
In 3f08a05, I added a simple proof-of-concept code generator, to provide all of the @method declarations for a wrapped task, so that they do not have to all be typed by hand. If this was to be kept, it should be factored out into a class, and its template moved with it.

This begs the question, is it better to just flesh out the code generator, and make it copy all of the public methods from the delegate into the wrapper class, so that __call() (and in fact all of WrapperTask) can simply be removed?

I think I still prefer the dynamic version, because it produces less code, and is more loosely coupled, making it more resilient in terms of adapting to additions in the delegate. The code generator would be a fair bit more complicated as well, and would be something extra to maintain. On the flip side, though, when using WrapperTask as a base task, it is still necessary to copy a method into the base class if it has default parameter values. This decreases the value of the WrapperTask slightly for every method of the delegate that has default parameter values.

src/Task/WrapperTask.php
```diff
@@ -0,0 +1,188 @@
...
+                    return $this;
+                }
+                // Otherwise, we'll defer calling this function
+                // until run(), and return $this.
+                $this->addToCommandStack($command, $args);
+                return $this;
+            }
+        }
+
+        die("Method $function does not exist.\n");
```
DavertMik: waaat? Don't die please :) Throw exceptions

RoboFile.php
```diff
@@ -3,6 +3,48 @@
 
 class RoboFile extends \Robo\Tasks
 {
+    public function wrap($className, $wrapperClassName = "") {
```
DavertMik: Thanks for this task. Could you write in comment how to use it? So other could use it to implement their tasks too

greg-1-anderson: Yes, I'll write documentation after this is factored out into a separate class.

greg-1-anderson:
Tests are failing here on php 5.4 due to:
```
symfony/console v3.0.1 requires php >=5.5.9 -> your PHP version (5.4.45) or "config.platform.php" value does not satisfy that
```
What do you want to do about that? Drop support for php 5.4, or try to peg to an earlier version of the Symfony components?  I'd recommend the former.

src/Task/WrapperTask.php
greg-1-anderson: Sorry, put this in as a placeholder until I decided which exception to throw, then forgot about it. Decided to use RuntimeException, like DynamicParams does.

greg-1-anderson:
Merged in #257 here so that tests would pass.

I renamed WrapperTask to StackBasedTask, as I realized this would also be useful for stack-based tasks that did not wrap some other API, just to get the feature to easily use taskFoo()->foo()->run() to get a deferred call to _foo().

On the subject of `_foo()`, elsewhere it was mentioned that this was contrary to a PSR style guideline against prepending `_` to function names to indicate that the method is protected or private. In this instance, the intention was not to identify the access level of the method, but to disambiguate `foo()` (the method that is called) and `_foo()` (the method that is executed at `run()` time). Other conventions could be used, though.

`foo()` and `do_foo()` -- disambiguate between the two functions by mixing snake_case with camelCase. This ensures that `do_foo()` will never be used elsewhere in the program.

`foo()` and `doFoo()` -- use camelCase for consistency. Runs the risk that `doFoo()` might be used in some classes, but this only matters if someone calls `foo()` from the wrapper.

The last option might be best if we want all of the sources to strictly pass the linter.  I prefer the underscore prefix exactly because it breaks the expected pattern, calling attention to the fact that something different is happening here.  Comments can server that purpose just as well, though.

Let me know if you have a preference here.

greg-1-anderson:
If you don't like `foo()` and `_foo()`, I might suggest `foo()` and `fooImplementation()` might be the best alternative. Putting the variable part at the end means that we won't need to `ucfirst()` the `foo` part.

burzum:
:+1: @DavertMik can we merge it?

greg-1-anderson:
Rebased with master.

The code generator needs to be refactored, but that can happen after merge if you want to go ahead and put this in.

greg-1-anderson:
Also, there is still an [important naming convention question to be answered](https://github.com/Codegyre/Robo/pull/256#issuecomment-170600816). I like it the way it is, but please consider which naming conventions you'd prefer here.

Issue #230, opened 2015-10-14
Future of Robo - what purpose should Robo serve?


boedah:
Through its lifespan Robo evolved from a simple task runner to a general purpose, one-for-all tool.

@DavertMik what is your vision for Robo? What are its core features?

Especially with all the Asset-Tasks added lately (just saw the PR for image minification):
How can Robo compete (and does it want to compete) with something like the Node.js ecosystem (gulp et al.)?
Have not tested it, but would say that compiling a complex SASS suite would take significantly longer (even longer than ruby-sass?) than compiling it with gulp.

I would question whether we should re-implement (and support!) long-living Node modules in PHP.

What is your opinion about that?

gabor-udvari:
My motivation to contribute all the new tasks was so I can create a PHP only version of the [Sage](https://github.com/roots/sage) Wordpress starter theme, which I named [Phlomis](https://github.com/gabor-udvari/phlomis). After switching my Wordpress workflow to use Composer for installing and updating the core and plugins I found it very frustrating that I needed to use Node, npm and gulp for building a Wordpress theme. I am still missing some features, like compiling with libsass and updating the browser, but I think the current state is promising.

Comparing the two building processes, Sage looks like this:
```shell
git clone https://github.com/roots/sage.git && cd sage
npm install
bower install
gulp build
```

Phlomis looks like this:
```shell
git clone https://github.com/gabor-udvari/phlomis.git && cd phlomis
composer.phar install
vendor/bin/robo build
```

Time required for the above commands with empty cache:

<table>
<tr><td>Sage</td><td>Phlomis</td></tr>
<tr><td width="50%">
<pre>
time npm install
real    5m13.584s
user    5m29.869s
sys     0m57.984s

time bower install
real    0m11.879s
user    0m5.250s
sys     0m1.268s

time gulp build
real    0m25.906s
user    0m24.060s
sys     0m1.383s
</pre>
</td>
<td>
<pre>
time composer.phar install
real    2m56.736s
user    0m25.762s
sys     0m4.583s

time vendor/bin/robo build
real    0m5.597s
user    0m5.417s
sys     0m0.107s
</pre>
</td></tr>
</table>

[...]

PR #257, opened 2016-01-11
Remove support for php 5.4, per requirements of symfony/console.

greg-1-anderson:
Tests are failing for PHP 5.4, as that version is no longer supported by symfony/console. This PR removes support for 5.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     },
     "bin":["robo"],
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=5.5.9",
         "symfony/finder": "~2.5|~3.0",
         "symfony/console": "~2.5|~3.0",
         "symfony/process": "~2.5|~3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "a50a68ff419ed8d20b14c4e82d687afd",
-    "content-hash": "6763afcb357b1145b2f5ff3f975ea076",
+    "hash": "65761cbe45c6920766a1168bb95cf00e",
+    "content-hash": "733f3af89e677b172b152aeb17947393",
     "packages": [
         {
             "name": "henrikbjorn/lurker",
@@ -2200,7 +2200,7 @@
     "prefer-stable": false,
     "prefer-lowest": false,
     "platform": {
-        "php": ">=5.4.0"
+        "php": ">=5.5.9"
     },
     "platform-dev": []
 }


DavertMik:
Ah.... No need to change requirements. We can just use older symfony version. 5.4 will work with Symfony 2.5, however composer.lock sets Symfony version to the latest one which is 3.0

greg-1-anderson:
It seems that keeping composer.lock as it is would be best. Maybe .travis.yml should do a `composer require ... ~2.5` for all relevant Symfony components if php version is 5.4?

greg-1-anderson:
Actually, `composer update` for php 5.4 (override composer.lock), vs. `composer install` for php 5.5 and later should suffice.

The problem here, though, is that then you would not have a composer.lock for php 5.4.  Would it make sense to keep php54-composer.lock, and move that into place during tests?

greg-1-anderson:
Maintaining two composer.lock files would be really hard to maintain.  Excluding Symfony 3 from composer.json is another possibility, but in the balance, it seems that testing with the latest version is better.

My previous suggestion (use `composer update` for php 5.4) seems better than that. The problem here, though, is that Travis would potentially load different versions of each component for each test,  and folks using php 5.4 would not be able to guarantee that they were running the same set of components that Travis was testing. The risks with doing this is: php tests can spontaneously break, and users may have trouble getting the right components.  These risks are mitigated by the fact that the older versions of the software probably will not change often, and when they do, they will probably remain roughly compatible, so it would probably tend to work most of the time.

In the balance, though, I think that maybe removing the php 5.4 tests, but allowing php 5.4 in composer.json might be a reasonable compromise. Users could be warned that 5.4 is at their own risk, and given instructions that they must run `composer update` to get the older versions.

greg-1-anderson:
This is another reason why it would be good to have a stable release of Robo. Then, it would be possible to make the Robo 1.x branch pin to Symfony 2.5, and the Robo 2.x / master branch pin to Symfony 3.x.  This is essentially what we do for Drush.

Taking this route would require making a 1.0 *and* a 2.0 release of Robo soon.  Might not be a bad idea, though.

greg-1-anderson:
I made a case for having two branches of Robo in #230.

DavertMik:
I'd rather keep both branches working as it is not a problem here. We don't have any incompatible changes, only some incompatible dependencies. It's just an issue of a travis ci, not a big deal. I will try to update `travis.yml`:

@greg-1-anderson hey, do you have Skype? I'm thinking on your proposal to be a comaintainer here :)

greg-1-anderson:
Yes, I am greg.1.anderson on Skype. My schedule is open for the next four hours or so, but I'll need to grab a conference room to have a place to talk. Message me on skype, and we can work out a convenient time to talk.

greg-1-anderson:
As for moving forward here, I think that the following would work:

1. Keep travis.yml as-is
2. Modify composer.json to only support Symfony ~2.5
3. Create the 1.x branch and tag Robo 1.0.  (master branch is now Robo 2.x)
4. Modify composer.json to only support Symfony ~3.0
5. Merge this PR on the master branch
6. Tag Robo 2.0 on the master branch

As long as no b/c breaking changes are introduced, future PRs can be committed to both branches in lockstep.  I don't think that any of my currently open proposals break b/c yet, but it might be a good idea to wait just a bit to see if that's going to be the case for sure before making the tags and branches as described above.

PR #265, opened 2016-02-05
Revert "Wrapper task"

DavertMik:
Reverts Codegyre/Robo#256


greg-1-anderson:
I was able to rebase #252, so it is not necessary to revert here.

PR #269, opened 2016-02-05
Mark Result::_invoke() deprecated.

greg-1-anderson:
`$result()` is an odd pattern. Overriding `__invoke()` should be done in instances where the object in question is acting like a php `Callable` object.  Introducing this as a shortcut for an accessor is less desirable than simply using an accessor, as the intent of the code is not obvious.

Since there are better ways to detect errors and recover, we should mark this deprecated for 1.0, and remove it in 2.0.

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -109,8 +109,12 @@ class Result implements \ArrayAccess
         return $this->exitCode === 0;
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


DavertMik:
Ok, no problems with that

PR #270, opened 2016-02-05
Make stop on fail the default for FilesystemStack.  BREAKS backwards compatibility.

greg-1-anderson:
Also turns off stop on fail for any task that is added via addToCollectionAndIgnoreErrors().

This change is best for consistency, so that collections work correctly.  While this breaks backwards compatibility, any Robo script that relies on stopOnFail defaulting to false could use a little updating anyway.

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -87,6 +87,8 @@ class Collection implements TaskInterface
      */
     public function rollback($rollbackTask)
     {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Although I have this marked as breaking backwards compatibility, this change only affects error recovery, so most scripts won't notice the behavior most of the time.

PR #271, opened 2016-02-06
Beginning of Collections documentation

greg-1-anderson:
Just getting started here; will continue with additional information.

### DIFF
--- a/docs/collections.md
+++ b/docs/collections.md
@@ -73,9 +73,7 @@ Since the concept of temporary objects that are cleaned up  on failure is a comm
 
 ### Temporary Directories
 
-It is recommended that operations that perform multiple filesystem operations should, whenever possible, do most of their work in a temporary directory that is moved into its final location near the end of the task collection. This makes it easy to clean up after an aborted task.
-
-If you use a temporary directory created by `$this->taskTmpDir()`, it will automatically be removed when the collection completes or rolls back. Move the temporary directory to another location to prevent its deletion.
+It is recommended that operations that perform multiple filesystem operations should, whenever possible, do most of their work in a temporary directory. Temporary directories are created by `$this->taskTmpDir()`, and are automatically be removed when the collection completes or rolls back. Move the temporary directory to another location to prevent its deletion.
 
 ``` php
 <?php
--- a/docs/collections.md
+++ b/docs/collections.md
@@ -1,0 +1,246 @@
+# Collections
+
+Robo provides task collections as a means of making error detection and recovery easier. When Robo tasks are added to a collection, their execution is deferred until the `$collection->run()` method is called.  When using collections, a Robo script will go through three phases:
+
+1. Determine which tasks will need to be run, and create a collection.
+  - Assign values to variables.
+  - Do not alter the state of the system.
+2. Create the necessary tasks and add them to the collection.
+  - Pass variables calculated in the first phase to task parameters.
+3. Run the tasks via `$collection->run()`.
+  - Check and report errors once after `run()` returns.
+
+Following this pattern will keep your code linear and easy to understand.
+
+## Basic Collection Example
+
+The 'publish' command from Robo's own RoboFile is shown below:
+
+``` php
+<?php
+class RoboFile extends \Robo\Tasks
+{
+    public function publish()
+    {
+        $current_branch = exec('git rev-parse --abbrev-ref HEAD');
+
+        $collection = $this->collection();
+        $this->taskGitStack()
+            ->checkout('site')
+            ->merge('master')
+            ->addToCollection($collection);
+        $this->taskGitStack()
+            ->checkout($current_branch)
+            ->addAsCompletion($collection);
+        $this->taskFilesystemStack()
+            ->copy('CHANGELOG.md', 'docs/changelog.md')
+            ->addToCollection($collection);
+        $this->taskFilesystemStack()
+            ->remove('docs/changelog.md')
+            ->addAsCompletion($collection);
+        $this->taskExec('mkdocs gh-deploy')
+            ->addToCollection($collection);
+        $collection->run();
+    }
+}
+?>
+```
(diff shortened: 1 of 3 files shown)


DavertMik:
Looks good so far. Please add some guideline on how to make a deploy script using collections. I'd like to see something like Capistrano to be there )

greg-1-anderson:
It might be time-consuming to get into real-world examples at this point. In the short term, I am just going to document the capabilities of Collections here using "hypothetical" tasks. As I evolve real-world Collections use in other projects, though, I could come back here and flesh out the examples with actual tasks used elsewhere, so the documentation can improve and evolve incrementally. Help in this area would be appreciated.

greg-1-anderson:
I'm going out of town for a few days; I think this has reached a Minimum Viable Document state for collections. If you don't find any errors or critical omissions, please merge this in and republish the documentation. We can continue towards more functional examples as time goes along.

docs/collections.md
```diff
@@ -0,0 +1,271 @@
...
+    {
+        $collection = $this->collection();
+        $this->taskFileSystemStack()
+          ->mkdir('work')
+          ->addToCollection($collection);
+        
+        // Add a rollback task
+        $this->taskFileSystemStack()
+          ->rmdir('work')
+          ->addAsRollback();
```
DavertMik: `$collection` should be specified

docs/collections.md
```diff
@@ -0,0 +1,271 @@
...
+?>
+```
+
+In the example above, if the `mkdir('work')` fails, then neither the rollback task nor the 'taskOther' task will be executed.  If it works, though, then `taskOther()` will also be executed; if `taskOther()` fails, then the rollback task will be run, and the 'work' directory will be removed.
+
+### Rollback Methods
+
+Any task may also implement \Robo\Contract\RollbackInterface; if this is done, then its `rollback()` method will be called if the task is `run()` on a collection that later fails.
+
+## Completions
```
DavertMik: I'd name that something like `finialize` to be sound more like  try>catch>final

greg-1-anderson: I'm not opposed to renaming, but a bit worried about confusion if we get too close to the reserved words. Also, completions do not fire if someone throws an exception that is not caught, so avoiding exception terminology might be best here.

docs/collections.md
greg-1-anderson: Good catch. Fixed.

docs/collections.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Temporary Files
+
+Robo also provides an API for creating temporary files. They may be created via `$this->taskTmpFile()`; they are used exactly like `$this->taskWrite()`, except they are given a random name on creation, and are deleted when their collection completes.  If they are not added to a collection, then they are deleted when the script terminates.
+
+### The Temporary Global Collection
+
+Robo maintains a global collection named Temporary that is used to keep track of object temporary objects that are created, but not added to any collection.  These objects are automatically removed when the script terminates.
```
DavertMik: object temporary objects

docs/collections.md
```diff
@@ -0,0 +1,271 @@
...
+?>
+```
+
+In the previous example, the path to the temporary directory is stored in the variable `$work`, and is passed as needed to the parameters of the other tasks as they are added to the collection. After the task collection is run, the temporary directory will be automatically deleted. In the example above, the temporary directory is renamed by the last task in the collection. This allows the working directory to persist; the collection will still attempt to remove the working directory, but no errors will be thrown if it no longer exists in its original location. Following this pattern allows Robo scripts to easily and safely do work that cleans up after itself on failure, without introducing a lot of branching or additional error recovery code.
+
+### Temporary Files
+
+Robo also provides an API for creating temporary files. They may be created via `$this->taskTmpFile()`; they are used exactly like `$this->taskWrite()`, except they are given a random name on creation, and are deleted when their collection completes.  If they are not added to a collection, then they are deleted when the script terminates.
+
+### The Temporary Global Collection
```
DavertMik: I tried to read this chapter 3 times and for me it it is hard to understand the problem being solved.
could you provide an example here?

docs/collections.md
DavertMik: ok, let's keep it as it is

docs/collections.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Basic Collection Example
+
+A simple example is shown below:
+
+``` php
+<?php
+class RoboFile extends \Robo\Tasks
+{
+    function myOperation()
```
DavertMik: I think we can have a use case and real command name for better understanding here

greg-1-anderson:
Okay, I'll try to come up with a very simple real-world example for the basic collection example, and clarify the use of the Temporary global collection.

DavertMik:
Documentation should focus not on theoretical aspects of Collections but on real purpose. So it should starts with scenarios where you will need them for sure. As I said it is deployment in capistrano style:

* connect by ssh
* git clone
* perform migrations
* symlink folders

Probably we should concentrate on this case, if you don't see other good examples. But Robo can be used for deployment and we should extend docs by providing users with the way on how to do it. So this chapter should not be just about collections its about using them in this scenario, however, unlike capistrano highly customizable.

greg-1-anderson:
I definitely agree with [^^](https://github.com/Codegyre/Robo/pull/271#issuecomment-183304929) in concept; I was just planning on implementing things of that nature as I needed them in external projects first, and then circle back to update the documentation.  Maybe I'll try converting Robo's own RoboFile to use collections where applicable, and see if that leads anywhere, example-wise.

greg-1-anderson:
Made some minor changes; still a work-in-progress.

greg-1-anderson:
@DavertMik: Better?

greg-1-anderson:
Rebased with master. Still need a real example using temporary directories.

greg-1-anderson:
Merging per yesterday's chat.  Will return later to enhance temporary directory example.

PR #273, opened 2016-02-15
Decrease coupling between code with Tasklib and PSR-3 logging

greg-1-anderson:
This PR decreases the coupling between code, making it easier to use Robo tasks in other contexts (composer custom installers and other similar tools) without being tied to the Robo Runner.

- Splits Tasklib trait out from Robo\Tasks
- Use Symfony dependency container in Robo\Config
- Create a PSR3 logger

Ultimately, it would be best if tasks only used the logger, so that the calling application had complete control over format.  The best way to do this would be to provide the logger to tasks via constructor injection. This can be done in the task loaders, so existing Robo scripts do not change their calling conventions, but other clients may substitute alternate loggers if desired.

~~Note that this PR uses the Symfony style classes for formatting, which somewhat changes the appearance of Robo output.~~  **UPDATE:** This PR adds custom styling to the PSR-3 logger, so you are no longer tied to the SymfonyStyle styling. Historic Robo formatting (which looks better anyway) is preserved.


greg-1-anderson:
With the latest commits, output now looks the same as it used to.

So far, only the Result class is decoupled from IO; tasks are still using TaskIO directly.

greg-1-anderson:
It was suggested that we might at some point switch from Symfony DI to [PHP-DI 5](http://php-di.org/index.html). This dependency injection container requires PHP 5.6. At the moment, we will stick with Symfony DI, but might switch later once our constructor injection is more consistent.

greg-1-anderson:
Cleaned this up and added a bunch of comments; ready for review.

greg-1-anderson:
One thing I left undone here is I still sort of think that TaskInfo should be split into CommandInfo and TaskInfo.  But not super important.

src/Common/CustomLogStyle.php
```diff
@@ -0,0 +1,154 @@
+<?php
+namespace Robo\Common;
+
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\OutputStyle;
+
+/**
+ * Custom Log Styler. Styles may be overridden via constructor parameters.
+ */
+class CustomLogStyle implements LogStyleInterface
```
DavertMik: Can be renamed to ConsoleLogStyle 
as it is not custom but common class (used by Robo) and available for cli usage

src/Common/CustomLogStyle.php
```diff
@@ -0,0 +1,154 @@
...
+    {
+        return $this->defaultStyles;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function style($context)
+    {
+        $context += ['_style' => []];
```
DavertMik: why the `_` prefix here?

src/Common/CustomLogStyle.php
DavertMik: This should be named like `ConsoleLogStyle` 
As it is common implementation for all Cli loggers.

src/Common/CustomLogStyle.php
greg-1-anderson: We can't predict what variables folks will put in their context, so I added _ as a simple namespacing. We could rename this to something else -- `.style` or whatever. This is one of the classes that I hope to put into Symfony Console.  If this isn't part of Symfony, then of course we could just get rid of it completely and hardcode Robo-styled logging into the base class.

greg-1-anderson: '_style' could also be a constant: `CustomLogStyle::LOG_STYLES` or something like that.

DavertMik: and what is context? How it is shared? How values can be added to it?

DavertMik:
Overall I like the ideas of abstracting output. This should have happened one day, and I really like this day is coming soon. However, I can't get clear understanding of some points:

* why `symfony/dependency-injection` is used. This is a good component, but I'm not sure it suites well for case where you have just 3 services, without need of yaml-xml-php configuration styles, params wiring, and so on. How about Pimple, or [Aura Di](https://github.com/auraphp/Aura.Di) (i like its syntax, btw). Anyway, you still use it more as service locator, real power in DI is not needed in our case.
* requiring `symfony/config` is not clear to me as well. Can it be avoided?
* still can't understand class hierarchy:

Logger extends StyledConsoleLogger extends AbstractLogger

* RoboLogStyle extends CustomLogStyle
* SymfonyLogStyle

At first. Why not to use DI to pass Style into the Logger?

Second, I don't see real purpose of having both StyledConsoleLogger and Logger.
Can they be merged into one ConsoleLogger class?

and yeah... we have a problem that Result prints itself just in constructor and you introduced `LogResultInterface` for that. It looks like we are mixing logging (an interface) with printing actual output. To me `logStopOnFail` sounds a bit strange as it is not related to logging at all but to printing to screen.

### Suggested Improvements

* Replace symfony-di and symfony-config with aura-di, or smth like it. 
* Use DI to inject styles into loggers
* we should have a `ResultPrinter` class inside a container which uses a Logger from it. Yes it will be responsible for printing result, so should contain methods `success`, `error`, `stopOnFail`.
* by making this change merge Logger with StyledConsoleLogger and name it `ConsoleLogger`
* `CustomLogStyle` is not custom actually, we use it for default styles, so I think it's better to be callsed `BaseLogStyle`. Maybe they can be merged as well?
* Move everything Logger-related from Contracts and Common to `Logger` namespace

greg-1-anderson:
Okay, lots of stuff here, and didn't have time to deal with this with SandCamp and post-SandCamp stuff.

ResultPrinter -- I agree. I think that something that derives from SymfonyStyle, to serve as a replacement for folks who are using that for logging in a way that gets them using an actual logger, is a good thing to provide.

Ultimately we should move towards stopOnFail(true) being the default.

Making the Result class a printing class is problematic, but I figured I would break too much existing code if I removed that responsibility.  Eventually, I'd like Robo commands to always end with `return $result;`, and have the Runner responsible for printing the final result. Doing this is good in general, and esp. important for supporting integration into Drush.  If you're ready to do a small amount of b/c breakage and require existing commands all add a `return`, then I'd be happy to do this cleanup now.

Maybe I should separate DI from the initial commit. The main reason it's in here is for future use, but I don't know that DI is likely to be used extensively in small task runners.

I think my next step will be to take all of your Logger advice, and integrate it into the Consolidation\Log interface in [consolidation-org](http://github.com/consolidation-org). Then I can come back here and clean up this PR to use that.

greg-1-anderson:
So, a lot of the mess from before was caused because the SymfonyStyle class was a little difficult to work with; that perturbed the whole design, because I wanted to have both Symfony and component-custom styling supported equivalently.  With a little extra work, I was able to straighten out the way the logging classes used SymfonyStyle, which cleaned everything up and made it simple to use DI to inject the style classes.

Not everything from the feedback above is done, but a lot of it is. The main logging classes are now in the [consolidation/log](https://github.com/consolidation-org/log) project.

Would you be interested in seeing a commit with the responsibility for printing removed from the Result class, requiring commands to all `return $result;`, or would you prefer that it stay as-is for now for backwards compatibility? If the later, I can definitely put the ResultPrinter into its own class that HasA Logger rather than IsA Logger.

greg-1-anderson:
Actually, the impact of removing printing from the Result constructor is that every TaskInterface::run() implementation would have to start printing its result -- this is not just limited to the final Result from the command.

In Drush, functions that roughly serve the same function as tasks must `return drush_set_error(...);` when failing with an error. This also logs the result. In Robo, something akin to `$this->reportResult($result);` or printResult or logResult, as you will, could serve a similar purpose.  For what it is worth, this operation -- displaying the result of one task's operations -- is very much logging, not merely output printed to the screen.  To see what I mean, imagine the same operations executed from a server, with no attached console.  In this instance, you would want to see these messages in the log, along with the other log items.

This is already a big PR, though, so for now I am going to refrain from removing printing from the Result constructor. Would really like to see that happen soon, but it would add too many additional changes to this already large PR, so let's finish up here and discuss the timing of that later.

What I will do next:

- I will move the printing from the Result constructor into a ResultPrinter that the Result constructor will retrieve via a Config method.
- I will leave the DI container as it is for now. As a follow-on PR, I'd like to see if we can decouple things sufficiently that the task loaders are providing needed objects for the tasks via constructor + parameter injection, and they are getting all of said objects themselves from constructor + parameter injection. Then we could separate the DI container from all of the task implementations, which would allow Drush (which would probably prefer to use Symfony DI for alignment with Drupal 8) to use a different DI container than Robo.
- I think `symfony/config` was added by mistake. I'll try taking it out and see what happens.
- I will move the remaining Logger classes and contracts to Robo\Log.

I think that all of your other comments were addressed by the Logger refactor, but let me know if you still have questions about any of that part.

greg-1-anderson:
The ResultPrinter is done. Method naming was a little different than suggested. Both Result and LoggerInterface contain `success` and `error` methods. Since these classes are quite different, this is not really a problem in my mind. However, adding `success` and `error` methods to ResultPrinter, and have them behave completely differently than the methods of the same name in Result and LoggerInterface, both of which this class is closely coupled to, seemed really confusing to me.  Went with `printStopOnFail` for consistency. Naming adjustments still possible here.

greg-1-anderson:
Squashed and rebased with master, and continued at #283.

PR #275, opened 2016-02-18
Source and destination of the rsync command needs to be escaped

webflo:
The rsync fails if the source or destination path contains spaces.


greg-1-anderson:
Thanks for this; merged.

PR #279, opened 2016-02-26
add atoum task

agallou:
This task allows to run [atoum](http://atoum.org/) tests inside robo.

```php
$this->taskAtoum()
	->debug()
	//->tags("databaseTest")
	//->lightReport()
	//->tap()
	//->files(["path/to/test1.php", "path/to/test2.php"])
	->directories('tests/units')
	//->configFile("config/dev.php")
	//->debug()
	->run();
```

### DIFF
--- a/src/Task/Testing/Atoum.php
+++ b/src/Task/Testing/Atoum.php
@@ -1,0 +1,182 @@
+<?php
+namespace Robo\Task\Testing;
+
+use Robo\Contract\CommandInterface;
+use Robo\Contract\PrintedInterface;
+use Robo\Task\BaseTask;
+
+/**
+ * Runs [atoum](http://atoum.org/) tests
+ *
+ * ``` php
+ * <?php
+ * $this->taskAtoum()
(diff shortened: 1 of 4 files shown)


src/Task/Testing/Atoum.php
```diff
@@ -0,0 +1,170 @@
+<?php
+namespace Robo\Task\Testing;
+
+use Robo\Contract\CommandInterface;
+use Robo\Contract\PrintedInterface;
+use Robo\Task\BaseTask;
+
+/**
+ * Runs atoum tests
```
DavertMik: Please put all documentation here, the same way it is done for other tasks like PHPUnit.
Real docs will be generated from it

DavertMik:
Thanks, nice addition. Could you update docblock for the task with documentation? It is necessary because all *.md docs are generated from docblocks

agallou:
@DavertMik docblocks updated and doc generated using the "robo docs" task

DavertMik:
Thank you!

PR #280, opened 2016-02-27
Feature/task changelog with handler

llaville:
Ability to customize our own changelog render with a callback.

Here is an example of new render:

```php
<?php
$closure = function ($changelogTask) {

    $anchor    = "# Change Log";
    $noversion = "## [Unreleased]";
    $filename  = $changelogTask->getFilename();
    $version   = $changelogTask->getVersion();
    
    $text = implode("\n", $changelogTask->getChanges());
    $ver  = "## \[{$version}\] - (\d+)-(\d+)-(\d+)\n\n";

    if (!file_exists($filename)) {
        //$changelogTask->printTaskInfo("Creating {$filename}");

        $contents  = $anchor;
        $contents .= "\n\n";
        $contents .= "All notable changes to this project will be documented in this file.\n";
        $contents .= "This project adheres to [Semantic Versioning](http://semver.org/),\n";
        $contents .= "using the [Keep a CHANGELOG](http://keepachangelog.com) principles.\n\n";
        $contents .= $noversion;
        $contents .= "\n\n";

        $res = file_put_contents($filename, $contents);
        if ($res === false) {
            return Result::error($changelogTask, "File {$filename} cant be created");
        }
    }

    // trying to append to changelog
    $result = (new Replace($filename))
        ->regex("~$ver~")
        ->to("\${0}$text\n")
        ->run();

    if (!$result->getData()['replaced']) {
        if (stripos('Unreleased', $version) === false) {
            $to = "## [{$version}] - " . date('Y-m-d') . "\n\n"
                . $text . "\n\n"
                . $noversion . "\n\n"
            ;
        } else {
            $to = $noversion . "\n\n" . $text . "\n";
        }

        $result = (new Replace($filename))
            ->from($noversion . "\n\n")
            ->to($to)
            ->run();
    }
    
};
                
$this->taskChangelog()
    ->handler($closure)
    ->version($version)
    ->askForChanges()
    ->run();                
```

### DIFF
--- a/src/Task/Development/loadTasks.php
+++ b/src/Task/Development/loadTasks.php
@@ -9,7 +9,7 @@ trait loadTasks
      */
     protected function taskChangelog($filename = 'CHANGELOG.md')
     {
-        return new Changelog($filename);
+        return new Changelog($filename, $this);
     }
 
     /**
@@ -65,4 +65,4 @@ trait loadTasks
     {
         return new OpenBrowser($url);
     }
-} 
+}
--- a/docs/tasks/Development.md
+++ b/docs/tasks/Development.md
@@ -26,6 +26,24 @@ $this->taskChangelog()
 ?>
 ```
 
+Changes may be formatted into a custom file format. Handler can be either a function,
+a public method or a closure.
+
+``` php
+<?php
+$handler = function ($changelogTask)
+{
+    // how to manage contents of your CHANGELOG file
+}
+
+$this->taskChangelog()
+ ->handler($handler)
+ ->version($version)
+ ->askForChanges()
+ ->run();
+?>
+```
+
 * `Development\Changelog filename(string $filename)` 
 * `Development\Changelog anchor(string $anchor)` 
 * `Development\Changelog version(string $version)` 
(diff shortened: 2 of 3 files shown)


DavertMik:
Thanks, that's really helpful

Could you add documentation and example of usage into docblock of Changelog task?

llaville:
Two examples provided. Last way to use it is inside a RoboFile class.

src/Task/Development/loadTasks.php
```diff
@@ -9,7 +9,7 @@
      */
     protected function taskChangelog($filename = 'CHANGELOG.md')
     {
-        return new Changelog($filename);
+        return new Changelog($filename, $this);
```
DavertMik: Is that really needed to bind to class? 
In `GenMarkdown` we also use closures everywhere but they are not bound to anything. I don't see any real necessity in this but it adds additional dependency to a task, so I don't see as the most clear solution

DavertMik:
Thanks, looks nice. But is that possible to remove binding to a class?

DavertMik:
Sorry this branch is no longer can be merged automatically. If it is still valid please resend

PR #282, opened 2016-03-04
make excludeVcs() recursive

boedah:
I just noticed that the excludeVcs() did not exclude git submodules.
So I added an option for git, svn, hg each.


boedah:
Initially I added '*/.git' (and similiar to the others) to the excluded items, but simply leaving off the trailing slash leads to the same results (exluding at any depth).
So amended the first commit.

boedah:
adjusted RsyncTest ;)

DavertMik:
Thanks!

PR #283, opened 2016-03-05
TaskLib, PSR-3 logger, and interim DI container.

greg-1-anderson:
Split off TaskLib from Robo\Tasks. Add a Dependency Injection container, and a PSR-3 logger based on Consolidation\Log.

Continuation of #273. Squashed and rebased with master.


greg-1-anderson:
Regarding which DI Container to use, see [How to Configure Monolog to Display Console Messages](http://symfony.com/doc/current/cookbook/logging/monolog_console.html) as an example. We want to start doing some bridging with existing code based on Symfony Console, so we'll want to make it easy to do things like `$logger = $this->getContainer()->get('logger');` (or similar, for other services). I've mentioned this desire for interoperability with Symfony DI in Drush for Drupal 8, but wanted to put in this link as another example.

I further improved the decoupling of the code in #284; although there are still a couple of backwards-compatibility-breaking changes needed to fully separate tasks and all task implementations from the DI container and Config class, that PR and this one gets Robo really close. If the changes here & there are okay, next steps are largely dependent on what we want to do vis-a-vis versioning and backwards compatibility.

DavertMik:
looks like this PR can be closed now, right?

greg-1-anderson:
This code should have been merged into master when you merged task-assembly, since I branched that PR off of this one. I may have rebased in the middle, confusing GitHub. I'll confirm the state of master and close here.

greg-1-anderson:
This has been merged.

PR #284, opened 2016-03-07
Task assembly

greg-1-anderson:
This PR introduces the concept of task assembly, which is done by way of the TaskAssembler, a helper object that helps do dependency injection on tasks that are created at runtime (as opposed to during container initialization, where the DI container can be used for this purpose).

The old way to build tasks looks like this:
```
    protected function taskExtract($filename)
    {
        return new Extract($filename);
    }
```
With task assembly, it now looks like this:
```
    protected function taskExtract($filename)
    {
        return $this->taskAssembler()->assemble(
            '\Robo\Task\Archive\Extract',
            [$filename]
        );
    }
```
The benefit of using the task assembler is threefold:

1. It will call `setLogger` & c. as necessary on the task, after it is constructed, reducing coupling between the different areas of the code.
3. It recognizes tasks that implement CompletionInterface, and ensures their `complete()` method will be called at shutdown time, if the task is `run()`.
2. It also supports 'simulated' tasks, which print out what they _would_ do when `run()` without actually making any persistent changes.

This PR does not break backwards compatibility with existing external Robo tasks. Existing tasks that continue to use the old builder pattern will still work; however, they will silently ignore the `--simulate` flag, and execute instead. To break backwards compatibility and force old tasks to update, all that needs to be done is [remove one of the last references to the Config class](https://github.com/Codegyre/Robo/pull/284/files#diff-eac88c5051967fcde94bef8be35b9116R29), and then all task printing operations will fail for tasks that have not updated to the new mechanism.


src/Task/Development/loadTasks.php
```diff
@@ -36,7 +45,10 @@ protected function taskSemVer($pathToSemVer = '.semver')
      */
     protected function taskServer($port = 8000)
     {
-        return new PhpServer($port);
+        return $this->taskAssembler()->assemble(
```
greg-1-anderson: I don't see what's wrong about the formatting of this line. Looks the same as all of the others like it. No tabs that I can see. Odd.

greg-1-anderson:
The accessor methods added here were showing up as commands, so I added a filter to automatically remove accessors from command lists. It would also be possible to add an annotation to opt-out of command selection.

greg-1-anderson:
It also seems that we could replace task assemblers with direct use of a DI container. This would be particularly convenient if we used a container that supported [inflectors, such as League\Container](http://container.thephpleague.com/inflectors/).

Doing this would substantially change the syntax of Robo, though; `$this->taskExec()` would become `$container->get('taskExec')`.  I don't know if that would be a definite improvement, but I thought I'd mention it as a possibility. Maybe defer until later, and keep the proposed implementation in this PR to maintain compatibility? I could go either way.

greg-1-anderson:
See the [discussion in the Drush issue queue](https://github.com/drush-ops/drush/pull/2018#issuecomment-194398806) on the thought process I went through to select a proposed DI container for Drush.  Aurora DI could work; it's syntax is fine, but I didn't really like the names of its methods. I like league/container quite a bit more than Symfony DI.

greg-1-anderson:
Switched to league/container for DI.  The inflector feature allows us to automatically inject a logger into any object (managed by the DI container) simply by having it implement LoggerAwareInterface and use LoggerAwareTrait. This gives us the convenience of auto-wiring, but in a more controlled, declarative way.

composer.json
```diff
@@ -21,7 +21,7 @@
         "symfony/console": "~2.5|~3.0",
         "symfony/process": "~2.5|~3.0",
         "symfony/filesystem": "~2.5|~3.0",
-        "symfony/dependency-injection": "~2.5|~3.0"
+        "league/container": "~2"
```
DavertMik: good choice )

src/Application.php
```diff
@@ -9,14 +9,70 @@
...
+        if ($roboTasks instanceof \Robo\Tasks) {
+            $roboTasks->setTaskAssembler($this->taskAssembler);
+        }
 
-        $commandNames = array_filter(get_class_methods($className), function($m) {
+        $commandNames = array_filter(get_class_methods($className), function ($m) {
             return !in_array($m, ['__construct']);
         });
+        $accessorNames = static::findAccessorMethods($commandNames);
+        $commandNames = array_diff($commandNames, $accessorNames);
```
DavertMik: I'm not sure why you need `findAccessorMethods` method actually. 
If you need to ignore getters in setters in command list, just ignore commands starting from `get` and `set` prefexes (but take camelCase into account `setUp` - ignored, but `setup` added as command)

src/Application.php
```diff
@@ -9,14 +9,70 @@
...
+    {
+        $this->taskAssembler = $taskAssembler;
+    }
+
+    public function taskAssembler()
+    {
+        return $this->taskAssembler;
+    }
+
+    protected static function findAccessorMethods($commandNames)
```
DavertMik: I don't think this method is really needed. See notes below

src/Config.php
```diff
@@ -41,18 +42,19 @@ public static function createContainer($input = null)
...
+        $container->share('logger', 'Robo\Log\RoboLogger')
+            ->withArgument('output')
+            ->withMethodCall('setLogOutputStyler', ['logStyler']);
+        $container->share('resultPrinter', 'Robo\Log\ResultPrinter');
+        $container->share('taskAssembler', 'Robo\TaskAssembler');
+
+        $container->inflector('Psr\Log\LoggerAwareInterface')
+            ->invokeMethod('setLogger', ['logger']);
 
         return $container;
```
DavertMik: btw, I think League's syntax is much cleaner

src/Config.php
```diff
@@ -41,18 +42,19 @@ public static function createContainer($input = null)
...
-            ->register('logger', 'Robo\Log\RoboLogger')
-            ->addArgument(new Reference('output'))
-            ->addMethodCall('setLogOutputStyler', array(new Reference('logStyler')));
-        $container
-            ->register('resultPrinter', 'Robo\Log\ResultPrinter')
-            ->addArgument(new Reference('logger'));
+        $container = new Container();
+
+        $container->add('input', $input);
+        $container->share('output', 'Symfony\Component\Console\Output\ConsoleOutput');
```
DavertMik: let's use `::class` definitions instead of strings, as we are not going to support PHP 5.4

src/Runner.php
```diff
@@ -80,18 +80,15 @@ public function execute($input = null)
...
-        // If we were not provided with a container, then create one
+        // If we were not provided a container, then create one
         if (!Config::hasContainer()) {
             $input = $this->prepareInput($input ? $input : $_SERVER['argv']);
             $container = Config::createContainer($input);
             Config::setContainer($container);
-
-            // Note: this freezes our container, preventing us from adding any further
-            // services to it.
-            $container->compile();
```
DavertMik: thanks for removing it :) I really hate to compile anything )

src/TaskAssembler.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public function assemble($taskClass, $taskConstructorParameters = [])
+    {
+        $class = new \ReflectionClass($taskClass);
+        $task = $class->newInstanceArgs($taskConstructorParameters);
+
+        // TODO: Would be more rigorous to have an interface for this.
+        if (method_exists($task, 'setLogger')) {
+            $task->setLogger($this->logger);
```
DavertMik: BaseTask can have this method as abstract, or implement LoggerAwareInterface

src/TaskSupport.php
```diff
@@ -0,0 +1,19 @@
+<?php
+namespace Robo;
+
+// Always `use \Robo\TaskSupport` if bringing in any other Robo
+// tasks via its loadTasks trait.
+trait TaskSupport
```
DavertMik: this name doesn't say anything...

DavertMik:
I really like your choice of container

But I think If we choose it, maybe we can use it in full and not include task assembler class. The overall concept seems like factory method, so it can be moved to the container.

> Doing this would substantially change the syntax of Robo, though; $this->taskExec() would become $container->get('taskExec'). I don't know if that would be a definite improvement, but I thought I'd mention it as a possibility.

Yes, this will be much cleaner if container would manage creation of tasks, and inject everything. Probably you can use http://container.thephpleague.com/factory-closures/ to implement `simulated` behaviors.

Also I'd ask you to use `::class` constant instead of strings in class names. They really save ass, in refactorings, and they really look nicer.

And thanks for your work. I really like the way it is moving, but I'm still desore to find a way to keep things simpler )

src/TaskAssembler.php
greg-1-anderson: Yeah, I'll switch to use LoggerAwareInterface

src/Config.php
greg-1-anderson: kk, done.

src/Runner.php
greg-1-anderson: League doesn't support compiling containers, so it was easy for me to make the right choice here. :)

greg-1-anderson:
So, there's a problem with using the DI container instead of taskFoo().  A lot of Robo tasks take parameters, e.g. `$this->taskExec($command)->run();`, whereas the container `get()` method does not allow constructor parameters to be passed in to the created object (except for parameters that are invariant throughout the program run, and provided when the container is initialized).

So, to move forward here, we would need to change many tasks, e.g. to `$container->get('taskExec')->command($command)->run();`.  Of course, a lot of changes are already required, so perhaps this isn't too much of an extra leap, but wanted to check in before going ahead and doing that everywhere. Backwards compatibility pretty much going away with this change.

greg-1-anderson:
There's another problem with factory closures (or any other technique we'd use to build the object from the container). The issue is that if we wrap our task in `Temporary` or `Simulated` in the factory, than `instanceof` on the wrapped object will no longer be true for interfaces such as LoggerAwareInterface, so none of the inflectors will work.

The only way to keep the inflectors working is to return the native, not-wrapped task instance from the factory.  We could apply our wrappers after the fact if we created our own RoboContainer class that derives from \League\Container\Container. Then, in the `get` method, we could call parent::get() (calls the factory and applies the inflectors), and then wrap the resulting task in `Temporary` or `Simulated` as appropriate.

This seems reasonable, if you don't mind having our own Container subclass.

DavertMik:
> This seems reasonable, if you don't mind having our own Container subclass.

Sure. Please proceed

greg-1-anderson:
The latest commits remove all of the loadTask traits, and replace them with League service providers that register tasks on demand, as they are used. Note that the tests only register providers for the services they are testing, increasing test isolation. We are also less dependent on generated code in the tests now as well.

Surprisingly enough, this code is still largely backwards compatible with older Robo code.  External Robo tasks that are still defined via loadTask traits should still work, although they will not benefit from the inflectors.  This may cause them to break in the future, as we become more dependent upon inflectors.  RoboFiles that use $this->taskFoo() will continue to work as written, though, thanks to a __call() function in Robo\Tasks that converts to the new preferred form: `$this->getContainer()->get('taskFoo', [$a, $b]);`

greg-1-anderson:
In the last commit, I change the new preferred form from:

`$this->getContainer()->get('taskFoo', [$a, $b]);`

to:

`$this->task('Foo', $a, $b);`

Compare this to the original API of `$this->taskFoo($a, $b);`. The new preferred form is pretty close, and benefits greatly from the DI container and inflectors.

greg-1-anderson:
My PR was already merged at https://github.com/thephpleague/container/pull/92

:smile:

src/Task/Assets/loadTasks.php
```diff
@@ -1,41 +0,0 @@
...
-
-trait loadTasks
-{
-    /**
-    * @param $input
-    * @return Minify
-    */
-    protected function taskMinify($input)
-    {
-        return new Minify($input);
```
DavertMik: this can be rewritten as:

```
return $this->task('minify');
```

and what's important we have docblock which states the type of return object. SO that makes tasks loaded with this method to be written easily

src/Task/Base/ServiceProvider.php
```diff
@@ -0,0 +1,24 @@
+<?php
+namespace Robo\Task\Base;
+
+use League\Container\ServiceProvider\AbstractServiceProvider;
+
+class ServiceProvider extends AbstractServiceProvider
+{
+    protected $provides = [
```
DavertMik: btw, why do you need $provides?

greg-1-anderson: $provides is required by league/container's service provider.  The 'register' function isn't called until someone calls 'get' on a provided service.

src/Task/Assets/loadTasks.php
greg-1-anderson: At the moment, I have these registered as 'taskMinify', and `task` adds the "task" part on the beginning (a sort of container namespacing). So, under this naming convention, this would be `return $this->task('Minify');` == `$container->get('taskMinify');`.  Could change this to `return $this->task('minify');` == $container->get('minify');` or whatever you think is best; no strong opinions here ATM.

src/Task/Assets/ServiceProvider.php
```diff
@@ -0,0 +1,22 @@
...
+        'taskScss',
+    ];
+
+    public function register()
+    {
+        $this->getContainer()->add('taskMinify', Minify::class);
+        $this->getContainer()->add('taskImageMinify', ImageMinify::class);
+        $this->getContainer()->add('taskLess', Less::class);
+        $this->getContainer()->add('taskScss', Scss::class);
+    }
```
DavertMik: I have an idea to simplify that declaration.
Let's create an abstract class

```php
class TaskProvider extends AbstractServiceProvider 
{
protected $tasks = [];

function provides($service)
{
     if (array_key_exists($service, $this->tasks)) {
         return $this->getContainer()->get($service);
     }
}

//
function register() 
{
     foreach ($this->tasks as $name => $class) {
         $this->getContainer()->add($name, $class);
     }
}
```

if we do that change, provider class changes just to:

```php
class AssetTaskProvider extends TaskProvider
protected $tasks = [
    'taskMinify' => Minify::class,
    // ....
];
```

greg-1-anderson: Yeah, that's a good idea.

greg-1-anderson: We also need to check to see if $service is null, and return array_keys($tasks) in this instance.

greg-1-anderson:
The latest technique does not quite work due to a limitation in League/Container.  I have [submitted a PR to rectify the situation](https://github.com/thephpleague/container/pull/93). I suspect the maintainer might want to see the strictly backwards-compatible implementation of this.

greg-1-anderson:
Submitted a backwards compatible version: https://github.com/thephpleague/container/pull/94

philipobenito:
Hope this is agreeable https://github.com/thephpleague/container/pull/94#issuecomment-197763210

greg-1-anderson:
Thanks, @philipobenito! Works great & is totally awesome. This will be great for both Robo and Drush.

philipobenito:
Brilliant stuff, I'll get you a tag sorted in the next few mins.

greg-1-anderson:
@DavertMik this is now using the tagged release of league/container. I think this should now meet all of your expectations per our last chat -- and, against all odds, even maintains bc (although --simulate won't work for folks who keep using the deprecated bc loadTasks conventions). Take a look and let me know. Excited to see this merged soon!

greg-1-anderson:
Not captured in the discussion above (from chat):  we want to preserve loadTasks for parameter type hinting, and autocomplete in IDEs.  However, now that we have combined the service provider into the loadTasks file, it isn't so bad to have both. The pattern for making new tasks is pretty apparent from the existing examples; I'll also update the documentation on this.

DavertMik:
Awesome :heart: :heart: :heart:

PR #285, opened 2016-03-10
Standalone Robo scripts.

greg-1-anderson:
This PR allows the user to write standalone Robo scripts that can be executed like other shell scripts. Both multi-command (./myscript mycommand arg) and single-command (./myscript arg) variants are supported.

Requires that `robo` be in your `$PATH`.

### DIFF
--- a/script.robo
+++ b/script.robo
@@ -1,0 +1,35 @@
+#!/usr/bin/env robo
+<?php
+
+/**
+ * Standalone Robo script.
+ *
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I added -f as a synonym for --load-file.

DavertMik:
Wow, that's awesome. Please merge if it is ready )

greg-1-anderson:
Thanks!

PR #287, opened 2016-03-11
add Exec::getProcess() which returns the internal Symfony Process instance

### LINKED ISSUE #286: No access to error output of ExecTask
When running commands via ExecTask, a Result object is created with:

```
return new Result($this, $this->process->getExitCode(), $this->process->getOutput(), ['time' => $this->getExecutionTime()]);
```

There is no way for the caller to get access to the error output of the process.

boedah:
The easy, versatile and future-proof way was to just return the Process instance itself (and e.g. not give the Result object another parameter for error output).

This fixes #286

### DIFF
--- a/src/Task/Base/Exec.php
+++ b/src/Task/Base/Exec.php
@@ -154,6 +154,17 @@ class Exec extends BaseTask implements CommandInterface, PrintedInterface
         return Result::success($this);
     }
 
+    /**
+     * Returns the Symfony process instance.
+     * If `run()` has not been called before, NULL is returned.
+     *
+     * @return Process
+     */
+    public function getProcess()
+    {
+        return $this->process;
+    }
+
     static function stopRunningJobs()
     {
         foreach (self::$instances as $instance) {
--- a/tests/unit/Task/ExecTaskTest.php
+++ b/tests/unit/Task/ExecTaskTest.php
@@ -41,6 +41,21 @@ class ExecTaskTest extends \Codeception\TestCase\Test
         verify($this->taskExec('ls')->getCommand())->equals('ls');
     }
 
+    public function testGetProcessReturnsNullIfNotStarted()
+    {
+        verify($this->taskExec('ls')->getProcess())->equals(null);
+    }
+
+    public function testGetProcessReturnsProcessInstanceIfStarted()
+    {
+        $result = $this->taskExec('ls')->run();
+
+        /** @var \Robo\Task\Base\Exec $task */
+        $task = $result->getTask();
+        $process = $task->getProcess();
+        $this->assertInstanceOf('Symfony\Component\Process\Process', $process);
+    }
+
     public function testExecStack()
     {
         $this->taskExecStack()


greg-1-anderson:
+1

DavertMik:
It looks like it can be added to a task result not to task itself: 

```
Result::error($this, message, ['process' => $this->process])
```

greg-1-anderson:
I guess the big question is, what is the use-case for using taskExec vs. just using Symfony Process directly?  taskExec is good for chaining multiple operations together in a collection; if you need access to the native Process object, though, maybe it would be better to just create a Symfony Process object directly. To this end, I might suggest it might be more convenient to just put stderr in the Result object's data, and not expose it to the caller at all.

boedah:
A couple of thoughts:

* stderr output is not only needed in case the process failed (exit code !== 0), as programs also print diagnostics information to this stream (e.g. drush to my distress ;)
* the `Exec` class constructs `Result` objects in 4 places (and in 3 different ways) in the `run()`-method: https://github.com/Codegyre/Robo/blob/master/src/Task/Base/Exec.php#L135,L154
```
        return new Result($this, $this->process->getExitCode(), $this->process->getOutput(), ['time' => $this->getExecutionTime()]);
        return Result::error($this, $e->getMessage());
        return Result::success($this);
```
* so in case the process is started in the background, the output seems not to be available at all (https://github.com/Codegyre/Robo/blob/master/src/Task/Base/Exec.php#L154)
* we would have to feed the error output into `data` in all 4 places
* the `Result` object
  * the namespace (`\Robo\Result`) suggests this class is generic, but I think, this is not quite the case. As it has e.g. the `getExitCode` method, so this assumes a Robo task consists of one command line call which has an exit code
  * `printError()` currently does not care about `stderr`, and just prints out whatever was given to the object in `message`
  * I would rather have a defined API instead of pumping more stuff to the object's `data`
* I think telling users to create a `Process` themselves to get something basic as the error output should be no option
  
So if you dislike giving access to the `Process` object (which I understand), how about this:

* give `Result` a `{set,get}ErrorMessage()` method (instead of bloating the constructor or putting random strings into `data`)
* in `Exec`: extract a method to create a `Result` object (so this happens only in one place)

And afterwards as extra PR:

* make `Result->printError()` more useful with also printing error output

Could make both, I you like.

greg-1-anderson:
While I am not disputing the need for access to stderr in the general case, if you are calling Drush programmatically, then you should be using backend invoke, via drush_invoke_process. This will give you access to both the structured result object and the structured log entries. Ask in the Drush queue, or on Drupal Stack Exchange if you have any trouble with this.

boedah:
Thanks.
I am not within a Drupal context but in a standalone management script which needs to target 40+ sites.
While using `drush_invoke_process` would certainly work somehow, it would be by far overly complex compared to a simple `getErrorMessage()` call...

greg-1-anderson:
You can always use Symfony Process directly if you need more control.

PR #288, opened 2016-03-16
Changed increment method to following semantic versioning specification

timtegeler:
I think the existing implementation of increment does not follow the semantic versioning specification:

- Increment of major must reset minor and patch to 0.
- Increment of minor must reset patch to 0.

Have a look at:
http://semver.org/#spec-item-7
http://semver.org/#spec-item-8


DavertMik:
Thanks!

Issue #258, opened 2016-01-11
Document coding standards for Robo project


greg-1-anderson:
Robo follows PSR-2 coding conventions, and these are now enforced by Nitpick-CI. These are both good things, but php-cs-fixer does not produce code that fully satisifes Nitpick-CI.

It would be very helpful to have documentation on which tool + settings produces the desired results. It would be nice if default settings for php-cs-fixer did the trick, but that's optional.

DavertMik:
yes, I'd like to have it PSR-2 
by myself I'm using PHPStorm without php-cs. But I think you can configure it to follow PSR2 standard.

boedah:
:+1: for php-cs-fixer satisfying Nitpick-CI...

greg-1-anderson:
Fixed by #289. See also #292

greg-1-anderson:
Re-opening, perhaps briefly, for some further discussion.

I ran Robo through the code beautifier, and Nitpick CI did not complain, so I committed the fixes.  I also fixed a bunch of phpsc complains manually.  There are still a few remaining reported issues.  There are a bunch of lines that are > 120 characters; I did not feel that this was critical to fix everywhere.

phpcs also complains "Method name "_copy" should not be prefixed with an underscore to indicate visibility" in FilesystemStack. This method is not prefixed to indicate visibility; `copy()` and `_copy()` are pairs, with the former being used to queue up the copy operation, and the later being used to run it. The execution method is handled automatically unless there are parameters with default values, in which case the underscore-prefixed method is needed. There was some discussion a while back about naming conventions here, and my preference (to keep "_copy" et. al.) stood. Don't know if anyone might want to revisit this in light of the phpcs warnings; Nitpick CI does not care, and this is not actually a PSR-2 violation -- phpcs just cannot tell if the purpose of the underscore is to indicate visibility or not. We could potentially just make these methods 'public', but phpcs still complains in this case (which is probably a phpcs bug).

Finally, there is a warning about the side-effects at the end of the Exec class. I think these might be replaced by implementing CompletionInterface. Robo will make sure the complete method is called at shutdown. We need to make sure that this also happens on a signal, not just on shutdown. We might also need to add a parameter to complete to differentiate between an ordinary collection completion and program termination.

greg-1-anderson:
This has been done. Thanks, @dustinleblanc

dustinleblanc:
:+1: glad to give a little help, hope to give more!

PR #289, opened 2016-03-17
Code standards

dustinleblanc:
@DavertMik this is just a first stab at #258 

I figure you may have some different preferences on how to handle this but these are my opinions on a fairly simple way to establish the coding standards and provide a little bit of help for folks to easily follow them. I am not sure if the Sniffer's implimentation of PSR-2 will have similar issues to those @greg-1-anderson mentioned. When I ran the sniff on the entire src directory it threw up a lot of issues so it is possible it is misconfigured out of the box. I like that this can just be included in dev dependencies and most of it is self-contained with the project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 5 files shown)


DavertMik:
Thank you. Robo was started as my personal project and I was using my preferred code style. But sitll I knew that this day will come and I will need to switch coding standards and follow them. So let's do it!

PR #291, opened 2016-03-20
Restore rsync test backed out in previous merge.

greg-1-anderson:
This PR restores an rsync test that I backed out when merging the DI and PSR-logger stuff.  It looks like an attempt to handle spaces in filenames (e.g. for Windows), so perhaps there was some rsync commit that was also lost somewhere along the way. This bears some investigation.


greg-1-anderson:
The merge of #283 / #284 somehow backed out a number of changes that had been made in parallel as those PRs were in development. I am working through re-applying the overwritten changes.

PR #292, opened 2016-03-20
Fix sniffer

greg-1-anderson:
Follow-on to #289.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -356,8 +356,11 @@ class RoboFile extends \Robo\Tasks
      */
     public function sniff($file = 'src/', $options = ['autofix' => false])
     {
-        $this->taskExec("./vendor/bin/phpcs --standard=PSR2 {$file}")->run();
-        if ($options['autofix'] == true) {
+        $result = $this->taskExec("./vendor/bin/phpcs --standard=PSR2 {$file}")->run();
+        if ((!$options['autofix']) && (!$result->wasSuccessful())) {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This does not work perfectly; the exit code from phpcs is 0 when no errors or warnings are found, and 1 if there are fixable or non-fixable errors. The result of this is that we prompt the user and ask if they'd like to fix the sniffer errors whether or not any of them are fixable.

greg-1-anderson:
I don't really have time to address this myself at the moment, but I opened https://github.com/squizlabs/PHP_CodeSniffer/issues/930 to track the status code issue.

RoboFile.php
```diff
@@ -352,14 +352,13 @@ public function trySuccess()
      *
      * @param string $file
      *    A file or directory to analyze.
-     * @param bool $autofix
-     *    Whether to run the automatic fixer or not.
+     * @option $autofix Whether to run the automatic fixer or not.
      */
-    public function sniff($file = 'src/', $autofix = false)
+    public function sniff($file = 'src/', $options = ['autofix' => false])
```
dustinleblanc: :+1:

greg-1-anderson:
I added a --strict option, and defaulted the 'sniff' command to show only errors.  Now, `robo sniff` does not report any problems when run against the Robo source code.

greg-1-anderson:
It is really useful to have `./robo sniff` run cleanly, so I am going to merge this. We can continue to adjust in follow-on issues as needed.

PR #294, opened 2016-03-20
Added optional parameter message to git tag

timtegeler:
I added an optional parameter message to the method tag from GitStack. 

Now I can tag with a message like this:
`$this->taskGitStack()->tag($semVer, $message)->run();`

Previously I needed to do it manually with Exec:
`$this->taskExec('git tag -m "' . $message . '" ' . $semVer)->run();`


DavertMik:
Thanks!

Issue #295, opened 2016-03-21
Tasks shouldn't call ask()


greg-1-anderson:
See [the Scrutinizer report for GitHubRelease](https://scrutinizer-ci.com/g/Codegyre/Robo/inspections/76927144-a2d2-4989-9c7d-6f5d1032614c/code-structure/operation/Robo%5CTask%5CDevelopment%5CGitHubRelease%3A%3Arun).

When I reduced the coupling between the Robo task classes, I removed the IO trait from BaseTask.  Tasks that still need to do things such as call 'ask' may `use IO` on their own; however, it is better if the command does all of the asking, and passes the result as parameters to the task.

On a related note, the DynamicParams trait is useful for eliminating some typing when putting together a task, but the existence of a `call()` method makes things a little harder on the code analyzer, making errors look like warnings. Perhaps we could maintain this trait for use by third party tasks, but avoid using it in Robo core.

Waiting for feedback before taking action on any of this, though.

greg-1-anderson:
Here are some things that still need to be fixed:
```
src/Task//Base/Exec.php: *  $this->say('tests passed');
src/Task//Base/ParallelExec.php:        $progress = new ProgressBar($this->getOutput());
src/Task//Base/ParallelExec.php:                        $this->getOutput()-> [various]
src/Task//Base/SymfonyCommand.php:            $this->command->run(new ArrayInput($this->input), $this->getOutput())
src/Task//Development/Changelog.php:        while ($resp = $this->ask("Changed in this release: ")) {
src/Task//Development/GitHub.php:        self::$user = $this->ask('GitHub User');
src/Task//Development/GitHub.php:        self::$pass = $this->askHidden('Password');
src/Task//Development/GitHub.php:            $this->askAuth();
src/Task//Development/GitHubRelease.php:        $this->name = $this->ask("Release Title");
src/Task//Development/GitHubRelease.php:        $this->body .= $this->ask("Description of Release\n") . "\n\n";
src/Task//Development/GitHubRelease.php:        while ($resp = $this->ask("Added in this release:")) {
src/Task//Development/PackPhar.php:        $progress = new ProgressBar($this->getOutput());
src/Task//Development/PackPhar.php:        $this->getOutput()->writeln('');
```
Notes:
- Use TaskIO for output, not `say` / `getOutput`
[...]

PR #296, opened 2016-03-21
Task 'ask' convenience functions must now be given $io

greg-1-anderson:
Pursuant to #295, Tasks that include a convenience function that calls 'ask' must now be provided an IO object. $this will usually suffice.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -87,7 +87,7 @@ class RoboFile extends \Robo\Tasks
 
         $this->taskGitHubRelease(\Robo\Runner::VERSION)
             ->uri('Codegyre/Robo')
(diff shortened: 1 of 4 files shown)


DavertMik:
No, passing `$this` is not a clean idea. If we are working on making Robo to follow SOLID principles (and you already did a great job with it), we should not allow this constructions. 

The simplest solution to me is just remove all `ask*` methods from core. Hopefully there are not many of them, and 1.0 is a good time to break things

greg-1-anderson:
Okay, I will work on a replacement PR that just removes ask* from the tasks, and moves the ask up to the RoboFile in places where these methods were used in core.

PR #297, opened 2016-03-21
Avoid use of DynamicParams trait in Robo core.

greg-1-anderson:
The existence of a magic `__call()` function in a class makes it harder for code analyzers, as they cannot tell if a call to a missing function is or is not handled by the magic function.

DynamicParams makes it a bit faster to write new tasks, and should be maintained for the convenience of external task writers. It's only a little extra code to add to Robo core, though, and having explicit functions will be better overall.

See #295.

### DIFF
--- a/src/Task/Archive/Extract.php
+++ b/src/Task/Archive/Extract.php
@@ -32,7 +32,6 @@ use Robo\Task\BaseTask;
  */
 class Extract extends BaseTask
 {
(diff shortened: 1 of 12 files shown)


DavertMik:
Yes-yes, I admit, I was too lazy to write those by hand, so I created this trait. If we can avoid using it (and you did the job for me) so that looks really good. At least If I was a static analyzer )

PR #298, opened 2016-03-21
Move 'task' shortcut to a new TaskAccessor trait.

greg-1-anderson:
Add the 'task' convenience method to BaseTask (by way of `use TaskAccessor`), so that Tasks as well as commands (RoboFiles) will be able to use the `$this->task('Foo', $a, $b);` shortcut.

Makes the code easier to read, too.

### DIFF
--- a/src/Tasklib.php
+++ b/src/Tasklib.php
@@ -3,6 +3,8 @@ namespace Robo;
 
 trait Tasklib
 {
+    use TaskAccessor;
+
     // standard tasks
     use Task\Base\loadTasks;
     use Task\Development\loadTasks;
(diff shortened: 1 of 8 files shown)


src/Tasklib.php
```diff
@@ -3,6 +3,8 @@
 
 trait Tasklib
```
DavertMik: what does TaskLib do, is that still needed?

DavertMik:
Yep, that's good move.

src/Tasklib.php
greg-1-anderson: From chat:

Tasklib is for classes (outside Robo) that want to be able to use $this->taskFoo(…) without extending Tasks.

PR #300, opened 2016-03-22
Decouple progress bar from tasks that use them

greg-1-anderson:
Tasks can now `implement ProgressIndicatorAwareInterface ` and `use ProgressIndicatorAwareTrait` to have access to a progress bar without having a direct reference to IO or an output object.

The task must still call `startProgressIndicator()` & c. to make the progress indicator go, of course. If the DI container does not have a progress bar, though, then these calls will be no-ops, so programs that do not provide progress displays may still use tasks that utilize them.

The TaskIO trait is aware of the progress interface, and will automatically hide and then re-show the progress bar if a task uses one of the `printTaskInfo` methods to produce output while the progress bar is running.

Note that ProgressIndicatorAwareTrait has a `use Timer` and manages the timer automatically.  Nearly automatically, anyway; tasks must still insert `['time' => $this->getExecutionTime()]` into their Result manually.

Note that I added a `parascript.php` script in the test data, and call it from RoboFile::tryPara().  There is a --printed option, so you can see parallel execution with or without the command output of each subprocess.

There are a couple of minor hacks in ProgressIndicatorAwareTrait to work around behaviors of the upstream. We could also fix these by subclassing the ProgressBar, but I think it would be sufficient to leave it like this for now, & eventually fix in the upstream.

### DIFF
--- a/src/Common/TaskIO.php
+++ b/src/Common/TaskIO.php
@@ -6,6 +6,8 @@ use Robo\TaskInfo;
 use Consolidation\Log\ConsoleLogLevel;
 use Psr\Log\LoggerAwareTrait;
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use Robo\Contract\ProgressIndicatorAwareInterface;
 
 /**
  * Task input/output methods.  TaskIO is 'used' in BaseTask, so any
@@ -51,7 +53,7 @@ trait TaskIO
         // The 'note' style is used for both 'notice' and 'info' log levels;
         // However, 'notice' is printed at VERBOSITY_NORMAL, whereas 'info'
         // is only printed at VERBOSITY_VERBOSE.
-        $this->logger()->notice($text, $this->getTaskContext($context));
+        $this->printTaskOutput(LogLevel::NOTICE, $text, $this->getTaskContext($context));
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Common/TaskIO.php
```diff
@@ -99,7 +101,18 @@ protected function printTaskError($text, $context = null)
...
+
+    private function printTaskOutput($level, $text, $context)
+    {
+        if ($this instanceof ProgressIndicatorAwareInterface) {
+            $this->hideProgressIndicator();
+        }
+        $this->logger()->log($level, $text, $this->getTaskContext($context));
+        if ($this instanceof ProgressIndicatorAwareInterface) {
+            $this->showProgressIndicator();
+        }
```
DavertMik: this thing was confusing to me and probably it is in general. I'd replace it with something like this:

```php
$inProgress = false;
if ($this instanceof ProgressIndicatorAwareInterface) {
  $inProgress = $this->inProgress();
}

if ($inProgress) {    
 $this->hideProgressIndicator();
}

$this->logger()->log($level, $text, $this->getTaskContext($context));

if ($inProgress) {
 $this->showProgressIndicator();
}
```

DavertMik:
Looks good to me overall. My proposed change is about to make code a bit more readable and easy to follow.

PR #301, opened 2016-03-23
Factor out consolidation/annotation-command

greg-1-anderson:
See external repository: https://github.com/consolidation-org/annotation-command


greg-1-anderson:
Scrutinizer [makes a good point about our handling of the simulated option](https://scrutinizer-ci.com/g/Codegyre/Robo/inspections/7acc9eef-f3ce-4c2e-a1e0-af5511b81293/issues/files/src/GlobalOptionsEventListener.php). This issue existed prior to this PR. I think we're one step closer to a good solution here, but I'd like to make it better still in the future.

composer.json
```diff
@@ -18,7 +18,7 @@
         "php": ">=5.5.0",
         "league/container": "~2.2",
         "consolidation/log": "~1",
-        "phpdocumentor/reflection-docblock": "~2",
+        "consolidation/annotation-command": "dev-master",
```
DavertMik: just not dev-master please. Have it released at least 0.1.0 and point to it

greg-1-anderson: OK, done.

PR #305, opened 2016-03-31
Add Write::appendIfMatches() and appendUnlessMatches() for taskWriteToFile.

greg-1-anderson:
Useful for setting up configuration files where the desired configuration may or may not already exist in the file.

### DIFF
--- a/src/Task/File/Write.php
+++ b/src/Task/File/Write.php
@@ -144,16 +144,28 @@ class Write extends BaseTask
     }
 
     /**
-     * Append the provided text to the end of the buffer if the
-     * provided regex pattern does not match any of the text already
-     * in the buffer.
+     * Append the provided text to the end of the buffer if the provided
+     * regex pattern matches any text already in the buffer.
      *
      * @param string $pattern
      * @param string
      */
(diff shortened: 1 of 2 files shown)


src/Task/File/Write.php
```diff
@@ -143,6 +143,20 @@ public function regexReplace($pattern, $replacement)
...
 
+    /**
+     * Append the provided text to the end of the buffer if the
+     * provided regex pattern does not match any of the text already
+     * in the buffer.
+     *
+     * @param string $pattern
+     * @param string
+     */
+    public function textIfMatch($pattern, $text, $shouldMatch = true)
```
DavertMik: hello again, its bad naming!
if you need a 3 lines of text to describe the method behavior it's something wrong here. The name of a method should be self-describing here, so my suggestions to have 2 methods:

* appendIfMatches($pattern, $text)
* appendUnlessMatches($pattern, $text)

..smth like that

greg-1-anderson: That looks like an improvement; I'll fix it up.

DavertMik:
Thanks!

PR #306, opened 2016-03-31
Improved taskGitStack, using cpliakas/git-wrapper.

greg-1-anderson:
Put this together some time ago, but didn't write the tests until now.

Provides a more complete git API. See https://github.com/cpliakas/git-wrapper.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "henrikbjorn/lurker": "~1.0",
         "natxet/CssMin": "~3.0",
         "pear/archive_tar": "~1.0",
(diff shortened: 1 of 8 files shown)


DavertMik:
Right now I'm for having this named as `GitPlus` task and moved to separate package. 
Current Git implementation is simple and it works without external deps. If someone need smth more powerful, they can install GitPlus, which we will mention in docs for GitStack

greg-1-anderson:
Sounds reasonable.

alexdesignworks:
@DavertMik could you please provide a link to GitPlus package. Thank you

PR #307, opened 2016-03-31
Introduce TaskInterface and Callable variants for Collection methods

greg-1-anderson:
And add appropriate type hints to all of them.

Originally, the Collection API was designed to allow clients to just call `add()` on whatever sort of object they had available. However, it is a common pattern to have things such as a `setCode()` method (e.g. Symfony's `Command::setCode()`); I followed this pattern to create `addCode()` et. al., and added typehints to all of these functions.

I don't think there are any (or many) users of Collections right now; if there are any clients that do need to adapt, the typehints will make it easy to figure out what changes are needed.

I think it's important to do this before our stable release; should have been typehinted in the first place.

### DIFF
--- a/src/Collection/CallableTask.php
+++ b/src/Collection/CallableTask.php
@@ -16,7 +16,7 @@ class CallableTask implements TaskInterface
 {
     private $fn;
 
-    public function __construct($fn)
+    public function __construct(callable $fn)
     {
         $this->fn = $fn;
     }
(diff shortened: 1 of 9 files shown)


DavertMik:
I don't like the `afterCode`, to me `after` was pretty good. I'd rather not make users choose between after and afterCode, learn them and investigate differences. I'd rather have one method which acceptns either a task or callable. Anyway 2 methods here would make a confusion.

P.S. Also `afterCode` mey be interepreted as a function which accesses error code of task.

greg-1-anderson:
Sure, no problem. I loosened up the type hinting for 'before' and 'after', but left the rest intact.

PR #308, opened 2016-03-31
Set exit code when commands complete.

greg-1-anderson:
It is important to exit with an appropriate exit code, so that `$?` is correctly set in Bash, etc.


DavertMik:
Thanks

PR #309, opened 2016-04-01
Refactor ResultData base class out of Result class

greg-1-anderson:
Drush has a function drush_user_abort() that is used to signal that a user cancelled an operation. This function works by setting a global; it would be more convenient to instead keep track of this status in the Result object.

User aborts act like errors, in that they cause the current operation to stop, and set a non-zero exit code when they are done. (There is no standard exit code for user abort, but Drush arbitrarily picked 75.) The main difference with a user abort is that any final error message is not printed.


greg-1-anderson:
This PR reveals a weakness in the design of the Result object. We want commands to be able to return Result objects as their function results, just as tasks; however, currently a Result object requires a task, or it cannot be constructed. The printResult function displays which task failed.

In the example usage this PR adds to the RoboFile, I use an empty Collection as a placeholder task, but this is not right.  Thoughts on what to do here?  We could maybe split off a base class of Result for use by commands, that does not take a Task. This would maintain the existing Result contract, but allow taskless results to be returned from commands. Maybe call the base class ResultCommand, and `class Result extend ResultCommand`?

greg-1-anderson:
I decided to go with my suggestion above, and factored out ResultData from Result. This has the added bonus that we now have a kind of result that does not print anything when instantiated, which might be handy in some places.

greg-1-anderson:
In this PR, Result::$message and Result::$data['output'] serve about the same purpose. The only reason I separated them was that $message historically only contained strings, whereas the structured output of a command can be other types, e.g. an array of table rows and columns. However, there are already places in the code where I put arbitrary, potentially non-string data into $message; see [CallableTask](https://github.com/Codegyre/Robo/blob/user-cancel/src/Collection/CallableTask.php#L38) for example. Also, with formatters, non-string data can be formatted and rendered, so it becomes more reasonable to put non-string data into $message.

In any event, the handling of $message and $data['output'] should be made consistent. Are you okay with unifying these, and putting non-string data into $message?

RoboFile.php
```diff
@@ -529,6 +567,23 @@ public function generateTask($className, $wrapperClassName = "")
         $template = file_get_contents(__DIR__ . "/GeneratedWrapper.tmpl");
         $template = str_replace(array_keys($replacements), array_values($replacements), $template);
 
+        // Returning a string will cause Robo to print it and then
+        // exit with a "no error" (status code 0) result.
         return $template;
     }
 }
+
+class ExceptionTask extends \Robo\Task\BaseTask
```
greg-1-anderson: RoboFile is just an example file, so perhaps it is okay to put this example task here?  I don't really want to put it anywhere in 'Tasks'.  The danger of doing this, I suppose, is that maybe some folks might start implementing their own custom tasks right in their RoboFiles. This will give them deprecation errors when their task attempts to print, of course. Not sure what to do here; we could just delete this example, but I think it's kind of nice to have.

greg-1-anderson:
I also unified the way that exception / exit code handling is done, and fixed up the usage of ResultData vs Result in Collection.

DavertMik:
Checked the code. Looks good. No objections
+1 for using constants, fromException and other poweful things

greg-1-anderson:
Thanks!

PR #311, opened 2016-04-01
Add a Collection::progressMessage() method

greg-1-anderson:
Allows Collection clients to insert log messages into the task stream.

Note that this PR branches from #307, since it is based on some API changes made there.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -111,6 +111,7 @@ class RoboFile extends \Robo\Tasks
     public function docs()
     {
         $collection = $this->collection();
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
The two questions here are:

1) Should this method be named `progressMessage()`, or something else?
2) Should the log message read (as this PR currently has it):
- `[Progress]` My message

If we remove the additions to `$context`, then it reads:

- `[notice]` My message

greg-1-anderson:
Merged.

PR #312, opened 2016-04-03
Do not suggest wildcard in composer

garak:
You should not suggest using `*` wildcard when requiring a package. This is a bad practice and prevents from breaking BC in major versions, making semver useless

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,7 @@ Now you can use it just like `robo`.
 
 ### Composer
 
-* Add `"codegyre/robo": "*"` to `composer.json`.
-* Run `composer install`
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Note that Robo the current version of Robo is 0.x, which, per SemVer, makes no promise about backwards compatibility.  (0.x versions can break backwards compatibility every release, regardless of which components of the version string are updated.)

Making the suggested documentation change is a good thing to do in preparation for the upcoming 1.0 stable release, though.

greg-1-anderson:
Merged; thanks.

PR #314, opened 2016-04-04
Fix the visibility of the Result::$stopOnFail property

coudenysj:
The "protected" visibility causes src/Tasks.php#L16 to throw exceptions (PHP Fatal error: Uncaught Error: Cannot access protected property Robo\Result::$stopOnFail in vendor/codegyre/robo/src/Tasks.php:16)


greg-1-anderson:
Did not see this PR -- already fixed this directly based on your comment on the commit that broke this.

greg-1-anderson:
https://github.com/Codegyre/Robo/commit/a10dc1103757fb94f1acef27dc16b5b5bc9e34dc

PR #316, opened 2016-04-06
Process collections/tasks returned from commands

greg-1-anderson:
If a Robo command returns a Collection or TaskInterface, then use a process hook to run it automatically. If a Robo Result contains an 'output' element in its data, then print the result (after applying formatters, if applicable).

This will conflict with #309, but I'll merge them when it's time.


DavertMik:
Looks good to me

PR #323, opened 2016-04-25
Add coveralls.

greg-1-anderson:


greg-1-anderson:
So there are a few problems here:

1. Only a few classes are selected. Perhaps there is something wrong with the path selector (`- src/*`). Different pattern?  Different path? ('- ../src/*`?)
2. `XML report generated in build/logs/clover.xml` emitted twice, once for CLI tests, and once for Unit tests.  Maybe the results here are merged, but if they are overwritten, that would be a problem for coverage calculations.
3. The coverage report cannot be pushed to Coveralls. Error message: `coverage_clover XML file is not readable`.  Perhaps the path should be `build/logs/clover.xml`?

RoboFile.php
```diff
@@ -10,10 +10,16 @@ class RoboFile extends \Robo\Tasks
...
      */
-    public function test($args = "")
+    public function test($args = "", $options = ['coverage' => false])
     {
-        return $this->taskCodecept()
+        $taskCodecept = $this->taskCodecept()
             ->args($args);
+
+        if ($options['coverage']) {
+            $taskCodecept->coverageXml('build/logs/clover.xml');
```
DavertMik: does the `build/logs` dir actually exist?

greg-1-anderson: phpunit created it for me, but I could try adding build/logs/.gitkeep if codeception would be happier with it there. If codeception just calls through to phpunit, this probably isn't necessary.

greg-1-anderson: Just added one; it's building now. We'll see if it does anything different.

codeception.yml
```diff
@@ -15,3 +15,7 @@ modules:
             user: ''
             password: ''
             dump: tests/_data/dump.sql
+coverage:
+    whitelist:
```
DavertMik: replace with:

```
coverage:
    enabled: true
    include:
      - src/*
```

RoboFile.php
DavertMik: by default codeception builds clover reports to `tests/_output`

greg-1-anderson: No change; I think it's probably some issue about relative paths. Not sure what the cwd is. Don't feel like setting up xdebug right now; could do that later to find out.

PR #327, opened 2016-05-04
Add possibilty to exclude files in _copyDir

yvesh:
This PR adds an optional third parameter to copyDir, allowing to pass a list of files / folders to exclude. 

$this->_copyDir('src', 'target', ['exclude', 'folders']);


greg-1-anderson:
Looks useful, but needs tests and docs.

PR #331, opened 2016-05-06
Fixed FileSystem shortcuts referring to the wrong container alias.

spideyfusion:
When trying to use the `FileSystem` shortcuts you get the following error:

```
Alias (taskFileSystemStack) is not being managed by the container
```

This happens because the `FilesystemStack` class is named with a lowercase **s**.


greg-1-anderson:
Thanks for this; I think that the "FileSystem" vs. "Filesystem" discrepancy in the source code is a common source of errors, so I proposed #343 to make this consistent.

greg-1-anderson:
Fixed with #343. Thanks.

PR #332, opened 2016-05-06
Adds initial new changelog task

stevewest:
Potential new change log task.

Currently this is missing docs, tests and breaks backwards compatibility, I am opening this pull request to gain feedback from any who wish to give it.

Current TODO:
 - [ ] Restore some kind of BC
 - [ ] Update docs with extra features
 - [ ] Add tests

Thoughts/ideas about what else the task can do are welcome, anything to make this more useful.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
         "symfony/finder": "~2.5|~3.0",
         "symfony/console": "~2.5|~3.0",
         "symfony/process": "~2.5|~3.0",
-        "symfony/filesystem": "~2.5|~3.0"
+        "symfony/filesystem": "~2.5|~3.0",
+        "stevewest/changelog": "^1.2"
     },
     "require-dev": {
         "patchwork/jsqueeze": "~1.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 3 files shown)


greg-1-anderson:
stevewest/changelog looks awesome, but it would be a little odd to have that stevewest/changelog require robo, and robo require stevewest/changelog.

@DavertMik should weigh in on whether or not he wants to pursue integrating this into Robo core vs. leaving it as an external Robo-based application.

stevewest:
Thanks for the feedback, I'd be happy to close this PR and turn this into an external package `robo-changelog` or somesuch. It's also not an issue to remove the robo dependency in changelog, it's just a habit to have all the dev tools used specified.

greg-1-anderson:
At the moment, we don't have guidance on what the best practices are for writing an external Robo-based tool.  I think that the way you did it is good, because it is easier for the end user, especially if they are not already familiar with Robo.

My suggestion vis-a-vis bringing changelog into Robo core was that, if that was done, then probably all of the source should be copied here.  That's just my initial impression, though; should wait for @DavertMik to weigh in before doing that work.

DavertMik:
I think this is good as standalone package. Changelog task was initially a really simple thing, I'm not sure it solve all tasks which stevewest/changelog does. 

Probably we should have a pages of 3rd party plugins and promote `stevewest/changelog` as much powerful alternative.

DavertMik:
P.S. Actually we have standard for naming all 3rd-party packages starting with `robo-` prefix. This way they can be easily found on packagist.

stevewest:
Thanks for the feedback, I'll look at getting this set up as an extra package then.

PR #334, opened 2016-05-08
Improved pack to zip logic

Rarst:
- don't silently skip dot files
- replace Windows slashes for cross-platform friendly result


greg-1-anderson:
Thanks.

PR #339, opened 2016-05-11
colored output of composer

ekho:


greg-1-anderson:
This looks good, but there is the question of how we should handle boolean options.  The APIGen tasks use [asTextBool](https://github.com/Codegyre/Robo/blob/master/src/Task/ApiGen/ApiGen.php#L75); perhaps this could be moved to BaseTask and used consistently?

cc: @DavertMik

DavertMik:
> This looks good, but there is the question of how we should handle boolean options. The APIGen tasks use asTextBool; perhaps this could be moved to BaseTask and used consistently?

No, I like this PR the way it is and not to be mixed with ApiGen.

Thanks!

@greg-1-anderson can we make coveralls not to fail builds?

greg-1-anderson:
I configured Coveralls so that the total coverage must be > 20% (currently 43%) to pass, and a decrease in coverage won't cause a failure unless coverage drops at least 40%. That should effectively stop coveralls from marking tests as failures if coverage goes down -0.05%.

PR #342, opened 2016-05-20
Fixed missing definition of run-method in the docs

timtegeler:
I think there was a missing definition of the run-method in the docs.


greg-1-anderson:
Thanks.

PR #344, opened 2016-05-21
Allow environment variable injection in stacked commands.

dustinleblanc:
Sometimes users may want to pass environment variables to a command stack the same way they can pass environment variables to individual command execution.

### DIFF
--- a/docs/tasks/Docker.md
+++ b/docs/tasks/Docker.md
@@ -29,6 +29,7 @@ Class Build
 * `addAsRollback($collection)` 
 * `addAsCompletion($collection)` 
 * `addToCollectionAndIgnoreErrors($collection, $taskName = null)` 
+* `env(array $env)`  Sets the environment variables for the command
 * `dir($dir)`  changes working directory of command
 * `printed($arg)`  Should command output be printed
 * `arg($arg)`  Pass argument to executable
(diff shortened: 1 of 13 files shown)


DavertMik:
Could you provide example for usage?
I'm not sure where to place this example, but its definitely needed

dustinleblanc:
Documentation added. Note: the docs seem outdated, I patch added only env stuff but probably time to do an overall bump...also apparently coveralls now tanks...but should we add some test coverage of this? I confess I'm not entirely sure how to unit test `env()`

docs/tasks/Docker.md
```diff
@@ -224,7 +229,7 @@ $this->taskDockerRun('wordpress')
 * `interactive()` 
 * `exec($run)` 
 * `volume($from, $to = null)` 
-* `env($variable, $value = null)` 
+* `injectEnv($variable, $value = null)`
```
greg-1-anderson: Maybe the `env` method should take $variable, $value parameters as the existing taskDockerRun does, rather than using array syntax.

src/Task/Docker/Run.php
```diff
@@ -101,7 +101,7 @@ public function volume($from, $to = null)
         return $this;
     }
 
-    public function env($variable, $value = null)
+    public function injectEnv($variable, $value = null)
     {
         $env = $value ? "$variable=$value" : $variable;
         return $this->option("-e", $env);
```
greg-1-anderson: Docker run could just build its -e options from the $env array in BaseTask.

greg-1-anderson:
Rebase with master to fix test failures.  Additional test coverage would be helpful.

greg-1-anderson:
Re-open if you want to return to this issue.

PR #350, opened 2016-05-29
Add task definition for "composer validate"

Sweetchuck:
Add a new task to wrap the "composer validate" command


DavertMik:
Thanks. Looks good to me. @greg-1-anderson what you think?

greg-1-anderson:
LGTM too -- thanks for including tests!

PR #353, opened 2016-06-02
Fix some documentation problem around \Robo\Config::getContainer()

Sweetchuck:
Fix some documentation problem around \Robo\Config::getContainer()


greg-1-anderson:
Yep, that was out of date -- thanks!

Issue #354, opened 2016-06-05
Configurable yell() and say() prefix 


Sweetchuck:
The prefix of the `\Robo\Common\IO::say()` and `::yell()` is hard-coded to `$char = strncasecmp(PHP_OS, 'WIN', 3) == 0 ? '>' : '➜';` and two spaces.
This means that if somebody wants to write Behat test for an application which uses Robo in the background then impossible to write step definition for all platform.
For example

```cucumber
Then I should see the following text in the output:
"""
➜  Line 1
➜  Line 2
"""
```

It would be awesome if the prefix were configurable. (or same on all platform)

greg-1-anderson:
Submitted a proposed solution in #355. With this PR, you could call `Config::setYellPrefix()` in your test initialization to set this to a constant value.

DavertMik:
`Config::setYellPrefix()` 
I think we will end up with filling Config class with lots of static methods for basic config changes. I think we can use `--no-ansi` option from Symfony console which disables non-ansi output. In our case it can also replace UTF chars to basic ANSI output.

greg-1-anderson:
That's a good idea.

PR #355, opened 2016-06-06
Define Config::setYellPrefix()

greg-1-anderson:
This allows tests to set this to a constant value, if needed.

Addresses #354.


greg-1-anderson:
We'll do #354 another way.

PR #361, opened 2016-06-12
Escape git commit message argument

tauno:
Commit messages are likely to contain special characters and it's reasonable for them to contain newlines. Both of these can cause trouble with the current hardcoded argument handling.


DavertMik:
Thanks!

PR #373, opened 2016-07-02
Introduce a task builder to Robo

greg-1-anderson:
The other day I was looking at UnitJS, which has a very interesting way of [chaining together test configuration methods](http://unitjs.com/guide/quickstart.html#example-proposal-of-structured-unit-tests-suite).  The basic idea is that a `test` object has chained methods such as `given()`, `when()` and `then()`, and each of these in turn also has chained methods for configuring its parameters.  How does that work?  The `test` object remembers the last `given()`, `when()` etc. that it created, and proxies calls to its configuration methods as necessary.  This can be done because Javascript is a dynamic language, and is very flexible about this sort of thing.

PHP also has enough dynamic language features to support this same pattern.  Here is an example of a section of the `phar:publish` command converted to use a builder:
```
        $this->builder()
            ->taskGitStack()
                ->checkout('gh-pages')
            ->taskFilesystemStack()
                ->remove('robo.phar')
                ->rename('robo-release.phar', 'robo.phar')
            ->taskGitStack()
                ->add('robo.phar')
                ->commit('robo.phar published')
                ->push('origin', 'gh-pages')
                ->checkout('master')
                ->run();
```
These tasks are automatically added to a collection and executed together when `run` is called. This enhancement represents a significant improvement over the existing collections API, I think.


greg-1-anderson:
`$this->builder()->taskFoo()` is flagged as an error by Scrutinizer, because all of the loadTask methods are protected in Robo.  These are protected because Robo makes all of the public methods of the RoboFile commands.  The `builder()` implementation allows these to be called via the `__call()` method, which can access these methods despite their `protected` status. I think this is fine, because these methods are only `protected` to prevent them from becoming commands, and are otherwise intended to be used as `public` methods.

My general feeling is that the improvement here is worth this small oddity for Robo 1.x.  Perhaps in Robo 2.x, after some period of investigation time, to see how we like the `builder()` model, we could consider moving all of the `taskFoo` loadTask() implementations into `builder()` as public methods, and remove them from the RoboFile.  This would then require the use of `builder()->` in RoboFile scripts for all tasks, but would no longer require indirect access to the protected methods.

greg-1-anderson:
The general problem with the current implementation of the builder is that it becomes hard to add new tasks. I think this could be fixed by:
 a) Make all task accessors (taskFoo) public
 b) Do not create commands for public methods that start with `task`  unless they have a `@command task:foo` annotation
  c) Give the builder a reference to the RoboFile, and use that to create tasks that are not found on the builder.

We had previously discussed and rejected the idea of not creating commands for methods that start with `task`, but perhaps doing this would be okay if it was possible to select these again via an annotation.  (Consolidation/annotated-command already allows the command name to be renamed via `@command`.)  We could also provide ide function completion for built-in tasks off of the builder, but I'm not sure that the ide would be able to figure out what the custom tasks were available to the builder in any scenario. It would be possible to start doing development with `$this->taskMyCustomTask()` using ide function completion for custom tasks, and then edit it to `$this->builder()->taskMyCustomTask()` to get the better collection management that the builder provides.

I think that the builder is a big improvement for collections, and justifies these changes.

greg-1-anderson:
The latest commit preserves the builder usage described in the initial post, but also gives access to any custom tasks added in a specific RoboFile.

src/TaskBuilder.php
```diff
@@ -0,0 +1,178 @@
...
+use Robo\Contract\TaskInterface;
+use Robo\Common\IO;
+use League\Container\ContainerAwareInterface;
+use League\Container\ContainerAwareTrait;
+use Robo\Contract\WrappedTaskInterface;
+
+class TaskBuilder implements ContainerAwareInterface, TaskInterface
+{
+    use ContainerAwareTrait;
+    use LoadAllTasks;
```
DavertMik: We can load all tasks as public methods:

```php
use LoadAllTasks { taskExec as public taskExec; 
    taskCopyDir as public taskCopyDir 
    /// etc...
}
```

DavertMik:
> We had previously discussed and rejected the idea of not creating commands for methods that start with task, but perhaps doing this would be okay if it was possible to select these again via an annotation. (Consolidation/annotated-command already allows the command name to be renamed via @command.) We could also provide ide function completion for built-in tasks off of the builder, but I'm not sure that the ide would be able to figure out what the custom tasks were available to the builder in any scenario. It would be possible to start doing development with $this->taskMyCustomTask() using ide function completion for custom tasks, and then edit it to $this->builder()->taskMyCustomTask() to get the better collection management that the builder provides.

I think this would be good for v2. RIght now using `__call` for builder seems ok to me

greg-1-anderson:
I think I can make `__call` work with protected task methods. I'll give it a try.

greg-1-anderson:
Latest commit restores tasks to protected status, and uses `__call` with TaskBuilder::build() to access tasks from the builder object.  Works great.

PR #377, opened 2016-07-09
Add a collection builder to Robo. 

greg-1-anderson:
Allows tasks to be chained together during initialization; all chained tasks are added to a container in the builder.

Note that the task builder is introduced in #373; this PR is branched off of that PR, so the diffs here only show the addition of the progress bar to the Collection class, and taskForEach.


greg-1-anderson:
This PR is now in pretty good shape. The new taskForEach is pretty easy to use, and progress bars are automatically added to any Collection that runs for more than two seconds (see `./robo try:progress`).

Still to do:

- [ ] Documentation updates.
- [ ] Some Scrutinizer-recommended fixes
- [ ] Decide if `Collection::add` should be renamed to `addTask`.  This method was called `addTask` in TaskBuilder to avoid name conflicts. This would break b/c, but few are probably using this method.
- [ ] Decide if `Collectable` should be removed; TaskBuiler is a much better way than this to construct collections of tasks.  This did not exist in Robo 0.6.x, so it would not be inconceivable to just remove this; however, if anyone is using collections, it is likely they are using these APIs. The alternative would be to keep these APIs in 1.x, and remove them in the 2.x branch once that is created.
- [ ] Decide if Timer and ProgressIndicatorAwareTrait should be simplified. These probably should not have been traits from the get-go; traits should be used when the relationship with the containing class is `is-a`; in the case of timer and progress indicators, the relationship is `has-a`, so a class is better. The changes in this PR necessitated making classes for these, so `TimeKeeper` and `ProgressIndicator` classes were introduced.  If `Timer` was renamed, then `TimeKeeper` could be renamed to `Timer`; this would make more sense, but since `Timer` has been in use since before Robo 0.6.0, that change should probably wait until the 2.x branch.

RoboFile.php
```diff
@@ -216,23 +217,17 @@ public function publish()
...
-        $this->taskFilesystemStack()
-            ->remove('docs/changelog.md')
-            ->addAsCompletion($collection);
-        $this->taskExec('mkdocs gh-deploy')
-            ->addToCollection($collection);
-        return $collection;
+        ->completion($this->taskFilesystemStack()->remove('docs/changelog.md'))
+        ->taskExec('mkdocs gh-deploy');
+
+        return $builder;
```
DavertMik: let's not return a builder object...

RoboFile.php
```diff
@@ -485,43 +481,89 @@ public function tryDeprecated()
...
+                $this->taskDeleteDir('g')
+            )
+            ->taskFilesystemStack()
+                ->mkdir('g/h')
+                ->touch('g/h/h.txt')
+            ->taskFilesystemStack()
+                ->mkdir('g/h/i/c')
+                ->touch('g/h/i/i.txt');
+
+        return $this->builder()
```
DavertMik: same here. Return only result of execution and not the commands

RoboFile.php
```diff
@@ -531,6 +573,46 @@ public function tryTmpDir()
...
+            ->taskFilesystemStack()
+                ->mkdir($workdir)
+            ->taskCleanDir($workdir)
+            ->taskForEach($processList)
+                ->withBuilder(
+                    function ($builder, $key, $value) use ($workdir) {
+                        return $builder
+                            ->taskFilesystemStack()
+                                ->mkdir("$workdir/$value");
```
DavertMik: oh, it looks just like JavaScript! 
I think we should publish Robo on NPM :)

docs/collections.md
```diff
@@ -1,20 +1,24 @@
-# Collections
+# Collections and Builders
 
-Robo provides task collections as a means of making error detection and recovery easier. When Robo tasks are added to a collection, their execution is deferred until the `$collection->run()` method is called.  When using collections, a Robo script will go through three phases:
+Robo provides task collections as a means of making error detection and recovery easier. When Robo tasks are added to a collection, their execution is deferred until the `$collection->run()` method is called.  
 
-1. Determine which tasks will need to be run, and create a collection.
+The easiest way to manage a collection of tasks is to use a task builder. Task builders allow tasks to be created via chained methods.  All of the tasks created by the same builder are added to a collection; when the `run()` method is called, all of the tasks in the collection run. If one of the tasks fail, then the operation will be aborted; rollback tasks may also be defined to restore the system to its original condition.
```
DavertMik: before this paragraph pleas provide an example of creating and using basic collection (with code)

docs/collections.md
```diff
@@ -24,28 +28,21 @@ class RoboFile extends \Robo\Tasks
...
-        $this->taskFilesystemStack()
-            ->remove('docs/changelog.md')
-            ->addAsCompletion($collection);
-        $this->taskExec('mkdocs gh-deploy')
-            ->addToCollection($collection);
-        $collection->run();
+        ->completion($this->taskFilesystemStack()->remove('docs/changelog.md'))
+        ->taskExec('mkdocs gh-deploy');
+
+        return $builder;
```
DavertMik: don't return builder... Please )

RoboFile.php
greg-1-anderson: Yeah, the strong inversion of control here is very reminiscent of node.js.  Overall, I prefer an Iterator object over a function that takes a callback object; however, in this instance, the fact that we are managing progress and failures in the collection necessitates this design.  This is in fact also similar to the design decisions that drove node.js, although the reasons are different.

PR #379, opened 2016-07-16
Add hg support

royallthefourth:
Added tests, tasks, and documentation for Mercurial

### DIFF
--- a/docs/tasks/Vcs.md
+++ b/docs/tasks/Vcs.md
@@ -89,3 +89,49 @@ $this->taskSvnStack('username', 'password')
 * `dir($dir)`  changes working directory of command
 * `printed($arg)`  Should command output be printed
 
+## HgStack
+
+
+Runs Mercurial commands in stack. You can use `stopOnFail()` to point that stack should be terminated on first fail.
+
+``` php
+<?php
+$this->taskHgStack()
+ ->stopOnFail()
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
This looks pretty cool. With this PR, we now have three VCS mechanisms that operate via a very similar API.  Perhaps we should add an appropriate VCS interface for each of these tasks to implement, to formalize what the least-common-denominator behavior is between the mechanisms.  This would allow folks to write VCS-agnostic operations, which might be useful for some.

royallthefourth:
I like that idea! After this gets merged, I'd be happy to write such an interface and adapt these VCS classes to work with it.

Also, how can I contribute to getting 1.0 out the door? Is it just the globbing feature and documentation remaining or does the milestone need to be updated?

greg-1-anderson:
+1 for merging this PR as-is and adding an interface as a follow-on task.

I don't think that the globbing feature should be a blocker for 1.0; we will need some documentation updates, but not everything mentioned in that issue necessarily needs to be done prior to the stable release. There are some potential b/c breaking changes in #377 that we may or may not choose to make.  Exact requirements for 1.0 are up to @DavertMik; hopefully we will get there soon.

DavertMik:
Great! Thank you very much!

PR #384, opened 2016-07-21
Fixed robo.phar to not contain dev dependencies.

fago:
I noticed that robo.phar is really big (12MB!!). Researching this it quickly turned out that it includes the dev dependencies. Problem is that the previous code had a bug, which lets it pick up currently install dev-dependencies when packing the phar. The PR fixes this, what results in a way smaller phar: ~1.5MB.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -238,10 +238,12 @@ class RoboFile extends \Robo\Tasks
     {
         $collection = $this->collection();
 
+        // Make sure to remove dev files before finding the files to pack into
+        // the phar.
         $this->taskComposerInstall()
             ->noDev()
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
LGTM.

Note that this PR removes the second run of composer install, so debug components will be removed from the current working project. As a follow-on task, it would probably be a good idea to make the pack phar command build from a fresh copy every time, perhaps after a successful build on CI.

DavertMik:
Yes, I think it should be merged. However, it is a good idea to reinstall all dev packages after phar package is built.

DavertMik:
Thanks

PR #388, opened 2016-07-23
GitHub stack

greg-1-anderson:
Work in progress -- porting my GitHub API scripts to Robo.


DavertMik:
That's really great stuff. I was so lazy to implement only release script but it would be cool to automate all github interactions with Robo. Keep the good work!

greg-1-anderson:
I'll re-implement this using https://github.com/KnpLabs/php-github-api as you suggested.

greg-1-anderson:
This isn't ready to merge, so removing "Reviewed" label.

greg-1-anderson:
Will return here later, but with a different implementation as noted above.

PR #390, opened 2016-07-26
Refactor config

### LINKED ISSUE #387: Clean up Robo globals
Folks in #383 want more control over how Robo does output. I think there are better ways to do this than add more globals.  Before I document this, though, I'd like to do some cleanup.

As a general rule, an application's data objects should not keep references to the container -- or to other objects that keep references to the container. To keep Robo clean in this respect, I would like to split out a `Robo` class from the existing `Config` class.  The static Robo class would contain the reference to the container, and all related things, like the `input()` and `output()` methods.  `Config`, then, would keep only configuration values, not service references.  This change would affect everyone who calls `Config::getContainer()` or `Config::output()`; however, this practice is deprecated anyway, so I don't feel bad about forcing folks who do this to change to `Robo::getContainer()`.  (n.b. we can name this new static class anything.)

Once we do this, then `Config` will be free of container references. We can remove the static from ContainerBuilder, and replace it with a reference to Config via ConfigAwareInterface / Trait.

After that, we could consider giving RoboFiles an init method, and hand them the container; then they would be able to replace ResultPrinter or the logger or whatever they want to do if they don't like the default printing options.

greg-1-anderson:
Fixes #387.


RoboFile.php
```diff
@@ -581,8 +581,7 @@ function ($context) {
 
     public function tryProgress()
     {
-        $delayUntilProgressStart = 2;
-        CollectionBuilder::setDefaultProgressBarAutoDisplayInterval($delayUntilProgressStart);
+        $delayUntilProgressStart = \Robo\Robo::config()->get(\Robo\Config::PROGRESSBARAUTODISPLAYINTERVAL);
```
DavertMik: `PROGRESSBARAUTODISPLAYINTERVAL` looks just like `T_PAAMAYIM_NEKUDOTAYIM`

src/Config.php
```diff
@@ -1,149 +1,44 @@
...
-     */
-    public static function getContainer()
-    {
-        if (static::$container === null) {
-            throw new \RuntimeException('container is not initialized yet. \Robo\Config::setContainer() must be called with a real container.');
-        }
-        return static::$container;
-    }
+    const SIMULATE = 'simulate';
+    const PROGRESSBARAUTODISPLAYINTERVAL = 'progress-delay';
```
DavertMik: use `_` for better naming

DavertMik:
> After that, we could consider giving RoboFiles an init method, and hand them the container; then they would be able to replace ResultPrinter or the logger or whatever they want to do if they don't like the default printing options.

Would you mind to add an example of overriding default logger?

Everything else looks good to me

greg-1-anderson:
I'll work on container manipulation + example docs on a follow-on PR; for now, I will fix the constant and merge.

Issue #383, opened 2016-07-20
How to suppress printing execution time using $this->taskExec() ?


disharmonized:
How to suppress printing execution time using `$this->taskExec()`? I have my own implemented execution time counter and i want to use it, not robo's version. Are there any opportunities to achieve that?

I am using 1.0.0beta1 robo's version.

greg-1-anderson:
Robo prints success/error results, including execution time, from the constructor of the Result object.  This gives clients no opportunity to suppress this output.

Ideally, this would be refactored in a future version of Robo. In the 1.x timeframe, the best we could do would be to provide some sort of static variable to disable the result output. I'm not sure if that would be worth the overhead.

Could you provide more details on your use case?  e.g. What would your desired output (including your custom time counter output) look like?

disharmonized:
Ok, here's my case:

```php
class RoboFile extends \Robo\Tasks
{
    public function exportFromMySql()
    {
        $taskStartTime = time();

        $sql = 'select * from bigtable;';
        $output = '~/output/out.csv';

        $this->taskExec('mysql')
            ->arg('testdb')
            ->arg('--host=localhost')
            ->arg('--port=1234')
            ->arg('--user=test')
            ->arg('--password=test')
            ->arg('--quick')            
            ->arg(' -c "' . $sql . '" > ' . $output)
            ->printed(true)
            ->run()
            ->stopOnFail();

        //it prints execution time like
        //Done for 05:10:45
        $this->say('Done for ' . Util::getFormattedExecutionTime($taskStartTime, time()));

        //and after this robo prints it's own execution time output, which is already unnecessary for me
        
    }

    public function importToMongoDb(){
        //similar task
    }
}
```

It will be very handy to have opprotunity to change behaviour of Result output, i.e disable it, format it, whatever.

disharmonized:
>  In the 1.x timeframe, the best we could do would be to provide some sort of static variable to disable the result output
I think that will be enough for me.

cyberhorse:
Static variable is fine. It should disable both the starting and ending output.

greg-1-anderson:
I refactored the Config class to make it easier to reference from other classes (Config no longer contains a reference to the Container).  TaskIO now checks Config::isSuppressed() before doing output.  Things emitted by `say` are not affected.

Some things I realized in working on that PR:

- Tasks should not care about output; they should just use TaskIO.
- Robo commands should not change the default output formatting; they should rely on the application to do that.
[...]

PR #396, opened 2016-08-01
Add a --supress-messages global option to turn off all task io messages.

greg-1-anderson:
Related to #383, this gives users a way to turn off Robo messages -- which is to say, anything produced by TaskIO.


greg-1-anderson:
Note that output produced by `say` et. al. are not affected.  Use `--quiet` to get rid of all output.

Also, this PR does not supress error and warning messages, as those are important and should probably always be shown.  This is open to discussion, though.

DavertMik:
Ok, that makes sense. But we have too much levels of hiding here :(

`--quiet`, `--supress-messages`, etc

So it could be like the opposite to verbosity `--quiet`, `--very-quiet`, `--very-very-quiet`, or `-q`, `-vq`, `-vvq`
Not the best naming and I don't have ideas for better ones. So let's stick with current behavior )

PR #397, opened 2016-08-04
Ensure that 'time' is always set in the result of any collection builder

greg-1-anderson:
Remove the 's' from the return value of Result::getExecutionTime(). Add a duration formatter.


DavertMik:
Thanks. Looks good to me, what about docs?

greg-1-anderson:
I didn't change any documented behavior here; the documentation in the `Progress` section of getting-started.md is still accurate.

LMK if there are any specific documentation gaps you'd like to see filled, though.

PR #399, opened 2016-08-05
Rename repository and packagist paths.

pingers:
This MR renames packagist paths to `consolidation/robo` and github paths to `consolidation-org/robo`, inline with the current naming of the repository.


DavertMik:
Thank you!
I will rename that on packagist soon as well

PR #400, opened 2016-08-05
Fix phar

greg-1-anderson:
The phar:build task should include all of the projects in the 'suggest' section, but should not include anything else (e.g. the projects used for running the unit tests).

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -84,7 +84,7 @@ class RoboFile extends \Robo\Tasks
             $stable = false;
             $this->say('non-stable release');
         }
-        
+
         $releaseDescription = $this->ask("Description of Release\n");
 
         $this->docs();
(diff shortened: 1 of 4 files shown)


DavertMik:
Do you revert back dev dependencies anywhere? I don't see.

greg-1-anderson:
@DavertMik: The new implementation creates a temporary copy of the Robo code via `git clone`, using the local working copy of the repository as the source of the clone. It then does the composer install on the temporary copy, leaving the local working copy unchanged. At the end, the temporary directory is automatically deleted.

This version could potentially benefit from two warnings:

- If you have uncommitted changes, it could warn that the phar you are building will not be exactly the same as `./robo`.
- If you have committed changes that have not been pushed, it could warn you that the phar you are building contains code that others do not have yet.

greg-1-anderson:
I merge/force-pushed a change to include a warning when there are uncommitted changes present during a phar:build. I didn't think the other change was that important.

greg-1-anderson:
Oh, this has 'Reviewed' label, so it's okay to merge?  Was thrown off by the question, and didn't notice the label.  Just let me know if there's anything else you want to see here.

PR #401, opened 2016-08-05
Exclude RoboFile.php from Scrutinizer.

greg-1-anderson:
Scrutinizer does not like our RoboFile. It is too long, too complicated, and coupled with too many components. However, this is expected, because this file is just a sample; it is expected to demonstrate a lot of different things.

Let's exclude it from analysis, since the results are not reflective of Robo as a whole.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,3 @@
+filter:
+    excluded_paths:
+      - "RoboFile.php"


DavertMik:
> Scrutinizer does not like our RoboFile. It is too long, too complicated

To be honest, I don't like it either. It's hard to differentiate valuable scripts and sample scripts. The worst thing I don't like is very long task descriptions that go to the next line in console.

greg-1-anderson:
We might do well to separate out some of the examples into another file, e.g. examples/RoboFile.php.  Fine-tuning the help text also sounds like a good idea.

PR #405, opened 2016-08-11
Add return type to @method tag to enable autocomplete

Korikulum:
Add return type to FilesystemStack @method tag to enable auto complete when chaining methods in IDEs, eg `$this->taskFilesystemStack()->copy($source, $destination)->run();`.


greg-1-anderson:
Thanks.

PR #407, opened 2016-08-14
Work-in-progress: use ProcessBuilder API with Exec chained initializers.

greg-1-anderson:
Not all of the code has been updated to use the changes introduced in this PR -- hence the failing tests.

This might be best saved for the Robo 2.x branch; however, I do think it will be important to escape the parameters to `arg()` and `option()` in Exec chained initializers.

### DIFF
--- a/src/Common/CommandArguments.php
+++ b/src/Common/CommandArguments.php
@@ -1,12 +1,53 @@
 <?php
 namespace Robo\Common;
 
+use Symfony\Component\Process\ProcessBuilder;
+
 /**
  * Use this to add arguments and options to the $arguments property.
  */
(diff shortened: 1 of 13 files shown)


greg-1-anderson:
Favoring `Process` and `ProcessBuilder` objects to strings has some impacts on the tests. Is there a better way to have Codeception [test to see if a method of a double is called](https://github.com/consolidation-org/Robo/pull/407/files#diff-1909cc524971473ac8cd934f1327bdccR40)?

greg-1-anderson:
I don't know that this is necessary.  Might revisit later, but closing for now.

PR #408, opened 2016-08-14
Rename 'Robo standalone scripts' to 'Robo scripts'.

greg-1-anderson:
It was a misnomer to call 'Robo scripts' "standalone", because these scripts require the Robo application in your $PATH, and this is not part of the script.

We should reserve the term "standalone" for applications derived from Robo that contain the entire Robo application, e.g. by requiring Robo in composer.json.


DavertMik:
Thanks

PR #410, opened 2016-08-17
fixed some docs e.g. the return type of the run() method

tsari:
As PHPstrom is always complaining about return types I had to fix this to get rid of the warning.
I also fixed the missing param while calling Result::error().


greg-1-anderson:
Thanks.

PR #411, opened 2016-08-17
Remove taskPackPhar in favor of using box.phar tool.

greg-1-anderson:
The existing taskPackPhar is buggy; phars built with this utility will always receive an empty string when calling \Phar::running().  This shortcoming makes it inconvenient to make standalone phars based on Robo that embed resources (e.g. their RoboFile) inside the phar.  It is possible to simply hardcode the `phar://` path, but it is preferable to have the phar behave as documented.

This PR removes taskPackPhar entirely, and replaces it with the [box.phar](https://github.com/box-project/box2) tool, which is still called via `./robo phar:build`, as usual. "Box" is what most people use; this tool requires the user to make a `box.json` or `box.json.dist` file that describes how the phar should be built.  It gets \Phar::running() right, and also makes the php compactor available, which cuts the size of the phar nearly in half (from 2.0M to 1.1M).

Before removing taskPackPhar, I did try to fix the \Phar::running() bug, but was not successful. I do not think it is sustainable / maintainable to continue to call the \Phar APIs directly in Robo. An alternate idea would be to re-introduce `taskPackPhar`, but implement it using [box2-lib](https://github.com/box-project/box2-lib) rather than via direct use of the \Phar APIs. This would be a fair bit of work, though, and it is unclear whether the result would be any better than just having folks maintain a box.json file.

My suggestion is to go with the box tool for now, and consider bringing back taskPackPhar later, as an enhancement, if anyone has enough need for it to spend the time to implement it.


DavertMik:
> This PR removes taskPackPhar entirely, and replaces it with the box.phar tool

I don't like this switch. I'm still building phars for Codeception and Robo with it and I didn't have problems for a while. For sake of Codeception project I'd like to keep things as they are as they worked good at least for some time. I'd rather not do this change for no good reason.

I think we can still have `$this->_exec('box')` if someone needs better phars.

>  and also makes the php compactor available, which cuts the size of the phar nearly in half (from 2.0M to 1.1M)

Not sure about compactor, I think that is the internal phar limiatation. If more than 1000 files are added compression is disabled. 

So if you want you can keep `box` and `phar:build` using box but please keep `packPhar` task. Let's have it, at some point we can deprecate it in favor of something new, but currently it just works. If it doesn't fit for someone we just recommend to use box instead

greg-1-anderson:
It doesn't work correctly, because `\Phar::running()` returns the wrong result. This change allows us to [fix the robo wrapper](https://github.com/consolidation-org/Robo/pull/411/files#diff-6d67f24e6f2cff6ed5b0495fab78618aR8).

The workaround for this bug in the robo wrapper isn't horrible, though, so I am fine closing this PR, and deferring this issue until I have time to use box2-lib to fix taskPackPhar in a backwards-compatible way.

PR #412, opened 2016-08-18
Remove services

greg-1-anderson:
This PR removes the concept of task "services" from Robo.

**Motivation**

Prior to this PR, Robo used the DI container as a service provider. This was initially a convenience, as Robo's subclass of the Container handled features such as the simulated interface, and allowed the dependencies for each task to be automatically populated.  However, this model is contrary to the [PHP FIG recommended usage](https://github.com/container-interop/fig-standards/blob/master/proposed/container.md#13-recommended-usage) for DI containers.

With the addition of `inflect` and container builders, the need to use the DI container as a service provider is negligible.  This PR therefore removes all references to the container from robo tasks, in favor of direct instantiation of the respective class objects + inflection in the collection builder.

**Benefits**

It is no longer necessary to declare services in `loadTasks`, eliminating a redundancy in task creation that Robo 1.0 had introduced.

Creating subtasks within a task can now be done via `$this->collectionBuilder()->taskFoo()`, increasing uniformity between task code and RoboFile command code.  (n.b. `$this->taskFoo()` is still only supported in RoboFiles, not in tasks.)


src/Task/Archive/loadTasks.php
```diff
@@ -1,31 +1,16 @@
...
-
-    /**
      * @param $filename
      *
      * @return Pack
      */
     protected function taskPack($filename)
     {
-        return $this->task(__FUNCTION__, $filename);
+        return $this->task(Pack::class, $filename);
```
DavertMik: yeah, I like the syntax much more

DavertMik:
Good :+1:

DavertMik:
Thanks!

Issue #414, opened 2016-08-22
output() or getOutput()


greg-1-anderson:
Although there is still some variation in style, nowadays php applications tend to use `getFoo()` for variable accessors, and `foo()` for service accessors.  I followed this convention when I wrote an [OutputAwareTrait](https://github.com/consolidation-org/Robo/blob/560cec46dfe909f3011c0a8574f8d96f042cee34/src/Common/OutputAwareTrait.php#L27) for Robo.

However, Robo has historically used [getOutput()](https://github.com/consolidation-org/Robo/blob/0.6.0/src/Common/IO.php#L18) for direct access to the output object in the IO trait.  I actually didn't intend to commit the aforementioned OutputAwareTrait with #412; I thought this was part of a working branch still in progress, where I am working on unifying the IO trait with the InputAwareTrait and OutputAwareTrait.

My thought is that not too many Robo scripts directly use `getOutput()` and `getInput()`, and for those that do, it would be easy enough to switch these calls to `output()` and `input()`, respectively.  (For what it's worth, Symfony provides neither variant in InputAwareHelper, and expects clients to just use `$this->input` directly.)  My preference would be to rename `getOutput()` to `output()`, and similarly for the input accessor, and have folks adapt in 1.0. Let me know if you'd prefer to stick with `getInput()` for backwards compatibility, though.

greg-1-anderson:
Note that the Symfony Process component has a method `getOutput()`. This is a different sort of method, though -- it returns the output from the process (datatype string).  Having the IO method be called `output()` (meaning an interface you can write to) therefore seems clearer to me.

greg-1-anderson:
On the flip side, Symfony\Component\Console\Event\ConsoleCommandEvent has a `getInput()` method.

I still favor `output()` and `input()`.

greg-1-anderson:
#416 renames `getOutput()` to `output()` everywhere it appears in Robo core, but maintains `getOutput()` (in one location) as a synonym for backwards compatibility.

greg-1-anderson:
Discuss in PR.

PR #416, opened 2016-08-22
IO interface

greg-1-anderson:
Rename getOutput() and getInput() to output() and input(); remove globals from the IO trait, and build IO off of InputAwareTrait and OutputAwareTrait.

`getInput()` and `getOutput()` methods maintained for backwards compatibility.


greg-1-anderson:
See background in #414.

PR #418, opened 2016-08-22
Fix: disallow passing multiple test files to PHPUnit via files()

abusque:
Unfortunately, PHPUnit does not support passing a list of test files
to run via the command-line interface. However, the current
implementation of the PHPUnit task allowed passing an array of files
to its `files()` method, which it would then append to the command as
a string of comma-delimited files, which would cause the PHPUnit
executable to fail.

The PHPUnit executable supports either a path to a directory
containing multiple test files or a path to a single test file as a
command-line argument.

This fixes the issue by allowing only a single string to be passed via
`files()`.

Signed-off-by: Antoine Busque <antoine.busque@pillrcompany.com>

### DIFF
--- a/docs/tasks/Testing.md
+++ b/docs/tasks/Testing.md
@@ -95,7 +95,7 @@ $this->taskPHPUnit()
 * `bootstrap($file)` 
 * `configFile($file)` 
 * `debug()` 
-* `files($files)`  Test files to run.
+* `files($files)`  Directory of test files or single test file to run.
 * `dir($dir)`  changes working directory of command
 * `printed($arg)`  Should command output be printed
 * `arg($arg)`  Pass argument to executable
--- a/src/Task/Testing/PHPUnit.php
+++ b/src/Task/Testing/PHPUnit.php
@@ -25,7 +25,8 @@ class PHPUnit extends BaseTask implements CommandInterface, PrintedInterface
     protected $command;
 
     /**
-     * Test files to run, they're appended to the command and arguments.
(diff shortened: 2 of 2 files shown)


abusque:
Arguably the `files` (plural) nomenclature might be misleading as we can only specify a single path, although technically passing a directory will result in multiple test files being executed.

greg-1-anderson:
Thanks for pointing this out.  For a moment, I thought that changing the list from comma-separated to a space-separated list would clear up the problem.  Making this change (c.f. 9f3b5cf4, which I have since backed out) does cause PHPUnit to stop complaining, and run tests. Unfortunately, it only considers the first parameter provided, and ignores the rest.

@DavertMik: Do you think we should have separate `file()` and `dir()` methods, for clarity?  Both methods could do the same thing (`dir()` could just call through to `file()`), and we could fail with an explanitory message if anyone tried to specify more than one file/dir parameter.

abusque:
Thanks for reviewing. My first though was to try the space-separated as well, I should've mentioned that in my original message. It is slightly unexpected behaviour (in my opinion) on PHPUnit's part, especially considering that it just runs one of the listed tests without any warning that the rest has been ignored.

I do feel that `file()` and `dir()` would be less misleading than the current `files()`

DavertMik:
> Do you think we should have separate file() and dir() methods, for clarit

Yes, please add `dir` and `file` methods. 
`files` method should be deprecated.

greg-1-anderson:
Will fix ^^ in a follow-on commit

PR #420, opened 2016-08-25
Supply application name and version when using Robo as a framework

greg-1-anderson:
Allow applications using Robo as a framework to supply thier own application name and version (if they let Robo set up the DI container), or even a complete application object (if they set up their own DI container).

This was an oversight.


DavertMik:
Looks good

PR #424, opened 2016-08-30
Add Behat

duxet:


greg-1-anderson:
Great idea -- +1 on merging this.

Sorry about the VersionEye failure here.  I wanted to do a VersionEye analysis on Robo, and the process of doing that caused VersionEye to be added as another check for all PRs as an unintentional side-effect. codeception/verify has a license file in dev-master, now, and aspect-mock already had one in a newer build than the one we're using.  I'll see if I can just fix the failure (separately), so that we don't need to de-configure VersionEye. I'd like to add a VersionEye badge to the README for documentation purposes. It'd be no good to do this if it made all of our PR tests marked as failed, though.

duxet:
No problem, it's a nice tool though, i might use it with my project too :)

greg-1-anderson:
Almost fixed. :)

PR #427, opened 2016-08-31
Fix versioneye

greg-1-anderson:
Update to codeception/aspect-mock ~1, and dev-master of codeception/verify to get license files for the last couple of components that are missing them.

This should make VersionEye come back green.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,9 +32,9 @@
         "natxet/CssMin": "~3",
         "pear/archive_tar": "~1",
         "codeception/base": "^2.1.5",
-        "codeception/verify": "dev-master",
+        "codeception/verify": "^0.3",
         "codeception/aspect-mock": "~1",
-        "satooshi/php-coveralls": "^1",
+        "satooshi/php-coveralls": "~1",
         "squizlabs/php_codesniffer": "~2",
         "phpunit/php-code-coverage": "~4"
     },
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "codeception/aspect-mock": "~1",
         "satooshi/php-coveralls": "^1",
         "squizlabs/php_codesniffer": "~2",
-        "phpunit/php-code-coverage": "~2"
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
Drat!  Apparently, VersionEye is still checking the latest tagged version (0.3.1), and does not pick up the license in the composer.json file in dev-master.

@DavertMik: Maybe you could tag a 0.3.2 release to pick up the [license element in composer.json](https://github.com/Codeception/Verify/blob/master/composer.json#L4), so that we have a clean dependency analysis for Robo?

composer.json
```diff
@@ -32,8 +32,8 @@
         "natxet/CssMin": "~3",
         "pear/archive_tar": "~1",
         "codeception/base": "^2.1.5",
-        "codeception/verify": "0.2.*",
-        "codeception/aspect-mock": "0.5.4",
+        "codeception/verify": "dev-master",
+        "codeception/aspect-mock": "~1",
```
greg-1-anderson: n.b. codeception/aspect-mock: ~2 caused failues, but ~1 worked out-of-the-box and provided us with a license in composer.json, so I did not take the time to upgrade to version 2.

greg-1-anderson:
I tried PHP 7.1, and got this error message:
```
Codeception PHP Testing Framework v2.2.4
Powered by PHPUnit 5.4.8 by Sebastian Bergmann and contributors.
                                                     
  [SebastianBergmann\CodeCoverage\RuntimeException]  
  No code coverage driver available       
```
Also, updating to the latest dev components broke our build, because not all of our dependencies work with php 5.5 in their latest versions. This makes it a little more challenging to (a) stay absolutely up-to-date, and (b) support older versions of PHP.  5.5 is EOL now, but it is common for older versions of PHP to stick around for a while (e.g. on low-cost hosting providers).

greg-1-anderson:
I'm going to go ahead and merge this so that we're one step closer to up-to-date on our dependencies.

PR #429, opened 2016-09-01
Fixes loading the _exec() shortcut. 'Class taskExec does not exist'.

pingers:
Any RoboFile using the _exec() shortcut is broken it seems.

E.g.
```
$  robo build
Class taskExec does not exist
```

### DIFF
--- a/src/Task/Base/loadShortcuts.php
+++ b/src/Task/Base/loadShortcuts.php
@@ -11,6 +11,6 @@ trait loadShortcuts
      */
     protected function _exec($command)
     {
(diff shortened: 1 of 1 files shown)


pingers:
Example RoboFile.php
```php
<?php
/**
 * This is project's console commands configuration for Robo task runner.
 *
 * @see http://robo.li/
 */
{
    public function build() {
        $this->exec('ls');
    }
}
```

greg-1-anderson:
Whoops, missed one!  Thanks for finding this.

PR #431, opened 2016-09-01
Introduce conditional tasks.

greg-1-anderson:
As the comment in the code says:

> Usually, it is best if a Robo command does all of its conditional
> testing using procedural code, before creating any tasks.  However,
> sometimes a particular test cannot run until some prior tasks
> finish running.  In those instances, there are two options:
> 
>  (1) Run one collection to set up, and then run any needed tests,
>      and build a second collection of tasks based on the result.
> 
>  (2) Use a Conditional task.
> 
>  The conditional task creates a task wrapper that holds three collections:
> 
>  check - runs unconditionally and records its result
>  onTrue - runs if the check collection returns successfully
>  onFalse - runs if the check collection fails
> 
>  Optionally, a 'test' function may be provided.  If it is,
>  then it will be passed the result object from the check
>  collection.  The result of the test function will be used
>  to determine whether the 'onTrue' or 'onFalse' collection
>  is called.
>

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -698,6 +698,28 @@ class RoboFile extends \Robo\Tasks
                 )
             ->run();
     }
+
+    public function tryConditional($file = 'RoboFile.php', $regex = 'robo')
+    {
+        return $this->collectionBuilder()->
+            taskConditional(
+                    $this->collectionBuilder()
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
Didn't need this for the use-case I was writing it for. May return to this later if needed.

PR #434, opened 2016-09-02
Always escape arguments CommandArguments.

greg-1-anderson:
One more change, for consistency.

Arguments passed to `->arg()` and values passed to `->option()` are always escaped. This is better and safer, and makes option() and optionList() consistent. `->rawArg()` is available for clients that have a value that has already been shell-escaped.

Note that entire command strings, e.g. those passed to the constructor of Exec(), are expected to be already escaped by the client.


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 * Use league/container to do Dependency Injection
    * *Breaking* Tasks' loadTasks traits must use `$this->task(TaskClass::class);` instead of `new TaskClass();`
    * *Breaking* Tasks that use other tasks must use `$this->collectionBuilder()->taskName();` instead of `new TaskClass();` when creating task objects to call. Implement `Robo\Contract\BuilderAwareInterface` and use `Robo\Contract\BuilderAwareTrait` to add the `collectionBuilder()` method to your task class.
+* *Breaking* The `arg()`, `args()` and `option()` methods now escape the values passed in to them.
+* *Breaking* taskWrite is now called taskWriteToFile
```
DavertMik: you should mention that `rawArg` was introduced

PR #440, opened 2016-09-04
Add accessor for SymfonyStyle object.

greg-1-anderson:
The SymfonyStyle class is very popular; see the [introductory documentation](http://symfony.com/blog/new-in-symfony-2-8-console-style-guide) for details.

This is how folks expect to do IO with Symfony Console applications nowadays, so we should provide ready access to it from RoboFiles.  It might even be a good idea to update the documentation to use SymfonyStyle for IO.

The existing methods in the IO trait may remain for backwards compatibility with Robo 0.x; perhaps these should be deprecated and removed in 2.x.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -316,6 +316,12 @@ $name = $this->ask("What is your name?");
 
 There are also `askDefault`, `askHidden`, and `confirm` methods.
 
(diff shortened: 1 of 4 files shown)


DavertMik:
That's really cool. But should be documented

greg-1-anderson:
I'll document this as an advanced output option somewhere nearish the bottom of getting started.

PR #447, opened 2016-09-12
Allow clients to provide their own config subclass to Robo.

greg-1-anderson:
Small improvements to Runner API for clients like Terminus.  Drush might use this too.


greg-1-anderson:
Argh, I accidentally merged this PR to master.

Ah well, I think it's a good change anyway.  I'm just going to go fix the other projects I broke now.

PR #448, opened 2016-09-13
Create Robo::run() static wrapper method.

greg-1-anderson:
Not sure if this is enough of an improvement.  Just going to put this here for now and think about it. Comments welcome.

### DIFF
--- a/docs/framework.md
+++ b/docs/framework.md
@@ -20,7 +20,7 @@ Next, add an "autoload" section to your composer.json to provide a namespace for
     }
 }
 ```
-Create a new file for your Robo commands, e.g. `class RoboCommands` in `namespace MyProject\Commands;` in the file `src\Commands\RoboCommands.php`.  Optionally, add more task libraries as described in the [extending](extending.md) document.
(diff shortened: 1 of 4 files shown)


docs/framework.md
```diff
@@ -42,10 +42,13 @@ if ($pharPath) {
...
-$runner = new \Robo\Runner($commandClasses);
-$output = new \Symfony\Component\Console\Output\ConsoleOutput();
-$statusCode = $runner->execute($_SERVER['argv'], $output, 'MyAppName', '0.0.0-alpha0');
+$commandClasses = [ \MyProject\Commands\RoboFile::class ];
+$statusCode = Robo::run(
+    $_SERVER['argv'], 
+    $commandClasses, 
+    'MyAppName', 
+    '0.0.0-alpha0'
+);
```
DavertMik: I like it, documentation is simplified. However there should be a way to disable default `\Symfony\Component\Console\Output\ConsoleOutput` and to replace it with something else. This should be mentioned in docs

greg-1-anderson: Yeah, that bothered me too, but adding another parameter seemed like it was making the parameter list too long. I agree with you, though; I'll put `$output = null` on the very end; then it won't bother anyone using the typical use case, and will also be testable.

I'll also fix the missing `function` typo. ;)

PR #454, opened 2016-09-27
Solves undefined function bug when calling taskImageMinify

chriscct7:
see title


greg-1-anderson:
Thanks!  It would be great to have some tests for the image minify task. Appreciate you fixing up this error for us.

PR #461, opened 2016-10-10
Commands may now implement ContainerAwareInterface and access the con…

marktheunissen:
…tainer through the methods provided


greg-1-anderson:
Oh, good catch!  We want to avoid commands implementing ContainerAwareInterface unless they need to; where it is necessary, though, it should be possible, so this should definitely be added to the container.

greg-1-anderson:
@DavertMik FYI merged as a bugfix.

PR #463, opened 2016-10-13
FIX notice

kusnir:
PHP Notice:  Use of undefined constant EXITCODE_USER_CANCEL - assumed 'EXITCODE_USER_CANCEL' in .../src/ResultData.php on line 82


greg-1-anderson:
Thanks.

PR #473, opened 2016-10-16
Fix type hints in PhpDocs

Sweetchuck:
### Motivation

From developer point of view (DX) not comfortable to work with Robo because I have to know by heart  all the methods and their arguments because my IDE (PhpStorm) can't give me autocomplete suggestions.

### Summary
Improved type hints in PhpDoc blocks

### DIFF
--- a/src/Collection/CollectionBuilder.php
+++ b/src/Collection/CollectionBuilder.php
@@ -1,9 +1,11 @@
 <?php
 namespace Robo\Collection;
 
+use Guzzle\Inflection\InflectorInterface;
 use Robo\Config;
 use Robo\Common\Timer;
 use Psr\Log\LogLevel;
+use Robo\Contract\InflectionInterface;
 use Robo\Contract\TaskInterface;
 use Robo\Contract\CompletionInterface;
 use Robo\Contract\WrappedTaskInterface;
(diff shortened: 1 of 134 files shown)


src/Collection/CollectionBuilder.php
```diff
@@ -1,9 +1,11 @@
 <?php
 namespace Robo\Collection;
 
+use Guzzle\Inflection\InflectorInterface;
```
greg-1-anderson: This seems to not be needed.

src/Collection/CollectionBuilder.php
```diff
@@ -369,6 +446,9 @@ public function getCommand()
         return $this->getCollection()->getCommand();
     }
 
+    /**
+     * @return \Robo\Collection\Collection
+     */
     public function original()
     {
         return $this->getCollection();
```
greg-1-anderson: This should be a TaskInterface, like in the parent class.

src/Common/ProgressIndicatorAwareTrait.php
```diff
@@ -4,18 +4,31 @@
...
+    /**
+     * @return int
+     */
     public function progressIndicatorSteps()
     {
         return 0;
     }
 
+    /**
+     * @param \Robo\Common\ProgressIndicator $progressIndicator
```
greg-1-anderson: There are places that call `setProgressIndicator(null)`; if you are going to typehint this, then we'll need to provide a `clearProgressIndicator()` method.

src/Task/Development/loadTasks.php
```diff
@@ -22,7 +24,8 @@ protected function taskGenDoc($filename)
     }
 
     /**
-     * @param $filename
+     * @param string $filename
```
greg-1-anderson: This docblock does not match the function parameters.

greg-1-anderson:
Wow, thanks for adding all of these.  I made a couple of comments on the diff, but I'm going to go ahead and merge this anyway. Feel free to make a follow-on PR for those, if you're up for it.

PR #480, opened 2016-11-01
Fix a bug in PHPUnit task: specifying a directory caused infinite loop.

dlancea:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature
- [x] Needs tests

### Summary
Fix a bug in PHPUnit task: specifying a directory caused infinite loop.

### Description
The dir function is currently only calling itself, causing an infinite loop. I'm assuming the intention was to call ExecOneCommand::dir instead. 

This fix relieves the method naming conflict and calls the aliased method.

### DIFF
--- a/src/Task/Testing/PHPUnit.php
+++ b/src/Task/Testing/PHPUnit.php
@@ -20,7 +20,9 @@ use Robo\Task\BaseTask;
  */
 class PHPUnit extends BaseTask implements CommandInterface, PrintedInterface
 {
-    use \Robo\Common\ExecOneCommand;
+    use \Robo\Common\ExecOneCommand {
+        dir as execOneCommandDir;
+    }
 
     /**
      * @var string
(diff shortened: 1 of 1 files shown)


src/Task/Testing/PHPUnit.php
```diff
@@ -189,7 +191,7 @@ public function file($file)
      */
     public function dir($dir)
     {
-        return $this->dir($dir);
+        return $this->execOneCommandDir($dir);
```
greg-1-anderson: Couldn't we just completely delete the `dir` function? It should do the right thing then. It would be helpful if this was confirmed with a unit test.

dlancea:
You're right, my method was not needed and removing the PHPUnit::dir method was easier. 

I looked into adding a test for this but I figured 1) I'm not sure how to test that, given that the dir option does not change the command and 2) at this point that would really be testing ExecCommand::dir(), which should already be tested.

PR #493, opened 2016-11-15
Fix: look for project bin directory in current working directory

abusque:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature
- [ ] Needs tests

### Summary
This allows a globally installed Robo to resolve an executable
installed locally in `vendor/bin` under the current working directory.

### Description
The use case that lead me to this fix is having a locally installed phpunit, and a globally installed robo. Prior to the fix, it would fail to resolve the phpunit executable, whereas it resolved it just fine prior to version 1.x.

Signed-off-by: Antoine Busque <antoine.busque@pillrcompany.com>

### DIFF
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -135,7 +135,8 @@ trait ExecCommand
      */
     protected function findProjectBin()
     {
-        $candidates = [ __DIR__ . '/../../vendor/bin', __DIR__ . '/../../bin' ];
+        $cwd = getcwd();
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
LGTM. @DavertMik can you think of any reason we shouldn't do this?  `__DIR__` == `getcwd()` in typical (local) usage, so there shouldn't really be any way this could go wrong, I think.

greg-1-anderson:
Seems safe. Merging.

PR #498, opened 2016-11-28
Removed redundant @method tags & unused imports

Korikulum:
### Overview
This pull request:

- [ ] Fixes a bug
- [ ] Adds a feature

### Summary
Removed redundant @method tags & unused imports.

### Description
Most @method tags are redundant since, in the meantime, those methods were rewritten to not be magic ones. I've moved all the descriptions to the phpdoc for the actual methods. Also updated the @method tags on the `FilesystemStack` class to match the Symfony filesystem methods.

Removed all unused imports & those that tried to import classes that have been removed.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -1,7 +1,5 @@
 <?php
 use Symfony\Component\Finder\Finder;
-use Robo\Result;
-use Robo\Collection\CollectionBuilder;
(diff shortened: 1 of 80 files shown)


greg-1-anderson:
Thanks for this.

PR #499, opened 2016-11-29
Fix Behat command

duxet:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Behat doesn't have run command, so i deleted it.

### Description
Fixes issue with Behat.


greg-1-anderson:
Tests run as written (prior to this PR), because it is mocked, but you are correct, this wouldn't run if it wasn't mocked.  Merging.

Issue #395, opened 2016-07-30
taskExec: Interactive Output


amenk:
When I run `n98-magerun.phar db:import` it is supposed to display progress information (it's calling `pv` internally). But it does not.

Can that be fixed?

amenk:
Can you give me a hint? Is it maybe already working via some option?

amenk:
It is strange ... when calling composerUpdate(), composer is displaying the progress in-line. But when calling `pv` or `wget` progress is not displayed / displayed in multiple lines instead of one (wget).

amenk:
What we need is a `process->setTty()`

PR #500, opened 2016-12-05
Allow to run exec tasks with input and/or interaction

### LINKED ISSUE #391: taskExec: supply stdin input
For example:

```
$this->taskExec('some-script-runner')->withStdIn("foo\nbar\n")->run()
```

HenningCash:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Currently all processes started with `Exec` read from `/dev/null` and can't be interacted with. This PR allows to pass a string/resource to the process and enables optional interaction.

Use-cases could be generators or password-prompts during deployment.

Will close #391

### DIFF
--- a/src/Task/Base/Exec.php
+++ b/src/Task/Base/Exec.php
@@ -67,6 +67,16 @@ class Exec extends BaseTask implements CommandInterface, PrintedInterface, Simul
     protected $process;
 
     /**
+     * @var resource|string
+     */
+    protected $input;
+
+    /**
+     * @var boolean
+     */
+    protected $interactive;
+    
+    /**
      * @param string|\Robo\Contract\CommandInterface $command
      */
     public function __construct($command)
@@ -133,6 +143,30 @@ class Exec extends BaseTask implements CommandInterface, PrintedInterface, Simul
         return $this;
     }
 
+    /**
+     * Pass an input to the process. Can be resource created with fopen() or string
+     *
+     * @param resource|string $input
+     *
+     * @return $this
+     */
+    public function input($input)
+    {
+        $this->input = $input;
+        return $this;
+    }
+
+    /**
+     * Attach tty to process for interactive input
+     *
+     * @return $this
+     */
+    public function interactive()
+    {
+        $this->interactive = true;
+        return $this;
+    }
+
     public function __destruct()
     {
         $this->stop();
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
+1 on the stdin redirection.

In terms of interactive mode, I think that we should default to a solution like https://github.com/pantheon-systems/terminus/issues/1507#issuecomment-272241035. We could allow the `interactive()` method to be used to force the tty allocation to `true` or `false`, but should revert to adopting the tty status of Robo's stdin if interactive is not called.

Robo should never set tty mode if `input()` is called.

However, perhaps `input()` should be renamed to `setInput()` (to be more like the Symfony Console method) or `stdin()` (to be less like the various methods of InputAwareTrait).

src/Task/Base/Exec.php
```diff
@@ -67,6 +67,16 @@ class Exec extends BaseTask implements CommandInterface, PrintedInterface, Simul
...
 
     /**
+     * @var resource|string
+     */
+    protected $input;
+
+    /**
+     * @var boolean
+     */
+    protected $interactive;
```
grasmash: Set default value to false

grasmash:
This should include an accompanying modification to the docs.

src/Task/Base/Exec.php
greg-1-anderson: I think that the default value should remain unset, to differentiate between unset and calling `->input(false);`

Then:

```
if (!isset($this->interactive) && function_exists('posix_isatty')) {
    $this->interactive(posix_isatty(STDOUT));
}
```

grasmash:
Opened #521 with requested changes.

greg-1-anderson:
Merged via #521

amenk:
Does this maybe also solve #395 ?

greg-1-anderson:
Yes, but this is also a good time to mention that calling `setTty()` when not needed can also cause problems. The current HEAD of master should solve your current use-case perfectly; however, you should call `->interactive()` when using `wget` or `pv`; otherwise, this will stop working after we remove the automatic `setTty()`, which will happen before the next stable tag.

amenk:
Okay, great. Maybe something like `interactiveAutoDetect()` can be added then with the current master-behavior?
For example I want to use `pv` as "interactive" as possible but still use the same Robofile on a CI server without TTY?

greg-1-anderson:
Yes, that is the intention.

I do wonder, though, if `interactive(true);` should simply be ignored / overridden when stdin is not a terminal (iff the method is available). Is an autodetect necessary?

amenk:
>  interactive(true); should simply be ignored / overridden when stdin is not a terminal

sounds good

greg-1-anderson:
@grasmash Can you think of any reason why [the suggestion above](https://github.com/consolidation/Robo/pull/500#issuecomment-274305896) might cause problems?

grasmash:
I prefer to have the ability to force interactive to true or false. There may be cases in which we do not correctly determine whether interactive(true) should truly be ignored.

amenk:
Interactive(true,true)?

PR #508, opened 2016-12-14
Use name variable for Githubrelease name

cboyden:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Use name variable for Github release name.

### Description
The GithubRelease task provides a name method to set the release name, but when creating the release, it uses the tag variable instead of the name variable.


greg-1-anderson:
Thanks.

PR #511, opened 2017-01-10
Change access to protected for IO methods.

grasmash:
In attempting to override IO methods such as yell() or ask(), I found that I was stymied by various IO methods having private access. Perhaps there is a reason for this, but I found it useful to be able to call methods such as formatQuestion() from my own overridden confirm() method. 

This makes it much easier, and DRYer, to override output formatting.


greg-1-anderson:
+1 yes this is a great idea.

PR #512, opened 2017-01-10
Adding printMetadata() option to exec command.

grasmash:
This PR introduces two new abilities:
- Use exec task _without_ printing command, working directory, and timer. This allows the task to be truly silent.`$this->taskExec("something")->silent(true)->run();`
- Use exec task to print output but _not_ print metadata, like the timer. This makes the call a bit more "passthru". `$this->taskExec("something")->printOutput(true)->printMetadata(false)->run();`

### DIFF
--- a/docs/tasks/Base.md
+++ b/docs/tasks/Base.md
@@ -56,7 +56,8 @@ $this->taskExecStack()
 * `stopOnFail($stopOnFail = null)`   * `param bool` $stopOnFail
 * `result($result)` 
 * `dir($dir)`  Changes working directory of command
-* `printed($arg)`  Should command output be printed
+* `printOutput($arg)`  Should command output be printed
+* `printMetadata($arg)`  Should command metadata (command, working directory, and timer) be printed
 
 ## ParallelExec
 
--- a/docs/tasks/Base.md
+++ b/docs/tasks/Base.md
@@ -58,6 +58,7 @@ $this->taskExecStack()
 * `dir($dir)`  Changes working directory of command
 * `printOutput($arg)`  Should command output be printed
 * `printMetadata($arg)`  Should command metadata (command, working directory, and timer) be printed
+* `silent($arg)`  Shortcut for setting printMetadata(false) and printOutput(false)
 
 ## ParallelExec
 
--- a/docs/tasks/Base.md
+++ b/docs/tasks/Base.md
@@ -56,6 +56,7 @@ $this->taskExecStack()
 * `stopOnFail($stopOnFail = null)`   * `param bool` $stopOnFail
 * `result($result)` 
 * `dir($dir)`  Changes working directory of command
+* `printed($arg)`  _Deprecated_. Should command output be printed
 * `printOutput($arg)`  Should command output be printed
 * `printMetadata($arg)`  Should command metadata (command, working directory, and timer) be printed
 * `silent($arg)`  Shortcut for setting printMetadata(false) and printOutput(false)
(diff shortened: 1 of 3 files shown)


docs/tasks/Base.md
```diff
@@ -56,7 +56,9 @@ $this->taskExecStack()
 * `stopOnFail($stopOnFail = null)`   * `param bool` $stopOnFail
 * `result($result)` 
 * `dir($dir)`  Changes working directory of command
-* `printed($arg)`  Should command output be printed
```
grasmash: This was removed from docs but is still present in code to prevent a backwards incompatible change.

greg-1-anderson: If you want to rename a function, mark it deprecated, don't take it out of the docs entirely.

greg-1-anderson:
I am generally in favor of this PR; however, I wonder if you could expand on your use-case where you're planning on using this.

For example, `silent(true)` + `getResult()` together is probably a use-case that should be avoided. If you are `exec`ing something just to get information (not change the state of the system), then you're better off just calling the Symfony Process Component (or even just `passthru`) directly.

In contrast, the `exec` task is best for commands that should be run as part of a collection to change the state of the system. If you want to silence these commands because they are very verbose (e.g. `composer update`, or a large git  push), then that can already be accomplished with `printed(false);`. Unless I am wrong about this -- you said "this allows the command to be truly silent". Is Robo printing the metadata currently w/ printed(false)?

I sort of think we should get rid of `getResult()`, though, because I don't want to encourage folks to set up and use a task when they're better off doing their work with regular PHP code. In RoboPHP, the general pattern is:

1. Run some normal php code to determine what needs to be done.
2. Prepare a collection of tasks.
3. Run the tasks.

Generally, you can have conditionals as needed in 1. and 2., and `exec`s can happen as needed in either 1. or 3., but you should avoid conditionals in 3.

grasmash:
>  Is Robo printing the metadata currently w/ printed(false)?

Yes. It still shows the command being executed and also the execution time.

greg-1-anderson:
Do you need `getResult()`, or would you be okay with resubmitting this without that method?

grasmash:
@greg-1-anderson I couldn't find any method named `getResult()` in the `src` directory. Are you referring to `getResultData()` introduced by this PR? That method is protected and used only internally by the command in order to make it DRYer.

grasmash:
If the intention is for the command to be totally silent when `printed()` is FALSE, then I can refactor this PR to make that the case rather than introducing new methods.

docs/tasks/Base.md
```diff
@@ -56,7 +56,10 @@ $this->taskExecStack()
...
+* `printed($arg)`  _Deprecated_. Should command output be printed
```
greg-1-anderson: This is good. Might also add "Use printOutput() instead."

greg-1-anderson:
Yes, my mistake. `getResultData()` is fine as-is.

+1

@DavertMik Any feedback on this change?

src/Common/ExecCommand.php
```diff
@@ -77,6 +98,33 @@ public function printed($arg)
...
     /**
+     * Should command output be printed
+     *
+     * @param bool $arg
+     *
+     * @return $this
+     */
+    public function printOutput($arg)
+    {
+        return $this->printed($arg);
```
greg-1-anderson: Could you reverse this logic, and have `printed` call `printOutput`? @DavertMik would like`printed()` to emit a warning via the logger if someone calls it, but this clearly should not happen if someone calls the recommended method (`printOutput`).

greg-1-anderson:
Looks great -- thanks!

PR #513, opened 2017-01-12
Website should use updated namespace consolidation

pniederlag:
the namespace of the package was changed from codegyre to consolidation a while ago. documenation still had old name.


greg-1-anderson:
So, the file changed in this PR is a generated file. The source file is [robotheme/home.html]($container->add('container') ) in the `site` branch.

greg-1-anderson:
This has been updated on robo.li.

Issue #517, opened 2017-01-13
Document example of overriding output format


grasmash:
From #519:
> Is there a mechanism in Robo for defining a custom logger that would permit customization of the output format for core tasks?

Documenting an example of how to do this would be great.

greg-1-anderson:
Sorry, could you be more clear on what you are trying to do here? An output format writes structured output to stdout -- something that can be transformed with the `--format` option. The logger, on the other hand, writes messages to stderr.

Robo, the executable, does not support customizing the dependency injection container from a RoboFile script. This can only be done when writing your own application based on Robo. See [Robo as a Framework](http://robo.li/framework/) for more details on this. I can expand on how to create a custom logger in a Robo-based application, if that's what you need to do. Let's be clear on the goals and use-case, though, as I think it might be better to use standard Robo output for Robo scripts, and maybe enhance how Robo does output if the use-case is generally useful for scripts.

grasmash:
Yes, I do need to clarify. 

Some specific goals:

1. Override say() and yell() formatting. This is now possible thanks to #511 being merged.
2. Set the verbosity level for any task. This is related to #512, but considered at a higher level . 

    It would be great if we could set the verbosity level of _any task's_ output. E.g, I'd like a given implementation of taskExec() to display it's output only if `-v` has been passed, and otherwise print nothing to the screen. `$this->taskExec('compass')->setVerbosity(SOME_CONSTANT);`.

    Likewise, `$this->taskCleanDir(['tmp','logs'])->setVerbosity(SOME_CONSTANT)->run();` to display the output of taskCleanDir() only at a specified verbosity level.
3. Output messages at specific points in the command execution process. 

    Ideally, I could define a hypothetical custom class, implement methods like `$this->beforeCommand()`, `$this->beforeTask()`, `$this->afterTask()`, and `$this->afterCommand()` to effectively add a prefix and suffix output before all tasks, after all tasks, etc.

Now that the goals have been expressed, this seems like multiple issues. Perhaps we can begin by indicating how 2 and 3 might be accomplished?

greg-1-anderson:
2. would be difficult to achieve from a plugin. I think this would be a good feature addition to Robo core.

[...]

PR #522, opened 2017-01-20
Connected to #517: Adding setVerbosityLevel() for tasks.

grasmash:
So, this PR technically achieves the second item in #517:
> Set the verbosity level for any task. This is related to #512, but considered at a higher level .
>
> It would be great if we could set the verbosity level of any task's output. E.g, I'd like a given implementation of taskExec() to display it's output only if -v has been passed, and otherwise print nothing to the screen. $this->taskExec('compass')->setVerbosity(SOME_CONSTANT);.

However, actually calling it is wonky. E.g., if I wanted to make use of the taskFilesystemStack's mkdir() method and set its verbosity, I would:

```php
<?php
$filesytemStack = $this->taskFilesystemStack();
$filesytemStack->getBuilder()->getCollectionBuilderCurrentTask()->setVerbosityLevel(LogLevel::INFO);
$filesytemStack->mkdir($dir)->run();
```

We can't simply `$this->taskFilesystemStack()->setVerbosityLevel(LogLevel::INFO)->mkdir($dir)` because `taskFilesystemStack()` actually returns an instance of CollectionBuilder and not the instance of FileSystemStack. The verbosity of the CollectionBuilder doesn't percolate into the FileSystemStack.

@greg-1-anderson Maybe you can  think of a better way to do this. There's a lot of complexity between the CollectionBuilder, calls the _call(), stacks, and their tasks.

### DIFF
--- a/src/Collection/CollectionBuilder.php
+++ b/src/Collection/CollectionBuilder.php
@@ -71,6 +71,21 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
     }
 
     /**
+     * @param int $verbosityLevel
+     *
+     * @return $this
+     */
+    public function setVerbosityLevel($verbosityLevel)
+    {
+        $this->verbosityLevel = $verbosityLevel;
+        if (method_exists($this->currentTask, 'setVerbosityLevel')) {
+            $this->currentTask->setVerbosityLevel($verbosityLevel);
+        }
+
+        return $this;
+    }
+
+    /**
      * @param bool $simulated
      *
      * @return $this
@@ -342,6 +357,9 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
     {
         $reflection = new ReflectionClass($name);
         $task = $reflection->newInstanceArgs($args);
+        if (method_exists($this->currentTask, 'setVerbosityLevel')) {
+            $this->currentTask->setVerbosityLevel($this->verbosityLevel);
+        }
         if (!$task) {
             throw new RuntimeException("Can not construct task $name");
         }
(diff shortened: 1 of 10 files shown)


grasmash:
Alright, added one more commit. This now supports the following usage:
```php
<?php
   $this->taskFilesystemStack()
     ->setVerbosityLevel(LogLevel::INFO)
     ->mkdir($dir)
     ->run();
```

Though frankly I haven't used the CollectionBuilder in all of the ways that one can, so I don't know if this will cause problems with different implementation.

src/Collection/CollectionBuilder.php
```diff
@@ -71,6 +71,17 @@ public function __construct($commandFile)
     }
 
     /**
+     * @param int $verbosityLevel
+     */
+    public function setVerbosityLevel($verbosityLevel)
+    {
+        $this->verbosityLevel = $verbosityLevel;
+        $this->currentTask->setVerbosityLevel($verbosityLevel);
```
grasmash: What I really want is to set the verbosity for all tasks belonging to this collection, but I couldn't see how to access them all. $this->collection is null even when $this->currentTask is not.

src/Collection/CollectionBuilder.php
```diff
@@ -71,6 +71,21 @@ public function __construct($commandFile)
...
 
     /**
+     * @param int $verbosityLevel
+     *
+     * @return $this
+     */
+    public function setVerbosityLevel($verbosityLevel)
+    {
+        $this->verbosityLevel = $verbosityLevel;
+        if (method_exists($this->currentTask, 'setVerbosityLevel')) {
```
grasmash: I don't love checking for the method. The alternative is to add `setVerbosityLevel()` and `getVerbosityLevel()` to `TaskInterface`

grasmash:
Note that errors and warnings from the task will still be printed at their normal verbosity levels. This seems like desired functionality, though I could see it argued either way.

grasmash:
Closing in favor of #525

PR #523, opened 2017-01-20
Connects to #517: Documenting how to style output.

grasmash:


greg-1-anderson:
This is a great docs page; thanks for writing it.  My only reservation is that this new page is really only applicable when creating your own application per "Robo as a Framework".

It seems like there may be a number of pages similar to this one that we might want to add over time. Just piling everything into the "Robo as a Framework" page is not ideal, as I imagine that it would eventually get to be pretty long if we did that.

It appears that [mkdocs now supports multiple levels of nesting](https://github.com/mkdocs/mkdocs/pull/482#issuecomment-98087113) in page menus. Maybe we could try that feature here?

grasmash:
@greg-1-anderson does this project use mkdocs? I don't see a mkdocs.yml file. Do I need to modify something else?

grasmash:
index.md updated.

greg-1-anderson:
Yes, you much switch to the `site` branch; then you will find the mkdocs.yml file for robo.li. It might be necessary to manually synchronize changes with mkdocs.yml and docs/index.md.  I'll have to build the docs and see if it works.

greg-1-anderson:
I confirmed that multiple-level submenus do in fact work. In the `site` branch, I changed `mkdocs.yml` like so:
```
-  - "Robo as Framework": framework.md
+  - "Robo as Framework":
+    - "Build an Application": framework.md
+    - "Customizing output": custom-output.md
```
That did the right thing. I'm still kind of ambivalent about whether a really long frameworks.md is better, or whether we should go ahead and use submenus. @DavertMik, do you have a preference?

koppieesq:
I'd like to do a time check on this discussion.  @greg-1-anderson I would recommend we go with submenus.  Do you agree?

greg-1-anderson:
I think submenus are fine. I'd like to deprecate `say` and `yell` and encourage folks to use standard SymfonyStyle methods. Could someone touch this up along those lines?

PR #524, opened 2017-01-20
Making printTaskProgress() json ouput more legible.

grasmash:
Currently, slashes are displayed as escaped when printed via printTaskProgress(). This is difficult to read and does not seem necessary since it's intended for humans.


greg-1-anderson:
Looks good. Thanks.

PR #525, opened 2017-01-23
Adding setLogLevel() to taskExec().

grasmash:
So, this PR achieves the second item in #517:
> Set the verbosity level for any task. This is related to #512, but considered at a higher level .
>
> It would be great if we could set the verbosity level of any task's output. E.g, I'd like a given implementation of taskExec() to display it's output only if -v has been passed, and otherwise print nothing to the screen. $this->taskExec('compass')->setVerbosity(SOME_CONSTANT);.

E.g., if I wanted to make use of the taskFilesystemStack's mkdir() method and set its log level to INFO, I would:
```php
<?php
   $this->taskFilesystemStack()
     ->logMetadata(true)
     ->logOutput(true)
     ->setLogLevel(LogLevel::INFO)
     ->mkdir($dir)
     ->run();
```

This would display task output only if the `-v` argument were passed to the command.

### DIFF
--- a/src/Collection/CollectionBuilder.php
+++ b/src/Collection/CollectionBuilder.php
@@ -71,6 +71,21 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
     }
 
     /**
+     * @param int $logLevel
+     *
+     * @return $this
+     */
+    public function setLogLevel($logLevel)
+    {
+        $this->logLevel = $logLevel;
+        if (method_exists($this->currentTask, 'setLogLevel')) {
+            $this->currentTask->setLogLevel($logLevel);
+        }
+
+        return $this;
+    }
+
+    /**
      * @param bool $simulated
      *
      * @return $this
@@ -342,6 +357,9 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
     {
         $reflection = new ReflectionClass($name);
         $task = $reflection->newInstanceArgs($args);
+        if (method_exists($this->currentTask, 'setLogLevel')) {
+            $this->currentTask->setLogLevel($this->logLevel);
+        }
         if (!$task) {
             throw new RuntimeException("Can not construct task $name");
         }
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -14,7 +14,12 @@ trait ExecCommand
     /**
      * @var bool
      */
-    protected $isPrinted = true;
+    protected $isOutputPrinted = true;
+
+    /**
+     * @var bool
+     */
+    protected $isOutputLogged = true;
 
     /**
      * @var bool
@@ -49,7 +54,7 @@ trait ExecCommand
      */
     public function getPrinted()
     {
-        return $this->isPrinted;
+        return $this->isOutputPrinted;
     }
 
     /**
@@ -67,7 +72,7 @@ trait ExecCommand
 
 
     /**
-     * Shortcut for setting isPrinted() and isMetadataPrinted() to false.
+     * Shortcut for setting isOutputPrinted, isOutputLogged, isMetadataPrinted.
      *
      * @param bool $arg
      *
@@ -76,7 +81,8 @@ trait ExecCommand
     public function silent($arg)
     {
         if (is_bool($arg)) {
-            $this->isPrinted = !$arg;
+            $this->isOutputPrinted = !$arg;
+            $this->isOutputLogged = !$arg;
             $this->isMetadataPrinted = !$arg;
         }
         return $this;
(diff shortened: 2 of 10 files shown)


src/Common/ExecCommand.php
```diff
@@ -233,13 +254,15 @@ protected function executeCommand($command)
...
-            $process->run(function ($type, $buffer) {
-                print $buffer;
-            });
-        } else {
-            $process->run();
-        }
+        $process->run(
+            function ($type, $buffer) {
+                if ($this->isOutputLogged) {
+                    $this->printTaskInfo($buffer);
```
greg-1-anderson: What is the use-case for logging the output instead of printing it?

src/Common/ExecCommand.php
```diff
@@ -233,13 +254,15 @@ protected function executeCommand($command)
...
+                } elseif ($this->isOutputPrinted) {
+                    print($buffer);
```
greg-1-anderson: It seems useful to be able to set the log level at which `exec` output is printed (e.g. if calling `exec(git merge);` on a the merge of a large build result set, perhaps you want to see that output only under -vvv mode). However, I am not convinced that we want to go so far as to add this same flexibility to all tasks, or to make the various changes to logging proposed here. If we could limit this PR to just the minimum changes needed to associate a log level with printing the output buffer, then I'd be in support of adding it.

src/Collection/CollectionBuilder.php
```diff
@@ -71,6 +71,21 @@ public function __construct($commandFile)
     }
 
     /**
+     * @param int $logLevel
+     *
+     * @return $this
+     */
+    public function setLogLevel($logLevel)
```
greg-1-anderson: Let's just put this on taskExec, and leave it out of the collection builder.

src/Common/ExecCommand.php
grasmash: If we print it, we can't control the log level at all. It's all or nothing.

Using logOutput() allows output to be printed conditionally, depending on the verbosity level.

src/Collection/CollectionBuilder.php
grasmash: If we leave this out of the collection builder, then we can't call this method on taskExec. The `taskExec()` method creates a task via the TaskAccessor, which uses the CollectionBuilder.

src/Common/ExecCommand.php
grasmash: I'm currently using this with both taskFilesystem() and taskExec(), and I can imagine this being useful for other tasks, like taskGitStack(). 

Without this, one needs to re-create the core tasks like taskFilesystem() in order to control the output. For me at least, this would represent a significant loss in the value of using Robo.

greg-1-anderson: Sounds plausible. I'll look at this PR again in light of that.

src/Collection/CollectionBuilder.php
greg-1-anderson: The collection builder will already call through to any method of the active task via the `_call()` special function. However, I have to re-consider my recommendation to account for multiple task support.

greg-1-anderson:
Could you list a few specific use-cases of output that you'd like to vary by log level?  e.g. `git clone`, `git merge`, `git push`.  Anything that's not calling through `exec`?

grasmash:
> Could you list a few specific use-cases of output (or log messages) that you'd like to vary by log level? 

Sure:
* taskFilesystem->mkdir() to create gitignored directories
* taskFilesystem->mirrorDir() to copy a large volume of files
* taskComposerInstall() to install dependencies
* Background check to see if Drupal is installed:
  taskExec("drush sqlq \"SHOW TABLES LIKE 'config'\"")

A couple of practical reasons for wanting to control the output level:

* In the past, I've received complaints from users regarding the volume of output for builds. If there is too much, it's difficult to parse and understand what the build is doing from a high level.
* Many CI tools have a maximum log size. `composer` and `git` and frequent culprits for exceeding this limit.

src/Collection/CollectionBuilder.php
grasmash: Hm. I tried removing this from Collection Builder but it breaks the functionality. It doesn't seem as though the property percolates into the required task object without this.

grasmash: I tried removing this from CollectionBuilder but got the error "[error]  Method setLogLevel does not exist."

greg-1-anderson:
I'm going to take a stab at re-doing this PR in a way that maintains the isolation between tasks and the logger, but still satisfies the use cases.

grasmash:
@greg-1-anderson Great. I think you're best suited to do the refactor.

One additional comment. It would be great if we implemented a solution that allowed users to take advantage of all of the verbosity levels provided by Symfony console commands.

Every robo command outputs the following help text:
>   -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

I personally find having more verbosity levels very helpful, and it seems logical to following the structure set forth by the underlying framework.

Either way, we should document the end functionality. I think the current is:

| args | max level output |
|-----|------------------|
|         | notice                  |
| -v     |  info                    |
| -vv   |                             |
| -vvv |  debug                |

greg-1-anderson:
Continued in #533.

PR #527, opened 2017-01-24
Disabling output for TaskExec() background processes.

grasmash:
Currently, a `TaskExec()` process with `background(true)` will still print output to screen. This does not seem like intended functionality. 

If we wanted, we could do this only when `printOutput(false)`.


greg-1-anderson:
Does `disableOutput` break commands that have output redirection to a file in the bash command string?

PR #531, opened 2017-01-25
Abstracting process execution logic into ExecTrait.php.

grasmash:
* Moved process execution logic outside of ExecCommand.php and Exec.php into ExecTrait.php, which is now used by them both.
* Added ProcessExecutor, which can be used outside of collection to execute arbitrary commands. 
    * ProcessExecutor extends BaseTasks so that it is able to print messages to screen and log messages.
* Added detectInteractive() method. `$this->interactive` now defaults to false.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ before_script:
   - if [ "$dependencies" = "highest" ]; then composer update --prefer-dist -n; fi;
 
 script:
-  - if [ "$dependencies" != "highest" ]; ./robo test
-  - if [ "$dependencies" = "highest" ]; ./robo test --coverage
+  - if [ "$dependencies" != "highest" ]; then ./robo test; fi;
+  - if [ "$dependencies" = "highest" ]; then ./robo test --coverage; fi;
 
 after_success:
   - if [ "$dependencies" = "highest" ]; travis_retry php vendor/bin/coveralls -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,9 @@ before_script:
   - if [ "$dependencies" = "lowest" ]; then composer update --prefer-dist --prefer-lowest -n; fi;
   - if [ "$dependencies" = "highest" ]; then composer update --prefer-dist -n; fi;
 
-script: "./robo test --coverage"
+script:
+  - if [ "$dependencies" != "highest" ]; ./robo test
+  - if [ "$dependencies" = "highest" ]; ./robo test --coverage
 
 after_success:
   - if [ "$dependencies" = "highest" ]; travis_retry php vendor/bin/coveralls -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_script:
 script: "./robo test --coverage"
 
 after_success:
-  - travis_retry php vendor/bin/coveralls -v
+  - if [ "$dependencies" = "highest" ]; travis_retry php vendor/bin/coveralls -v
 
 # Prior to a deploy, build a fresh robo.phar
 before_deploy:
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -124,7 +124,8 @@ trait ExecCommand
         return $cmd;
     }
 
-    public function getCommand() {
+    public function getCommand()
+    {
         return $this->process->getCommandLine();
     }
 
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -136,8 +136,7 @@ trait ExecCommand
      */
     protected function executeCommand($command)
     {
-        $this->process = new Process($command);
-        $result_data = $this->execute();
+        $result_data = $this->execute(new Process($command));
         return new Result(
             $this,
             $result_data->getExitCode(),
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -11,20 +11,7 @@ use Symfony\Component\Process\Process;
  */
 trait ExecCommand
 {
-    /**
-     * @var bool
-     */
-    protected $isPrinted = true;
-
-    /**
-     * @var bool
-     */
-    protected $isMetadataPrinted = true;
-
-    /**
-     * @var string
-     */
-    protected $workingDirectory;
+    use ExecTrait;
 
     /**
      * @var \Robo\Common\TimeKeeper
@@ -40,89 +27,6 @@ trait ExecCommand
             $this->execTimer = new TimeKeeper();
         }
         return $this->execTimer;
-    }
-
-    /**
-     * Is command printing its output to screen
-     *
-     * @return bool
-     */
-    public function getPrinted()
-    {
-        return $this->isPrinted;
-    }
-
-    /**
-     * Changes working directory of command
-     *
-     * @param string $dir
-     *
-     * @return $this
-     */
-    public function dir($dir)
-    {
-        $this->workingDirectory = $dir;
-        return $this;
-    }
-
-
-    /**
-     * Shortcut for setting isPrinted() and isMetadataPrinted() to false.
-     *
-     * @param bool $arg
-     *
-     * @return $this
-     */
-    public function silent($arg)
-    {
-        if (is_bool($arg)) {
-            $this->isPrinted = !$arg;
-            $this->isMetadataPrinted = !$arg;
-        }
-        return $this;
-    }
-
-    /**
-     * Should command output be printed
-     *
-     * @param bool $arg
-     *
-     * @return $this
-     */
-    public function printed($arg)
-    {
-        $this->logger()->warning("printed() is deprecated. Please use printOutput().");
-        return $this->printOutput($arg);
-    }
-
-    /**
-     * Should command output be printed
-     *
-     * @param bool $arg
-     *
-     * @return $this
-     */
-    public function printOutput($arg)
-    {
-        if (is_bool($arg)) {
-            $this->isPrinted = $arg;
-        }
-        return $this;
-    }
-
-    /**
-     * Should command metadata be printed. I,e., command and timer.
-     *
-     * @param bool $arg
-     *
-     * @return $this
-     */
-    public function printMetadata($arg)
-    {
-        if (is_bool($arg)) {
-            $this->isMetadataPrinted = $arg;
-        }
-        return $this;
     }
 
     /**
@@ -227,21 +131,6 @@ trait ExecCommand
      */
     protected function executeCommand($command)
     {
-        $process = new Process($command);
-        $process->setTimeout(null);
-        if ($this->workingDirectory) {
-            $process->setWorkingDirectory($this->workingDirectory);
-        }
-        $this->getExecTimer()->start();
-        if ($this->isPrinted) {
-            $process->run(function ($type, $buffer) {
-                print $buffer;
-            });
-        } else {
-            $process->run();
-        }
-        $this->getExecTimer()->stop();
-
-        return new Result($this, $process->getExitCode(), $process->getOutput(), ['time' => $this->getExecTimer()->elapsed()]);
+        return $this->execute($command);
     }
 }
--- a/src/Common/ExecCommand.php
+++ b/src/Common/ExecCommand.php
@@ -131,6 +131,13 @@ trait ExecCommand
      */
     protected function executeCommand($command)
     {
-        return $this->execute($command);
+
+        $result_data = $this->execute($command);
+        return new Result(
+            $this,
+            $result_data->getExitCode(),
+            $result_data->getMessage(),
+            $result_data->getData()
+        );
     }
 }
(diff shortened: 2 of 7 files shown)


src/Common/ExecTrait.php
```diff
@@ -0,0 +1,355 @@
...
+     */
+    public function __destruct()
+    {
+        $this->stop();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run()
```
greg-1-anderson: It's a little odd that `Robo\Task\Base\Exec` does not contain its own `run()` implementation. Would it work out to move this method there, and have users of `ExecTrait` just use `execute()`?  Is `$this->getCommand()` really a thing in ExecTrait?  Seems better to use `getCommand()` only in `Robo\Task\Base\Exec`.

src/Common/ProcessExecutor.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Robo\Common;
+
+use Robo\Contract\TaskInterface;
+use Robo\Robo;
+use Robo\Task\BaseTask;
+
+class ProcessExecutor extends BaseTask implements TaskInterface
```
greg-1-anderson: It doesn't seem right for ProcessExecutor to extend BaseTask or implement TaskInterface; since this class is not intended to be used as a Task, it should be satisfy `isa` for either BaseTask or TaskInterface.

Since ProcessExeuctor is not used as a task, it shouldn't need any of the `inflect` methods / interfaces. It would therefore be a lot better if ProcessExecutor simply `use`d the traits from BaseTask that it needs, and forgo extending this class.

src/Common/ExecTrait.php
grasmash: This is here so that users of Process executor can also use run(), mostly just to keep in line with existing terminology.

E.g., it might be odd to use `taskExec()->run()` in one place and `ProcessExecutor()->execute()` in another, but it's really just semantic. I can move if you think it would be better.

grasmash: Ah, there was another reason. We're returning a RoboResult object, which takes `$this` as its first argument. RoboResult's first argument must implement TaskInterface, which in turn requires a `run()` method.

grasmash: Also, yes, getCommand is a thing in ExecTrait. Exec uses defines $this->command as a CommandInterface object and returns a string from getCommand(). ProcessExecutor defines $this->command as a string and its getCommand() just returns that string.

greg-1-anderson: Can you use ResultData instead?  As you noted, Robo\Result holds the result of a task. `ProcessExecutor` should not be a task, so it should not use a Robo\Result.  `Robo\Task\Base\Exec::run()` should call `ExecTrait::execute()` (or equivalent, which should return a ResultData object. `Exec::run()` should then make a new `Result` object, and populate the fields of the constructor from the ResultData. Maybe a new static constructor in `Result` that creates a new Result from a ResultData would be in order here.

I would rather remove the reference to `$task` from Result entirely, but our backwards compatibility goals (to make 1.x not too different from 0.x) and the way Result & ResultPrinter work precluded this. Maybe in 2.x.

It's important to completely detach ProcessExec from TaskInterface; if ProcessExec is a task, then you might as well just use Exec.

grasmash:
> PHP Strict standards:  Robo\Task\Testing\Codecept and Robo\Common\ExecOneCommand define the same property ($command) in the composition of Robo\Task\Testing\Codecept. This might be incompatible, to improve maintainability consider using accessor methods in traits instead. Class was composed in /home/travis/build/consolidation/Robo/src/Task/Testing/Codecept.php on line 263

Odd. So we can't have a command property?

greg-1-anderson:
AFK so can't check, but maybe Robo\Task\Testing\Codecept is using an exec trait that has `$command` too.

greg-1-anderson:
Are you renaming `$command` to `$commandString` to avoid conflicting with two separate instances of `$command`? Couldn't this be unified rather than leaving the duplicate items?

One thing I was thinking about here is that it would be useful if folks could use a [ProcessBuilder](http://api.symfony.com/3.2/Symfony/Component/Process/ProcessBuilder.html) here. In some instances, passing in an entire command string is fine; however, if you are composing a command from arguments, then every user of the API needs to get shell escaping right. This is error-prone. If the ExecTrait stored a Process object instead of a command string, that would give the client the flexibility to choose between composition (with an arguments array, via ProcessBuilder) and simplicity (a single command string, when there are no user-supplied arguments and escaping is not an issue).  Would that over-complicate things?  To start out, the existing API of taskExec could be maintained, and the command string could just be used to generate a Process object immediately before using the ExecTrait to run the command. Just thinking; didn't check the code to see if that works out.

I also haven't had a chance to look at why the Codeception task had a duplicate $command instance. Regardless of how we move forward here, that should be straightened out, not worked around.

grasmash:
>  If the ExecTrait stored a Process object instead of a command string

It does already store a process in `$this->process` but perhaps we want to allow the user to pass a process object? At present, they could still used a process builder and then call `getCommandLine()` on the process to get the command string, and then use `ProcessExecutor`. It may also be confusing is a user created a process, passed it to ProcessExecutor, and then ProcessExecutor overrode some setting, like the time limit.

Anyway, let's imagine what we want the user experience to look like. Let me know if this is what you had in mind.


Builder approach:
```php
<?php
$builder = new ProcessBuilder();
$process = $builder->setArguments(array('--list', '--file=archive.tar.gz'))->getProcess()
$executor = ProcessExecutor::fromProcess($process);
$result = $executor->run();
return $result->wasSuccessful();
```

String approach
```php
<?php
$executor = new ProcessExecutor("echo 'hello world'");
$executor->run();
...

**Comment by greg-1-anderson** [comment]
The goal is to have the equivalent of `->setArguments(array('--list', '--file=archive.tar.gz'))` available in your process exec class. Mostly I was thinking along the lines of the last example, where you always take a Process object, and the client could decide to use a ProcessBuilder if it suited their use-case. You could provide a couple of static factory methods, one that took a Process object, and one that took a command string, and built the Process object for the client.

I agree that it would be confusing if ProcessExecuter took a ProcessBuilder and overrode some fields such as the timeout. Could you simply use the timeout from the ProcessBuilder? I'm just trying to think of ways to simplify and reduce duplicate code; if this suggestion makes things more complicated, then we should skip it.

The other thing I was thinking about is whether it would be easier if you wrote the functions of your code -- the classes that are not commandfiles, that wish to call ProcessExecutor -- as tasks?  See [Extending Robo](http://robo.li/extending/). It would not be necessary to put your custom tasks in a separate project (unless you wanted to share BLT tasks with other Robo-based applications); however, you would have the overhead of creating your own `loadTasks` trait, and you'd have to `use` it in every commandfile where it was needed. However, if you went this route you could just extend BaseTask, and perhaps Robo's existing exec traits would meet your needs.

**Comment by grasmash** [comment]
> The other thing I was thinking about is whether it would be easier if you wrote the functions of your code -- the classes that are not commandfiles, that wish to call ProcessExecutor -- as tasks? 

The issue with using tasks for this is that tasks use the collection builder, which causes issues with accessing the output data.

**Comment by greg-1-anderson** [comment]
Make your wrapper code a task. Have your task `use` the ExecCommand trait. Call it directly, without a builder.

Or are you saying that your wrapper code needs to return a result that is inconvenient to access?

I think it might be helpful to talk about your use case and figure out what the lowest-impedence way to implement it is.

File: src/Common/ExecCommand.php [hunk 3]
```diff
@@ -124,13 +124,24 @@ protected function useCallOnWindows($cmd)
...
     /**
      * @param string $command
      *
      * @return \Robo\Result
      */
     protected function executeCommand($command)
     {
-        return $this->execute($command);
+        $this->process = new Process($command);
+        $result_data = $this->execute();
```
greg-1-anderson: Seems better to have the `execute()` method take a Process as a parameter. Other places would then become `$this->execute($this->process);`, but here you could just pass a local Process object.

grasmash:
```
XML report generated in ../../build/logs/clover.xml
Segmentation fault (core dumped)
 [Testing\Codecept]  Exit code 139  Time 50.059s
```

For 2 of the 4 builds. Ugh.

greg-1-anderson:
Well that's not right. I re-ran the failed test, and it segfaulted again.

Try this:
```
after_success:
  - if [ "$dependencies" = "highest" ]; travis_retry php vendor/bin/coveralls -v
```
The php7 build seems to work. We don't need to analyze coverage for more than one build.

.travis.yml
```diff
@@ -29,7 +29,7 @@ before_script:
 script: "./robo test --coverage"
 
 after_success:
-  - travis_retry php vendor/bin/coveralls -v
+  - if [ "$dependencies" = "highest" ]; travis_retry php vendor/bin/coveralls -v
```
greg-1-anderson: Sorry, there was a typo in my comment. Missing `then` and `fi`

greg-1-anderson: if [ "$dependencies" = "highest" ]; then travis_retry php vendor/bin/coveralls -v; fi;

greg-1-anderson: Last semicolon probably not necessary.

grasmash:
A PHP bug may be the cause, not sure what about my change would have trigged this though:
https://github.com/Codeception/Codeception/issues/3781
https://bugs.php.net/bug.php?id=72286

grasmash:
Super strange. I can't reproduce locally on 5.6. And, the segfault happens after all tests pass!

grasmash:
Ah. My vague intuition paid off.

I had initially removed that __destruct() method because it appeared to completely defeat the purpose of running a task in the background. I'd spawn a background task, the Exec task would complete, and then the background task would be immediately killed.

Anyway, it's still outside of ProcessExecutor, so everyone should be happy.

grasmash:
@greg-1-anderson All tests passing.

src/Task/Base/Exec.php
```diff
@@ -26,7 +27,7 @@
  * ?>
  * ```
  */
-class Exec extends BaseTask implements CommandInterface, PrintedInterface, SimulatedInterface
+class Exec extends BaseTask implements CommandInterface, SimulatedInterface
```
greg-1-anderson: This should still implement PrintedInterface

greg-1-anderson:
Looks good to me -- just one accidental change (looks like PrintedInterface was accidentally removed) that needs to be fixed.

Maybe should make sure that this integrates with the input/output control PR I am going to do.

grasmash:
@greg-1-anderson added back.

greg-1-anderson:
Now that #533 is  in shape (although not yet merged), my final question is whether this PR is entirely orthogonal to that one -- in other words, are you fine manually controlling whether or not the output from your ProcessExecutor is printed, or do you want to have verbosity threshold control here as well?

If the later, then could that facility be duplicated here in just a few lines?

If verbosity threshold is desired here & non-trivial (or if you are tempted to use TaskIO in ProcessExecutor), then I think we should return to the fallback idea of seeing whether one builder per exec is reasonable, or perhaps investigate making it easier to pull out individual results from a builder with multiple taskExec calls.

grasmash:
@greg-1-anderson I would like to incorporate your work from #533 into this. I'll have to investigate in order to estimate the level of effort.

I'd like to avoid using the builder here if possible, since it adds complexity that is not strictly necessary for direct process execution.

greg-1-anderson:
It might be sufficient to simply `use VerbosityThresholdTrait` in the process executor; however, if you did that, you'd need a strategy for injecting the output adapter. That will happen automatically via inflection if you always get the process executor from the container -- but that would be an unfortunate requirement.

If you always have an Output object in any context you will be using the process executor from, you could always instantiate your own OutputAdapter, and inject your Output object into it. You could even make a method of ProcessExecutor to set that up given an Output interface (passed in as a parameter. Using the static `Robo::output()` to get a default value here is a possibility -- although avoiding static methods is best. We're trying to move away from those. 

Still sort of feels like your use-case points towards using taskExec, since it does what you want (logging with verbosity threshold control). Weigh the complexity of the builder against the complexity of wiring up the class references you need. One of the roles and purposes of the builder is doing the needed dependency injection.

greg-1-anderson:
#533 has been merged, so you can rebase here and utilize that, if you still want to go in that direction.

grasmash:
@greg-1-anderson I've closed this in favor of https://github.com/consolidation/Robo/pull/536.

Refactoring the new VerbosityThresholdTrait into ProcessExecutor was trivial. But, I did use a static method to get Robo's output.

greg-1-anderson:
Please see [Scrutinizer's advice about abstract methods in ExecTrait](https://scrutinizer-ci.com/g/consolidation/Robo/inspections/085a9b8c-884f-47d7-be06-45f9d2472a05/issues/files/src/Common/ExecTrait.php?status=new&orderField=path&order=asc&honorSelectedPaths=0)

PR #533, opened 2017-02-01
Add TaskIO::setVerbosityThreshold()

greg-1-anderson:
Allows commands to control the verbosity level at which different tasks will emit their progress notices et. al.

### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Commands can now call `$builder->setRequiredVerbosity(OutputInterface::VERBOSITY_VERBOSE)` (or other verbosity constant) to suppress output of tasks when command executed below a certain verbosity level.

### Description
For Robo scripts that are used from CI servers, it is important to be able to suppress the output of particularly talkative commands (e.g. `composer update`, `git add -A .` et. al.), to avoid being unable to see more interesting messages later, should the CI server truncate the output. This PR gives commands the ability to set a minimum verbosity level on a task-by-task basis, so that the lengthy command output may be easily displayed should the command fail for any reason.

For example, imagine that the `try:builder` command is modified as follows:
```
        return $this->collectionBuilder()
            ->setRequiredVerbosity(OutputInterface::VERBOSITY_VERBOSE)
            ->taskFilesystemStack()
                ->mkdir('a')
                ->touch('a/a.txt')
            ->taskFilesystemStack()
                ->setRequiredVerbosity(OutputInterface::VERBOSITY_VERY_VERBOSE)
                ->mkdir('a/b')
                ->touch('a/b/b.txt')
            ->taskFilesystemStack()
                ->mkdir('a/b/c')
                ->touch('a/b/c/c.txt')
            ->run();
```
Then:
```
$ ./robo -f examples try:builder 
$ ./robo -f examples try:builder -v
 [Filesystem\FilesystemStack] mkdir ["a"]
 [Filesystem\FilesystemStack] touch ["a/a.txt"]
 [Filesystem\FilesystemStack] mkdir ["a/b/c"]
 [Filesystem\FilesystemStack] touch ["a/b/c/c.txt"]
$ ./robo -f examples try:builder -vv
 [Filesystem\FilesystemStack] mkdir ["a"]
 [Filesystem\FilesystemStack] touch ["a/a.txt"]
 [Filesystem\FilesystemStack] mkdir ["a/b"]
 [Filesystem\FilesystemStack] touch ["a/b/b.txt"]
 [Filesystem\FilesystemStack] mkdir ["a/b/c"]
 [Filesystem\FilesystemStack] touch ["a/b/c/c.txt"]
```
Regarding backwards compatibility, everything should still continue to work for existing commands / tasks. However, if a task does **not** implement BaseTask, but **does** use TaskIO, then it must also implement OutputInterface. If it does not, the only consequence is that its output is displayed always, rather than as stipulated by `setRequiredVerbosity()`.

Some time ago, I had an aspirational goal that tasks should not depend on Symfony Console at all, so that hypothetically, a Robo tasklib could be split out and used in non-Console applications. However, I no longer think that is particularly useful to avoid Symfony Console, so this PR takes a simpler implementation and uses OutputInterface et. al. directly.

@grasmash @DavertMik Let me know if you have any feedback on naming the methods, or anything else.

Could still use some tests and documentation.

### DIFF
--- a/src/Collection/CollectionBuilder.php
+++ b/src/Collection/CollectionBuilder.php
@@ -12,6 +12,7 @@ use ReflectionClass;
 use Robo\Task\BaseTask;
 use Robo\Contract\BuilderAwareInterface;
 use Robo\Contract\CommandInterface;
+use Robo\Contract\VerbosityThresholdInterface;
 
 /**
  * Creates a collection, and adds tasks to it.  The collection builder
@@ -377,9 +378,7 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
         if ($task instanceof BuilderAwareInterface) {
             $task->setBuilder($this);
         }
-        // TODO: It would be better if we had a TaskIOInterface
-        // to go with the TaskIO trait. Perhaps in 2.0.
-        if (method_exists($task, 'setVerbosityThreshold')) {
+        if ($task instanceof VerbosityThresholdInterface) {
             $task->setVerbosityThreshold($this->verbosityThreshold());
         }
 
--- a/src/Collection/CollectionBuilder.php
+++ b/src/Collection/CollectionBuilder.php
@@ -236,6 +236,18 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
         return $this;
     }
 
+    public function setRequiredVerbosity($requiredVerbosity)
+    {
+        $currentTask = ($this->currentTask instanceof WrappedTaskInterface) ? $this->currentTask->original() : $this->currentTask;
+        if ($currentTask) {
+            $currentTask->setRequiredVerbosity($requiredVerbosity);
+            return $this;
+        }
+        parent::setRequiredVerbosity($requiredVerbosity);
+        return $this;
+    }
+
+
     /**
      * Return the current task for this collection builder.
      * TODO: Not needed?
@@ -257,6 +269,8 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
         $collectionBuilder = new self($this->commandFile);
         $collectionBuilder->inflect($this);
         $collectionBuilder->simulated($this->isSimulated());
+        $collectionBuilder->setRequiredVerbosity($this->requiredVerbosity());
+
         return $collectionBuilder;
     }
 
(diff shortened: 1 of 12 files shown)


grasmash:
I think that the method name `setRequiredVerbosity()` is a bit confusing, namely the word "required". I'd prefer one of the following:

* setVerbosityLevel()
* setOutputLevel()
* setVerbosityThreshold()

I also considered setMaximumVerbosity() but I think that might be confusing due to the numbers associated with the output level constants. Technically, debug is higher than verbose, etc.

greg-1-anderson:
I like `setVerbosityThreshold()`. Some of the other names are easily confused with `OutputInterface::setVerbosity()`.

greg-1-anderson:
Renamed the functions / variables, and added some simple tests.

The `getPrinted()` mechanism also needs to be hooked into the verbosity threshold; otherwise, the primary purpose of this facility (silencing `composer update` output, et. al.) is missed.

greg-1-anderson:
With this, we are back to including a sanctioned `TaskIO::write()` method. At the moment, only `taskExec` is using it.

src/Collection/CollectionBuilder.php
```diff
@@ -363,6 +377,11 @@ protected function fixTask($task, $args)
         if ($task instanceof BuilderAwareInterface) {
             $task->setBuilder($this);
         }
+        // TODO: It would be better if we had a TaskIOInterface
+        // to go with the TaskIO trait. Perhaps in 2.0.
+        if (method_exists($task, 'setVerbosityThreshold')) {
```
DavertMik: yes, this is a code smell. 
I think you should make OutputVerbosity interface and add it to BaseTask

src/Common/TaskIO.php
```diff
@@ -19,6 +20,24 @@
...
     use ConfigAwareTrait;
+    use OutputAwareTrait;
+
+    protected $verbosityThreshold = 0;
+
+    /**
+     * Required verbocity level before any TaskIO output will be produced.
+     * e.g. OutputInterface::VERBOSITY_VERBOSE
+     */
+    public function setVerbosityThreshold($verbosityThreshold)
```
DavertMik: why not to move it to OutputAwareTrait?

greg-1-anderson:
OK, I'll clean up the missing interface and make a matching trait, and make adapters between Console and TaskIO & inject them at startup.

greg-1-anderson:
@DavertMik OK, Symfony Console separated from TaskIO again, and interfaces created where needed.

PR #536, opened 2017-02-06
Abstracting process execution logic into ExecTrait.php.

grasmash:


src/Common/ProcessExecutor.php
```diff
@@ -0,0 +1,37 @@
...
+    use TaskIO; // uses LoggerAwareTrait and ConfigAwareTrait
+    use ProgressIndicatorAwareTrait;
+    use OutputAwareTrait;
+
+    public function __construct($process)
+    {
+        $this->process = $process;
+        $this->logger = Robo::logger();
+        $this->setOutputAdapter(new OutputAdapter());
+        $this->outputAdapter()->setOutput(Robo::output());
```
grasmash: We are using two static calls in this constructor. This was a very efficient and direct way to achieve the desired behavior. Let me know if there's a preferred alternative.

greg-1-anderson:
Please see [Scrutinizer's advice about abstract methods in ExecTrait](https://scrutinizer-ci.com/g/consolidation/Robo/inspections/085a9b8c-884f-47d7-be06-45f9d2472a05/issues/files/src/Common/ExecTrait.php?status=new&orderField=path&order=asc&honorSelectedPaths=0)

src/Common/ProcessExecutor.php
greg-1-anderson: Regarding the static methods used here, things like `Robo::log()` and `Robo::output()` are provided for the convenience of Robo users who just need access to these objects without doing a lot of work with dependency injection. I prefer they not be used within any of the central parts of the Robo framework. There may be some that still remain, but these are deprecated and to be removed (e.g. the reference to Robo::logger() in the TaskIO trait).

I'd be satisfied to consider `ProcessExecutor` to be similar to these static constructors -- that is, something provided primarily for the convenience of Robo clients, but not itself used by Robo. In this context, the use of the static service provider methods is fine with me, as long as ProcessExecutor itself is not used by Robo internal functions (which should always be tasks that use ExecTrait with proper DI). However, you should definitely allow the logger and the output interface to be passed in as optional parameters to the constructor, and provide `Robo::logger()` and `Robo::output()` if these are omitted.

greg-1-anderson: @DavertMik Do you concur with the strategy described ^^? LMK if you have a different perspective. If we wanted to be really strict, we could make the logger and output interface constructor parameters required, and force users to utilize `Robo::logger()` and `Robo::output()`, if that is what's desired.

Perhaps a middle-ground in terms of strictness would be to make all of the constructor parameters required, and then provide a static factory method (`ProcessExecutor::create()`, or `Robo::process()`) that supplies the logger and output interface via the static service methods. There's some nice symmetry to that.

DavertMik: I agree, it is better way to inject output using DI service. At least we have it, why not to use it for its primary purpose

src/Common/ExecCommand.php
```diff
@@ -220,28 +124,29 @@ protected function useCallOnWindows($cmd)
         return $cmd;
     }
 
+    protected function getCommandDescription()
+    {
+        return $this->getCommand();
+    }
+
+    public function getCommand()
```
greg-1-anderson: Can `getCommand()` simply be removed from `ExecTrait` now?  Make `getCommandDescription()` call `$this->process->getCommandLine()`, but first test to see if `$this->process` has been initialized yet, and return some default "uninitialized" string if not. (Guard against incorrect usage.)

greg-1-anderson:
Looks great -- thanks for working through all of this!

PR #541, opened 2017-02-21
Fix the signature of the addCode() wrapper so that we can name a task.

bwood:
### Overview
This pull request:

- [X] Fixes a bug
- [ ] Adds a feature

### Summary
Make it possible to name a task by providing a second argument to `collectionBuilder::addCode()`.

### Description
See https://github.com/consolidation/Robo/issues/540

```
$ robo  sniff src/Collection/CollectionBuilder.php
 [Exec] Running ./vendor/bin/phpcs --standard=PSR2 -n src/Collection/CollectionBuilder.php
 [Exec] Done in 0.143s
```

### DIFF
--- a/src/Collection/CollectionBuilder.php
+++ b/src/Collection/CollectionBuilder.php
@@ -139,9 +139,9 @@ class CollectionBuilder extends BaseTask implements NestedCollectionInterface, W
         return $this;
     }
 
-    public function addCode(callable $code)
+    public function addCode(callable $code, $name)
     {
-        $this->getCollection()->addCode($code);
+        $this->getCollection()->addCode($code, $name);
         return $this;
     }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This breaks backwards compatibility as submitted by adding a second parameter to addCode(). Please provide a default value for $name, like the other methods that take that parameter.

bwood:
Hopefully this is better.  Let me know if you see anything else.

src/Collection/CollectionBuilder.php
```diff
@@ -139,7 +146,16 @@ public function addTask(TaskInterface $task)
...
+  /**
+   * Add arbitrary code to execute as a task.
+   *
+   * @see \Robo\Collection\CollectionInterface::addCode
+   *
+   * @param callable $code
+   * @param int|string $name
+   * @return $this
+   */
+    public function addCode(callable $code, $name = self::UNNAMEDTASK)
```
greg-1-anderson: This should use the existing `\Robo\Collection\CollectionInterface::UNNAMEDTASK` rather than defining a new constant. Folks generally won't need to ever put this identifier in their code; they'll just leave the variable off if a name is not desired.

bwood: Ah. Makes sense -- wasn't sure if it was kosher to use a FQSEN to that class here.

greg-1-anderson:
Thanks!

PR #542, opened 2017-02-22
Default apigen operation to 'generate'

dehrk:
### Overview
This pull request:

- [x] Fixes a bug
- [x] Adds a feature

### Summary
Allows users to run `apigen generate` more simply

### Description
If a user passes no operation to the constructor, this change will default to generate. Passing an operation (and optionally arguments) will now separate these out into the appropriate class variables. If a user passes an operation to `arg()` or `args()`, it will overwrite the current operation so that only one will be assembled into the final command.

### DIFF
--- a/src/Task/ApiGen/ApiGen.php
+++ b/src/Task/ApiGen/ApiGen.php
@@ -12,7 +12,8 @@ use Traversable;
  * ``` php
  * <?php
  * // ApiGen Command
- * $this->taskApiGen('./apigen.neon')
+ * $this->taskApiGen('./vendor/apigen/apigen.phar')
+ *      ->config('./apigen.neon')
  *      ->templateConfig('vendor/apigen/apigen/templates/bootstrap/config.neon')
  *      ->wipeout(true)
  *       ->run();
@@ -77,7 +78,7 @@ class ApiGen extends BaseTask implements CommandInterface
             }
             return $arg;
         }, $args);
-		$args = array_filter($args);
+        $args = array_filter($args);
         $this->arguments .= ' ' . implode(' ', array_map('static::escape', $args));
         return $this;
     }
--- a/src/Task/ApiGen/ApiGen.php
+++ b/src/Task/ApiGen/ApiGen.php
@@ -71,8 +71,8 @@ class ApiGen extends BaseTask implements CommandInterface
             $args = func_get_args();
         }
         $args = \array_map(function($arg) {
-            if (\preg_match('/^\w+$/', $arg) === 1) {
-                $this->operation = $arg;
+            if (\preg_match('/^\w+$/', trim($arg)) === 1) {
+                $this->operation = " $arg";
                 return;
             }
             return $arg;
--- a/src/Task/ApiGen/ApiGen.php
+++ b/src/Task/ApiGen/ApiGen.php
@@ -30,6 +30,7 @@ class ApiGen extends BaseTask implements CommandInterface
      * @var string
      */
     protected $command;
+	protected $operation = ' generate';
 
     /**
      * @param null|string $pathToApiGen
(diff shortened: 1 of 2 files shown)


src/Task/ApiGen/ApiGen.php
```diff
@@ -30,6 +31,7 @@ class ApiGen extends BaseTask implements CommandInterface
      * @var string
      */
     protected $command;
+    protected $operation = ' generate';
```
greg-1-anderson: Don't include the space in `$operation`; this is harder to read, and inconsistencies can lead to errors. Instead, ensure that appropriate spacing is added when command is composed, and keep the various parts trimmed.

dehrk: I was only mimicking how it's done in `CommandArguments::args()`, but that should be an easy enoguh fix.

src/Task/ApiGen/ApiGen.php
```diff
@@ -39,6 +41,15 @@ class ApiGen extends BaseTask implements CommandInterface
...
     {
         $this->command = $pathToApiGen;
+        $command_parts = [];
+        preg_match('/((?:.+)?apigen(?:\.phar)?)( \w+)? ?(.+)?/', $this->command, $command_parts);
+        if (count($command_parts) === 3) {
+            list($full_command, $this->command, $this->operation) = $command_parts;
+        }
+        if (count($command_parts) === 4) {
+            list($full_command, $this->command, $this->operation, $arg) = $command_parts;
+            $this->arg($arg);
```
greg-1-anderson: There doesn't appear to be an `arg` method. Would be handy, though.

dehrk: `arg` is handled by the `CommandArguments` trait.

src/Task/ApiGen/ApiGen.php
```diff
@@ -48,6 +59,31 @@ public function __construct($pathToApiGen = null)
...
+     *
+     * @return $this
+     */
+    public function args($args)
+    {
+        if (!is_array($args)) {
+            $args = func_get_args();
+        }
+        $args = array_map(function ($arg) {
+            if (preg_match('/^\w+$/', trim($arg)) === 1) {
```
greg-1-anderson: Pulling the operation out of the args seems error-prone to me, or at least inconsistent. Provide another method `operation()` to set the op.

dehrk: That was just done to maintain backwards compatibility, but I saw your comment in https://github.com/consolidation/Robo/issues/177, so I could easily get rid of that.

src/Task/ApiGen/ApiGen.php
greg-1-anderson: Yeah, I was just looking at the rest of the code and realizing that the way I want the code to be, and the way it currently is, is not always the same.

src/Task/ApiGen/ApiGen.php
greg-1-anderson: I see. This is an override. It might be clearer if you called `parent::` at some point, if possible.

Backwards compatibility trumps API design (until we're ready for a new major release), so forget about what I said about an `operation()` method if doing that would introduce ambiguities.

If doing things a certain way to maintain backwards compatibility, please add comments explaining why the code is like this, and what the preferred method is.

src/Task/ApiGen/ApiGen.php
greg-1-anderson: Maintaining consistency is better than meeting aspirational goals; I think in this case, though, it should still be possible to keep the space out of `$operation`.

greg-1-anderson:
Adding a test or two would be cool.

src/Task/ApiGen/ApiGen.php
dehrk: The issue that I saw with calling `parent::` is that the stuff I needed to do falls into the middle of the parent code.

greg-1-anderson:
Ah, APIGen actually already has tests. Handy. Shouldn't be too hard to fix up the false-negatives here.

dehrk:
Sure enough. I just didn't think of that when I submitted the PR.

src/Task/ApiGen/ApiGen.php
dehrk: It also doesn't look like `parent::` works with traits.

greg-1-anderson:
Thanks for this!

PR #545, opened 2017-03-17
Update to Symfony 3.2.6 et. al.

greg-1-anderson:
### Overview
This pull request:

- [X] Updates dependencies
- [ ] Fixes a bug
- [ ] Adds a feature

### Summary
Symfony 3.2.6 deprecates a method we use in testing.


greg-1-anderson:
Keep Symfony at 3.2.4 to avoid bug. Pin phpcs to 2.7.0 to avoid failing code standards on new tests added in later versions that we cannot fix until we release a new major version.

Keeping the Symfony 3.2.6 fix for later, when we do update.

PR #546, opened 2017-03-18
Add factory methods for CollectionBuilder and ProcessExecutor.

greg-1-anderson:
### Overview
This pull request:

- [X] Fixes a bug
- [ ] Adds a feature

### Summary
Remove call to undocumented two-parameter `$container->get('ref', [$arg]);`

### Description
League/Container allows objects to be fetched from the container using a two-parameter 'get' method. For objects that are not shared, this method will construct a new object using the parameters provided. However, this method is undocumented, and PSR-11 does not allow clients of the container to provide parameters for object construction, all of which are presumed to be provided from the container.

### DIFF
--- a/src/Common/ProcessExecutor.php
+++ b/src/Common/ProcessExecutor.php
@@ -24,9 +24,9 @@ class ProcessExecutor implements ConfigAwareInterface, LoggerAwareInterface, Out
         $this->process = $process;
     }
 
-    public static function create($container)
+    public static function create($container, $process)
     {
-        $processExecutor = new self($commandFile);
+        $processExecutor = new self($process);
 
         $processExecutor->setLogger($container->get('logger'));
         $processExecutor->setProgressIndicator($container->get('progressIndicator'));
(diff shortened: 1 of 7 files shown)


src/Common/ProcessExecutor.php
```diff
@@ -15,11 +15,27 @@ class ProcessExecutor implements ConfigAwareInterface, LoggerAwareInterface, Out
...
+     * @return type
+     */
     public function __construct(Process $process)
     {
         $this->process = $process;
     }
 
+    public static function create($container)
+    {
+        $processExecutor = new self($commandFile);
```
grasmash: The constructor for this class takes a Process object, but a command file is being passed here.

greg-1-anderson: Typo -- missing an argument and etc.  I guess there are no tests for ProcessExecutor?

greg-1-anderson:
I fixed the typo, but did not add any tests.

Issue #516, opened 2017-01-13
Add concept of build properties to Robo


grasmash:
From #509:
> Most task runners incorporate the concept of a "build properties" file, which allows a set of key value pairs to be defined in a file and passed to the task runner at run time. One particularly useful feature of Ant and Phing is that they permit property expansion in such files, see build.yml for example. Additionally, there is a convention for overriding such properties via CLI arguments. E.g., robo -Dkey=value. Is there any precedent for using such a file with Robo?

I've created a standalone tool that at least handles the property expansion feature:
https://github.com/grasmash/yaml-expander

Requirements:
- [x] Look for YAML config file in default location and load prior to command execution. E.g., `app/config.yml`. Config should be parsed, expanded, and made available to commands through a service.
- [x] Add ability to define runtime values for build properties. E.g., `robo my:command --key=value`. This would override values loaded from config file.

greg-1-anderson:
+1 on this idea

greg-1-anderson:
Robo already contains a simple Config class.  Terminus [expands on this](https://github.com/pantheon-systems/terminus/tree/master/src/Config) to provide more capabilities.

It would also be good to be able to define configuration values that are injected into:
 a) command options
 b) task setter methods

DavertMik:
Looks good, except that `.env` sholdn't be parsed by `parse_ini_string` :) There are some edge cases

How to can this config providers be used with Robofile? Or you need to extend a Runner for that?

greg-1-anderson:
My idea for RoboFiles is that the config should have some specific section for commands. For example:
```
command:
  foo:
    options:
      bar: baz
```
Then, in your RoboFile:
```
public function foo($options = ['bar' => ''])
{
    $this->io()->writeln("The option is {$options['bar']}");
}
```
The result of `./robo foo` would then be: `The option is baz`.

Similar things could be done with tasks, if you wanted to configure **every** instance of some task's setter (not explicitly overridden by a command) to some default value.

greg-1-anderson:
cc: @weitzman

weitzman:
This looks like a great direction IMO. Drush config would likely switch to the system proposed here.

daften:
Just FYI, we're working on this problem too. The main thing we're struggling with is property overrides through the command-line. Our current solution allows to read properties files throughout the project, see https://github.com/digipolisgent/robo-digipolis-general/blob/develop/src/Common/DigipolisPropertiesAware.php

[...]

PR #552, opened 2017-03-23
Basic scaffolding to inject configuration into the $input object.

greg-1-anderson:
### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Inject configuration into `$input` via a Symfony Event listener. Our desire is that this configuration technique should work even if mixing standard Symfony Console commands with annotated commands -- so we stick to Symfony-provided features to implement this.

See #516


src/Config/ConfigLoader.php
```diff
@@ -0,0 +1,51 @@
...
+        return $this->source;
+    }
+
+    public function setSourceName($source)
+    {
+        $this->source = $source;
+        return $this;
+    }
+
+    public function import($data)
```
grasmash: Add type hint.

src/Config/ConfigLoader.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->config = [];
+        return $this->add($data);
+    }
+
+    public function add($data)
```
grasmash: Add type hint.

src/Config/ConfigLoader.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Robo\Config;
+
+use Grasmash\YamlExpander\Expander;
+
+/**
+ * Load configuration files, and fill in any property values that
+ * need to be expanded.
+ */
+class ConfigLoader implements ConfigLoaderInterface
```
grasmash: In general, I think we should add type hints and param/return docs for public methods.

src/Config/ConfigLoader.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->config = [];
+        return $this->add($data);
+    }
+
+    public function add($data)
+    {
+        if ($data instanceof ConfigLoaderInterface) {
+            $data = $data->export();
+        }
+        $this->config = Expander::expandArrayProperties($data, $this->config);
```
grasmash: This will expand properties using dot notation, but it will not expand dot notated keys. For that, you need something like

```
   $this->config = ArrayManipulator::expandFromDotNotatedKeys($this->config);
   $this->config = Expander::expandArrayProperties($data, $this->config);
   $this->config = ArrayManipulator::expandFromDotNotatedKeys($$this->config);
```

grasmash: I'm actually going to refactor my own implementation and verify this is the case.

grasmash: I can confirm that this PR currently does not expand dot-notated keys when importing key-value pairs like:

```
bin.path = 'some/path'
```

src/Config/ConfigLoader.php
```diff
@@ -0,0 +1,51 @@
...
+
+    public function setSourceName($source)
+    {
+        $this->source = $source;
+        return $this;
+    }
+
+    public function import($data)
+    {
+        $this->config = [];
```
grasmash: This prevents an implementer from calling YamlConfigLoader->load() multiple times, because it wipes out the previous configuration for each newly imported array.

src/Robo.php
```diff
@@ -88,14 +90,36 @@ public static function hasContainer()
...
+    }
+
+    /**
+     * Use a simple config loader to load configuration values from specified paths
+     */
+    public static function loadConfiguration($config, $paths)
+    {
+        $loader = new YamlConfigLoader();
+        foreach ($paths as $path) {
+            $loader->load($path);
```
grasmash: This loader doesn't have access to $config, so it can't use $config to as a reference when expanding properties in the newly loaded keys.

src/Config/ConfigLoader.php
grasmash: Also, this does not add to the config, it overwrites it.

greg-1-anderson: I'm not sure that I want to support expanding dot-notation keys. Instead of `bin.path = '/some/path'`:

```
bin:
  path: /some/path
```
Configuration values from different config files definitely should be merged, not overwritten , though. I will look into that.

Of course, if you are using Robo as a Framework, you can load configuration however you'd like. I would suggest making dot-notation key expansion a function of the yaml loader, not the config object.

src/Config/ConfigLoader.php
greg-1-anderson: The intention is that `import` should be used to give a fresh state, and `add` should be used to merge. My thought, though, was that the loaders should only be responsible for loading, and the config option should be responsible for merging. Maybe the later wasn't the best idea, though: see next comment.

grasmash: That makes sense to me, but the YamlConfigLoader->load() method calls this. Maybe it should be calling add()?

grasmash: Perhaps I should be implementing these classes differently, but I'm currently doing this:

```
$config = new DefaultConfig();
$yaml_config_loader = new \Robo\Config\YamlConfigLoader();
$yaml_config_loader->load($config->get('blt.root') . '/phing/build.yml');
$yaml_config_loader->load($config->get('repo.root') . '/blt/project.yml');
$yaml_config_loader->load($config->get('repo.root') . '/blt/project.local.yml');
$config = new BltConfig($yaml_config_loader->export());
```

This currently has the following problems:
* YamlConfigLoader() doesn't have access to the initial `$config` var, which is required for property expansion.
* Calling YamlConfigLoader()->load() wipes out previously loaded config.

src/Robo.php
greg-1-anderson: I see. I thought that was happening in extend. As you mentioned, I should add more tests that include overridden configuration to ensure this works correctly.

My thought now is that there should be three classes, each with distinct responsibilities:

- Config loader: Responsible only for loading, from yaml files or whatever. Does any necessary transformations (e.g. dot-notation key expansion) to produce a nested array.
- Config merger: Takes any number of nested arrays. Merges them together and then does property expansion.
- Config class: Initialized with a nested array from a loader / merger. Provides simple get/set access for use at runtime.

If you need to load configuration in multiple stages, then you would export a nested array from the config object, use a config merger to combine the nested arrays, and then re-insert the result into the config object. This reduces coupling between the config classes, allowing folks to drop in Symfony/Config for their config loader if they need those features.

grasmash: So you're thinking...

```
$config = new DefaultConfig();
$yaml_config_loader = new \Robo\Config\YamlConfigLoader();
$config_merger = new \Robo\Config\ConfigMerger();

$yaml_config_loader->load($config->get('blt.root') . '/phing/build.yml');
$build_config = $yaml_config_loader->export();
$config = ConfigMerger()->merge($config, $build_config);

$yaml_config_loader->load($config->get('repo.root') . '/blt/project.yml');
$project_config = $yaml_config_loader->export();
$config = ConfigMerger()->merge($config, $build_config);

$yaml_config_loader->load($config->get('repo.root') . '/blt/project.local.yml');
$project_local_config = $yaml_config_loader->export();
$config = ConfigMerger()->merge($config, $build_config);

$config = new BltConfig($config);
```

greg-1-anderson: More like:
```
$config = new DefaultConfig();
$yamlLoader = new \Robo\Config\YamlConfigLoader();
$merger = new \Robo\Config\ConfigMerger($config->export());
$merger->extend($yamlLoader->load($root .  . '/blt/project.yml'));
$config->replace($merger->export());
```

greg-1-anderson:
One small change in behavior here.

Previously, calling `Foo::configure('bar', 'value')` to set some configuration for `taskFoo` would set the Config property `Foo.bar`, which would be interpreted as the literal array key `Foo.bar` (i.e. $config['Foo.bar'] = 'value';`).  Now, this same call will set the Config property `task.Foo.bar`, which is interpreted as a dot-notation index into a nested array (i.e. `$config['task']['Foo']['bar'] = 'value';`)

This should not be a backwards-compatibility break, so long as the Config API is used consistently.

docs/framework.md
```diff
@@ -143,3 +144,22 @@ class MyApplication {
...
+```
+use Robo\Config\Config;
+use Robo\Config\YamlConfigLoader;
+use Robo\Config\ConfigProcessor;
+
+$config = new Config();
+$loader = new YamlConfigLoader();
+$processor = new ConfigProcessor();
+$processor->extend($loader->load('defaults.yml'));
+$processor->extend($loader->load('myconf.yml'));
```
grasmash: I think there may be a missing step here. At this point, $processor->unprocessedConfig will have two rows. How do those get processes and imported into $config?

grasmash: Perhaps line 161 is intended to be `$processor->export()`?

src/Config/ConfigProcessor.php
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Evaluate one configuration item.
+     *
+     * @param array $processed
+     * @param array $config
+     * @return array
+     */
+    protected function evaluate($processed, $config)
+    {
+        return Expander::expandArrayProperties(
```
grasmash: I'm not sure why yet, but this doesn't appear to be expanding things.

grasmash: Nevermind. I had to add my default config using add() first. Will comment in example docs.

docs/framework.md
```diff
@@ -143,3 +144,22 @@ class MyApplication {
...
+Robo provides a very simple configuration loader. If you wish to use more capable loader, you may opt to do so. Replace the call to `Robo::createConfiguration()` with code similar to the following:
+```
+use Robo\Config\Config;
+use Robo\Config\YamlConfigLoader;
+use Robo\Config\ConfigProcessor;
+
+$config = new Config();
+$loader = new YamlConfigLoader();
+$processor = new ConfigProcessor();
+$processor->extend($loader->load('defaults.yml'));
```
grasmash: I would suggest adding:
```
$processor->add($config->export(), 'default');
```

This provides the default configuration as a reference for the subsequently loaded config, which is required for property expansion.

docs/framework.md
greg-1-anderson: Yep, that is correct. The line below has a typo. Thanks.

docs/framework.md
greg-1-anderson: This is only necessary if you subclass `Config` and add configuration in code. If you load your configuration from a `defaults.yml` file, or if you make a `DefaultsLoader` class instead of extending Config, then the code is more uniform.

Uncertain whether to adjust the documentation as suggested. There are certainly a number of reasons why `$config` might already be populated with values; if it is, then adding an export of the configuration to the processor is the right thing to do.

greg-1-anderson: Maybe `$config->extend($processor);` should be allowed. Note that `Config::extend()` creates a processor, and adds its current state to it. I was ambivalent about adding this method, as it creates coupling between the Config object and the config processor. I should either take out `Config::extend()`, and force the loose-coupling shown in the example, or embrace it. n.b. The current code allows repeated calls of `$config->extend($loader->load('defaults.yml'));`. However, in that instance you're stuck with the default processor.

I like decoupled better, but am still undecided here. @DavertMik ?

src/Config/ConfigProcessor.php
grasmash: There's an issue here. This takes _new_ config and expands it's properties using values from the already processed config. That's good.

But, it's possible that the already processed config has unexpanded placeholders for which the new config provides values. In that case, we must re-expand the already processed config. Perhaps this could be done once against the entire data array just prior to export?

grasmash: I resolved this in a child class of ConfigProcessor:
```
  public function export() {
    $processed_config = parent::export();
    $this->processedConfig = Expander::expandArrayProperties($processed_config);

    return $this->processedConfig;
  }
```

But, it seems like a edge case that should be covered in the parent.

grasmash: I'd also argue that dot notated keys should be expanded. It's a fairly widespread convention to represent YAML nesting using dot syntax, and we've already embraced that in Robo. 

I do understand that this is easily addressed in a child class. This works fine:
```
  protected function preprocess($config)
  {
    $config = ArrayManipulator::expandFromDotNotatedKeys($config);

    return $config;
  }
```

docs/framework.md
greg-1-anderson: Config::extend() is to support the use-case of [extending configuration in a RoboFile](https://github.com/consolidation/Robo/pull/552/files#diff-a39c9f7aa728d5fa3b973bc6ba49228aR436). I think that maybe a better place to put this method is in the `Robo` class itself. That seems to be the place where we put references to our statics, and other things we'd rather not do, except to support the simpler use-cases of stand-alone RoboFiles.

greg-1-anderson: Regarding `$processor->add($config->export(), 'default');`, it's fine for BLT to continue to do this for historic reasons, but in the general case, labeling the config export as `'default'` isn't necessarily good advice, because `$config` might contain something other than default values. The "source name" for loaders is to support a Terminus feature where the Config object keeps track of where each configuration item came from. I'm not sure if I should have put those methods into the base implementation or not, but it seemed generally useful to have in the loaders, anyway. Perhaps I should remove it from the Config object, though, and let Terminus' subclasses handle that. In any event, I think I'll add a note to the documentation after the example to put in a `$processor->add($config->export());` if you're extending an existing configuration.

src/Config/ConfigProcessor.php
greg-1-anderson: I agree that it's widespread to represent YAML nesting with dot syntax in code, certainly. I don't know how widespread it is to represent nesting in yaml keys. In any event, it seems superfluous to do so.

In order to provide this feature, we'd need to add another dependency to our composer.json to get the ArrayManipulator. If we did this, we could then remove one method that I copied from ArrayManipulator into Robo core. However, the current implementation of `ArrayManipulator::expandFromDotNotatedKeys()` is incomplete, which could lead to some confusion.

I'm willing to put this in if @DavertMik wants it.

As for re-expanding existing configuration to pick up undefined substitutions that become available in a later configuration file, it hadn't really occurred to me that this would be a requirement. It seems that the config files loaded by an application would be loaded in a specific order, and the earlier config shouldn't have forward-references to config loaded at later stages. Is that not the case?

docs/framework.md
greg-1-anderson: I removed `$config->extend()` and updated the documentation as suggested.

Jelle-S:
@greg-1-anderson (coming from https://github.com/digipolisgent/robo-digipolis-helpers/issues/4#issuecomment-289925010) Looks very promising. As far as I can tell, it meets our use case. We use the abstract robofile in that package to extend from in our deployments for Symfony, Laravel and Drupal 8. So if we load our config in the constructor of that class, we should be good.

One question: Is there a way to use a custom config loader without using Robo as a framework? Or for example to subscribe to an event (using http://symfony.com/doc/current/components/event_dispatcher.html#using-event-subscribers) so we can add config files to load to the config loader?

src/Config/ConfigProcessor.php
grasmash: > In order to provide this feature, we'd need to add another dependency to our composer.json to get the ArrayManipulator

What dependency would you add? I'd expect you to just copy this method too.

src/Runner.php
```diff
@@ -133,7 +133,8 @@ public function run($input = null, $output = null, $app = null, $commandFiles =
 
         // If we were not provided a container, then create one
         if (!$this->getContainer()) {
-            $container = Robo::createDefaultContainer($input, $output, $app);
+            $config = Robo::createConfiguration(['robo.yml']);
```
nerdstein: Will this always reference robo.yml?

greg-1-anderson: As proposed, by default Robo always (and only) loads robo.yml from the cwd. Additional configuration files may be loaded from a RoboFile's constructor. Currently, there is no provision to skip loading robo.yml; however, the RoboFile could choose to re-import the new configuration, erasing whatever was in robo.yml if desired.

greg-1-anderson:
@Jelle-S: Yes, a RoboFile can follow the same steps shown in the configuration section of "Robo as a Framework" if it wishes to do more advanced configuration file handling. This will work just fine; the main reason this is not documented is that there is the question of where the custom configuration loader classes would come from if you did this. If you are calling Robo as a Composer script, then you have an easy answer to this question, and may go ahead and load your own configuration in this way.

It is easy to register an event subscriber in Robo; simply use `@hook command-event`, and the annotated method will be registered as a command event subscriber for you automatically. I don't think this helps particularly vis-a-vis configuration file loading. You should get equivalent results if you use `@hook init` or `@hook command-event` instead of the constructor to load your configuration. The one advantage of using a hook is that your commandfile will be fully constructed at that time, whereas during the constructor, your *AwareInterface dependencies will not have been injected yet.

greg-1-anderson:
@grasmash Please see the last two commits on this PR, and see if this addresses your suggestion regarding cross-file-reference evaluation.

PR #554, opened 2017-04-10
Add $separator parameter to ->option() method.

grasmash:
This permits a user to define the separator and execute commands like `command --option=value`.


greg-1-anderson:
Should work for `optionList` too.

greg-1-anderson:
Thanks

PR #555, opened 2017-04-11
Modify CopyDir task to leverage symfony Filesystem copy.

malikkotob:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Currently, the `CopyDir` task does not allow a user to specify whether or not destination files should be overwritten in the event that they are newer than the source files (a la the `overwrite` option provided in [Phing's copy task](https://www.phing.info/docs/guide/trunk/CopyTask.html). This pull request adds in an optional `$overwrite` parameter to the `CopyDir` task that defaults to `true` to ensure backwards compatibility (could consider changing the default to `false` for a major release).

### Description
To achieve the above behavior, `CopyDir` was modified to use the existing symfony `fs` object's `copy` method instead of what it is currently using - php's built in `copy` method.

### DIFF
--- a/docs/tasks/Filesystem.md
+++ b/docs/tasks/Filesystem.md
@@ -30,6 +30,7 @@ $this->_copyDir('dist/config', 'config');
 
 * `dirPermissions($value)`  Sets the default folder permissions for the destination if it doesn't exist
 * `exclude($exclude = null)`  List files to exclude.
+* `overwrite(false)` Overwrite destination files newer than source files - defaults to true.
 
 ## DeleteDir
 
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
Thanks for this enhancement.

Could you set `$overwrite` with a setter method rather than a constructor parameter?  Backwards compatibility is important for the `CopyDir` class as well as for `taskCopyDir`; switching to a setter method will also take care of this concern.

malikkotob:
Thanks for the quick response @greg-1-anderson. Good point. I went ahead and refactored to use a setter method and removed the attribute from the constructor.

greg-1-anderson:
Perfect -- thanks.

PR #559, opened 2017-04-13
Codecept suite and testname fix

Mitrichius:
### Overview
This pull request:
- Fixes a bug
- Has tests that cover changes

### Summary
Changed suite and testname in codecept task command - now they are handled as simple options.

### Description
**Before fix**: suite and testname always be in the end of the command. This behaviour breaks some rawArg() , e.g. "> log", "| tee log". 
**After fix**: suite and testname are treated like simple options.

### DIFF
--- a/src/Task/Testing/Codecept.php
+++ b/src/Task/Testing/Codecept.php
@@ -28,17 +28,7 @@ use Symfony\Component\Process\Process;
 class Codecept extends BaseTask implements CommandInterface, PrintedInterface
 {
     use \Robo\Common\ExecOneCommand;
-
-    /**
-     * @var string
-     */
-    protected $suite = '';
-
-    /**
-     * @var string
-     */
-    protected $test = '';
-
+    
     /**
      * @var string
      */
@@ -68,7 +58,7 @@ class Codecept extends BaseTask implements CommandInterface, PrintedInterface
      */
     public function suite($suite)
     {
-        $this->suite = $suite;
+        $this->option(null, $suite);
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
`$this->option(null, ...)` is deprecated, as is using redirection via rawArg().

Robo could really use this sort of functionality, though, and currently does not. I am sympathetic to your use-case. @DavertMik might want to accept this as an interim solution, but I am -1 on the implementation.

A better way would be to add a new method similar to `rawArg()` that appended I/O redirection items to the end of the commandline.

DavertMik:
@greg-1-anderson  What is wrong with `option()`?

I like this PR, so I'd rather merge it.

greg-1-anderson:
Using `option` with a key of `null` and relying on a side-effect of `option` to put the argument that is not an argument, but is actually an I/O redirection instruction is deprecated, because there is nothing in the contract of `option` that says this is how you do I/O redirection. It just happens to work that way on accident due to its implementation. If folks use side-effects, then that code will break when the contract is enforced more rigorously.

The right solution would be to add a new method for adding `>` and `|` IO redirection. If these methods called `option(null, ...)` internally, that wouldn't be quite as bad, since it would be easy to fix these uses inside the class if its implementation changed.

I'm still -1 on this implementation for those reasons. It would be better to add the extra methods.

greg-1-anderson:
Clarification: when I said "inside the class" et. al., I meant inside the `\Robo\Common\ExecOneCommand` trait.

greg-1-anderson:
Actually, this PR only moves existing uses of `option()` from one place to another. Since it's not making anything worse, might as well merge it, and clean up my various objections in a later PR.

PR #560, opened 2017-04-14
Setting $this->interactive to NULL by default.

grasmash:
Currently, `detectInteractive()` does not work as intended because its check for `isset($this->interactive)` will always be return true.

### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### DIFF
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -44,7 +44,7 @@ trait ExecTrait
     /**
      * @var boolean
      */
-    protected $interactive = false;
+    protected $interactive = null;
 
     /**
      * @var bool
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Looks good. Hm, I can't change your branch. You can fix the test like this:
```
<?php

class ExecCest
{
    // tests
    public function toExecLsCommand(CliGuy $I)
    {
        $command = strncasecmp(PHP_OS, 'WIN', 3) == 0 ? 'dir' : 'ls';
        $res = $I->taskExec($command)->interactive(false)->run();
        verify($res->getMessage())->contains('src');
        verify($res->getMessage())->contains('codeception.yml');
    }
}
```
n.b. Call to explicitly set `interactive` mode now necessary. There might be a more elegant way to configure the tests to abstract this.

greg-1-anderson:
Thanks!

PR #563, opened 2017-04-19
Update ConfigProcessor.php

grasmash:
Adding type hints for arrays.

### DIFF
--- a/src/Config/ConfigProcessor.php
+++ b/src/Config/ConfigProcessor.php
@@ -108,7 +108,7 @@ class ConfigProcessor
      * @param array $toBeProcessed
      * @return array
      */
-    protected function process($processed, $toBeProcessed)
+    protected function process(array $processed, array $toBeProcessed)
     {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I'm ambivalent -- and usually negative towards -- typehints for `array`. While typehints are valuable and generally valuable, typehinting to `array` prevents users from passing an `ArrayObject` here.

grasmash:
Is ArrayObject otherwise usable here? Can an ArrayObject be used with functions like array_map() and array_reduce()?

greg-1-anderson:
I'd need to do a test to confirm. There are some things that ArrayObject can't be used for. If we're already in that space here, then typehinting would be entirely appropriate.

grasmash:
Yes I suspect we are already in that space.

greg-1-anderson:
Confirmed -- ArrayObject doesn't work with any of the php `array_*` methods, so we might as well typehint wherever that is done (most places).

PR #575, opened 2017-05-09
Fixes #574: In 'stopOnFail' mode in CommandStack with a single command.

### LINKED ISSUE #574: calling stopOnFail(TRUE) prevents result data from being accessed
The value returned by taskExec() changes dramatically depending on whether stopOnFail(TRUE) is used, and whether the executed task fails. This is very confusing:
https://github.com/consolidation/Robo/blob/master/src/Task/CommandStack.php#L107

The result is that output data becomes inaccessible if stopOnFail(TRUE) is used, regardless of whether the task fails.

greg-1-anderson:
Ensures the output from said command is returned in the exec result.

### Overview
This pull request:

- [X] Fixes a bug
- [ ] Adds a feature

### Summary
If a CommandStack has only a single command, always return the result of that command with its output.

### Description
The CommandStack throws away command output when it runs multiple commands individually in `stopOnFail` mode. Commands are run individually in this mode so that the exact command that failed is identified in the result when there is a failure. The results from the command are discarded when doing this because it is inconvenient to merge together multiple Result objects under the current implementation. Result objects print a result message in their constructor, which is a poor time to be printing output; however, the implementation of progress output was easier that way, and there is a focus on keeping things both simple and easy for tasks. Also, command stacks were not intended to be used for collecting command output; the presumption was that a lower-level function (`exec`, Symfony Process component, etc.) would be used for information collection, and CommandStack would be used for running commands that changed the state of the system.

This shortcoming could potentially be solved with additional work; however, in most cases, if the output from a command is desired, it should probably be run individually so that its output is not intermixed with the output of the other commands in the command stack, so this simple fix is probably sufficient.

### DIFF
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -98,7 +98,7 @@ trait ExecTrait
         if (!isset($this->interactive) && function_exists('posix_isatty') && $this->verbosityMeetsThreshold()) {
             $this->interactive = posix_isatty(STDOUT);
         }
-        
+
         return $this;
     }
 
--- a/src/ResultData.php
+++ b/src/ResultData.php
@@ -131,6 +131,25 @@ class ResultData extends \ArrayObject implements ExitCodeInterface, OutputDataIn
     }
 
     /**
+     * Merge another result into this result.  Data already
+     * existing in this result takes precedence over the
+     * data in the Result being merged.
+     *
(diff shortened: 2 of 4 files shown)


greg-1-anderson:
I updated this to do a better job at accumulating results when there are multiple commands in the stack.

greg-1-anderson:
I think this fixes All The Things -- merging.

PR #577, opened 2017-05-12
Remove superfluous printTaskInfo() call

kitgrose:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Removes the explicit call to `printTaskInfo()` inside Docker tasks since it effectively duplicates a similar call made implicitly by `ExecTrait`.

### Description
This seems to resolve consolidation/Robo#576.


greg-1-anderson:
Thanks!

Issue #582, opened 2017-05-23
How does the exclude of subfolders work in copyDir-Task?


thomasgrass:
Hi,
i know that it is possible to exclude folders and files from the taskCopyDir doing it like:

`$this->taskCopyDir(['./' => sprintf('%s', $deployFolderName)])
                ->exclude(['.idea', 'node_modules'])
                ->run();`

But what if i want to exclude subfolders (but not the parent folder)? This command won't work - the subfolders will be copied anyway:

`$this->taskCopyDir(['./'` => sprintf('%s', $deployFolderName)])
                ->exclude(['parent/the_sub_folder'])
                ->run();`

Thomas

greg-1-anderson:
As currently implemented, the `exclude` method of `taskCopyDir` only accepts names of files immediately inside the source directory. This can apply recursively, so `exclude(['foo'])` will exclude `foo` and `bar/foo`.  However, `exclude(['bar/foo])` does not work at all in the current implementation.

A PR would be welcome here, but it would also have to consider how to reconcile the current behavior, where excluding `foo` excludes **any** foo anywhere in the copy tree, with the proposed new behavior to only exclude items named `foo` that are inside a `bar` folder.

alexdesignworks:
@greg-1-anderson 
I'm not sure why this is closed as it is an active issue - specifying the full path to excluded file/dir should work out of the box.

greg-1-anderson:
`taskCopyDir` behaves as documented. A pull request would be welcome, but we do not necessarily hold open feature request issues.

greg-1-anderson:
As an alternative, you might want to try taskRsync, which has more options.

thomasgrass:
Thank you so much for your support! The exclude of subfolders works like a charme!

PR #587, opened 2017-06-02
[#582] Added support for CopyDir to exclude files in subdirs.

alexdesignworks:
### Overview
This pull request:

- [x] Fixes a bug
- [x] Adds a feature
- [x] Updates documentation

### Summary
Allows to exclude files and directories in sub dirs.
Fixes https://github.com/consolidation/Robo/issues/582

### Description
Say, file exists in path `some/deeply/nested/nested_more/file.txt`and the command is called
```
$this->taskCopyDir(['some/deeply' => 'some_destination/deeply'])
```

This PR now allows to specify exclusion in 3 ways:
1. File/directory name basename: `file.txt` for file or `nested_more` for dir

2. File/directory full path from the same location were source is called: `some/deeply/nested/nested_more/file.txt` for file or `some/deeply/nested/nested_more` for dir

3. File/directory path from the same location as source items: `nested/nested_more/file.txt` for file or `nested/nested_more` for dir

### DIFF
--- a/src/Task/Filesystem/CopyDir.php
+++ b/src/Task/Filesystem/CopyDir.php
@@ -100,10 +100,11 @@ class CopyDir extends BaseDir
      *
      * @param string $src Source directory
      * @param string $dst Destination directory
+     * @param string $parent Parent directory
      *
      * @throws \Robo\Exception\TaskException
      */
-    protected function copyDir($src, $dst)
+    protected function copyDir($src, $dst, $parent = '')
     {
         $dir = @opendir($src);
         if (false === $dir) {
@@ -113,14 +114,15 @@ class CopyDir extends BaseDir
             mkdir($dst, $this->chmod, true);
         }
         while (false !== ($file = readdir($dir))) {
-            if (in_array($file, $this->exclude)) {
-                 continue;
+            // Support basename and full path exclusion.
+            if (in_array($file, $this->exclude) || in_array($parent . $file, $this->exclude) || in_array($src . DIRECTORY_SEPARATOR . $file, $this->exclude)) {
+                continue;
             }
             if (($file !== '.') && ($file !== '..')) {
                 $srcFile = $src . '/' . $file;
                 $destFile = $dst . '/' . $file;
                 if (is_dir($srcFile)) {
-                    $this->copyDir($srcFile, $destFile);
+                    $this->copyDir($srcFile, $destFile, $parent . $file . DIRECTORY_SEPARATOR);
                 } else {
                     $this->fs->copy($srcFile, $destFile, $this->overwrite);
                 }
--- a/tests/cli/CopyDirRecursiveExcludeCept.php
+++ b/tests/cli/CopyDirRecursiveExcludeCept.php
@@ -7,6 +7,8 @@ $I->seeDirFound('some/deeply/nested');
 $I->seeDirFound('some/deeply/nested2');
 $I->seeDirFound('some/deeply/nested3');
 $I->seeDirFound('some/deeply/nested3/nested31');
+$I->seeDirFound('some/deeply/nested4');
+$I->seeDirFound('some/deeply/nested4/nested41');
 $I->seeFileFound('structu.re', 'some/deeply/nested');
 $I->seeFileFound('structu1.re', 'some/deeply/nested');
 $I->seeFileFound('structu2.re', 'some/deeply/nested');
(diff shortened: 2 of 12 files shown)


alexdesignworks:
@greg-1-anderson The `master` branch build currently legitimely fails on AppVeyor and local, but the fail is in totally unrelated test that does not use any of the updated code.

src/Task/Filesystem/CopyDir.php
```diff
@@ -113,14 +114,15 @@ protected function copyDir($src, $dst)
             mkdir($dst, $this->chmod, true);
         }
         while (false !== ($file = readdir($dir))) {
-            if (in_array($file, $this->exclude)) {
-                 continue;
+            // Support basename and full path exclusion.
+            if (in_array($file, $this->exclude) || in_array($parent . $file, $this->exclude) || in_array($src . DIRECTORY_SEPARATOR . $file, $this->exclude)) {
```
greg-1-anderson: This seems to disallow exclusions such as `nested3/nested31/structu311.re` -- i.e. only the full relative path, or one directory + current file / dir is checked. Is this intended?

If this was intended, I suppose it is fine, as long as it is documented.

greg-1-anderson:
Thanks for this contribution. Looks pretty good; could use some minor documentation updates.

Regarding the AppVeyor tests, I didn't see a failed test on master, but I'm unsure on that point. In any event, they are passing in #586. I will merge there pretty soon, at which point you can rebase here, which should take care of any problems inherited from master.

src/Task/Filesystem/CopyDir.php
alexdesignworks: Provided that we are talking about the command arguments as they are specified in test, `nested3/nested31/structu311.re` will be excluded because the whole directory `nested3/nested31` is already excluded.
But if you are referring to the case that only one relative directory can be excluded - this will work with more than one directory (test for this case in next commit).

alexdesignworks:
@greg-1-anderson sorry to continuing discussion here, but I tried running tests locally on that branch and `PackExtractCept` test still fails (same as on `master`).

alexdesignworks:
Added minimal documentation for `exclude()` values.

tests/cli/CopyDirRecursiveExcludeCept.php
```diff
@@ -0,0 +1,49 @@
...
+        // File in subdir exclusion.
+        'some/deeply/nested/structu3.re',
+        // Dir exclusion.
+        'nested2',
+        // Subdir exclusion.
+        'some/deeply/nested3/nested31',
+        // Subpath within source exclusion.
+        'nested3/structu31.re',
+        // File in deeper subpath within source exclusion.
+        'nested4/nested41/structu411.re',
```
greg-1-anderson: Yep, this is the case I was talking about. Thanks for adding this to the unit tests to clarify.

greg-1-anderson:
> I tried running tests locally on that branch and PackExtractCept test still fails (same as on master).

I guess you are testing on a Windows machine? That test is passing on AppVeyor. Not sure what else to say; I'm doing local dev on a Mac.

alexdesignworks:
I'm developing on Mac. The test that I was referring before is no longer failing in Appveyor, but now my tests are failing there (but not locally) - strange. I'm not sure what to do with Appveyor fails.

greg-1-anderson:
Robo supports Windows, so it is necessary to get the Appveyor Windows tests working. It would be cool if you also submitted fixes to clear up spurious local fails, if they can be identified, but this is not necessary.

The existing AppVeyor failures say things such as:

`Failed asserting that file "C:\projects\robo\tests\_data\sandbox\some_destination\deeply\nested\structu3.re" does not exist.`

Looks like the path checks for excluding files is not matching correctly on Windows. Maybe force all directory separators (e.g. `\`) to `/` before inserting into the array or comparing would do the trick? If you do the cleanup only at these times, then you shouldn't need to worry about actually using the strings on the OS. Generally speaking, using `/` works most of the time on Windows, as long as the string in question is being used just with the php apis & is not passed to `exec` et. al.

greg-1-anderson:
I fixed the appveyor failures for you so that your contribution could make it into 1.0.8.

greg-1-anderson:
Thanks for providing this.

thomasgrass:
Thanks @alexdesignworks, @greg-1-anderson

PR #594, opened 2017-06-12
Add stdout and stderr to background task ResultData.

grasmash:
Presently, there is no mechanism for getting the output of a taskExec() call that uses background(TRUE). Therefore, if the task fails, the user cannot be informed of the reason for the failure.

This pull request adds the stdout and stderr output from the background task to the task's ResultData. This prevents it from being printed to screen, but makes it accessible via $result->getData().

I'd be open to other approaches, but this seemed simple and backwards compatible.

### DIFF
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -345,7 +345,10 @@ trait ExecTrait
                 $this->getResultData()
             );
         }
-        return new ResultData($this->process->getExitCode());
+        return new ResultData($this->process->getExitCode(), '', [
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Isn't the point of the background task to return to the caller before it has completed execution? I would expect that the full output for the command would not necessarily be available at the point in time when the code in this PR runs.

Could you clarify the sequence of operations in your use case?

grasmash:
Yes you are correct. I've found that this breeds inconsistent behavior. I've refactored my approach to redirect all output to a file that can then be read.

PR #595, opened 2017-06-13
Introduce printTaskNotice method and refactor usages of printTaskInfo…

malikkotob:
… to use it.

### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Addressing the following comment in the comment block from `printTaskInfo`:
```
* n.b. We should probably have printTaskNotice for our ordinary
* output, and use printTaskInfo for less interesting messages.
```

### Description
Introducing a `printTaskNotice` method that functions the way the `printTaskInfo` method always has. `printTaskInfo` is modified to only print when the verbosity level is verbose. All calls to `printTaskInfo` have been replaced with calls to `printTaskNotice` to make the switch seamless. This stemmed from a desire to print different levels of information when commands are output to the console. Chained commands currently look pretty messy, so my hope is to introduce an additional PR to clean up full command output (maybe by printing each individual command to a new line) and use `printTaskInfo` to print it. `printTaskNotice` can maybe print an abbreviated version or something else that is less verbose (suggestions/thoughts are more than welcome).

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -135,9 +135,9 @@ To use the tasks you define in a RoboFile, use its `loadTasks` trait as explaine
 
 To allow tasks access IO, use the `Robo\Common\TaskIO` trait, or inherit your task class from `Robo\Task\BaseTask` (recommended).
 
-Inside tasks you should print process details with `printTaskInfo`, `printTaskSuccess`, and `printTaskError`.
+Inside tasks you should print process details with `printTaskNotice`, `printTaskInfo`, `printTaskSuccess`, and `printTaskError`.
 ```
-$this->printTaskInfo('Processing...');
+$this->printTaskNotice('Processing...');
 ```
 The Task IO methods send all output through a PSR-3 logger. Tasks should use task IO exclusively; methods such as 'say' and 'ask' should reside in the command method. This allows tasks to be usable in any context that has a PSR-3 logger, including background or server processes where it is not possible to directly query the user.
 
--- a/src/Collection/TaskForEach.php
+++ b/src/Collection/TaskForEach.php
@@ -74,7 +74,7 @@ class TaskForEach extends BaseTask implements NestedCollectionInterface, Builder
             $context = ['key' => $key, 'value' => $value];
             $context += $this->context;
             $context += TaskInfo::getTaskContext($this);
-            $this->printTaskInfo($this->message, $context);
+            $this->printTaskNotice($this->message, $context);
         }
     }
 
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -76,6 +76,7 @@ trait ExecTrait
      */
     abstract public function hideTaskProgress();
     abstract public function showTaskProgress($inProgress);
+    abstract public function printTaskNotice($text, $context = null);
     abstract public function printTaskInfo($text, $context = null);
 
     /**
@@ -355,7 +356,7 @@ trait ExecTrait
     {
         if ($this->background && isset($this->process) && $this->process->isRunning()) {
             $this->process->stop();
-            $this->printTaskInfo(
+            $this->printTaskNotice(
                 "Stopped {command}",
                 ['command' => $this->getCommandDescription()]
             );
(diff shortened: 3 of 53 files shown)


greg-1-anderson:
Thanks for taking this on. Since this comment was not addressed for 1.0, this PR will need to wait for Robo 2.0, since changing the conventions for output verbosity is a b/c breaking change.

Also, I still have some doubts about the comment. While it is technically correct, I wonder how many people think of the "Info" in `printTaskInfo` as being a log level. Following the advice of the comment might introduce some confusion.

I'm not sure what the timeframe will be for reconsidering this decision. @DavertMik does not like to advance the major version too frequently; 1.0.0 was tagged 2016-10-13. So far we do not have any big features waiting in the wings that need to break b/c; the timing of the creation of a branch for 2.x (which in practice will probably be done by creating a 1.x branch and repurposing master for 2.x) will probably be contingent on the appearance of such a need.

malikkotob:
Thanks for the response @greg-1-anderson. Do you have any suggestions then, for varying verbosity for command output when using tasks based on the command stack? In BLT's case, we have a custom drush task, that can get pretty messy when the chain of commands is output. To start, I'd like to clean that up and maybe separate each command on its own line for readability (not sure if that's possible with the current setup, but am willing to refactor). Outside of that, we're thinking it would be nice if the verbosity level affected that output, such that verbosity normal did not print the full command chain (which is the current behavior), and maybe verbosity verbose did. Thoughts?

greg-1-anderson:
We have https://github.com/consolidation/Robo/pull/533 that gives some control over task output vs verbosity levels. Does this meet your needs?

malikkotob:
@greg-1-anderson so that's the first part of it. My plan was to basically invoke `printTaskNotice` and pass it minimal information about the command (not sure what exactly), and follow it up with an invocation of `printTaskInfo` and pass it what it is currently being passed (the full command, but cleaned up a bit). Then if the verbosity threshold was set to normal, a user would only see the output of the `printTaskNotice` invocation, but if it was set to verbose, they would see the output of the notice, as well as the info invocation. The way things currently behave it's more of an all or nothing ordeal. As in, if I were to set the verbosity threshold to verbose, the output simply wouldn't display.   

One way I could get around this would be to change calls to `printTaskInfo` inside of `ExecTrait.php` to take in the minimal information (since the log level is `NOTICE`), and then make a call to `printTaskOutput` with `LogLevel::INFO` to pass it a cleaned up version of what is currently being passed. That would prevent a breaking change. Down the road we could refactor if a method is introduced dedicated to calling printTaskOutput with `LogLevel::INFO`. Does that make sense? If so, do you have thoughts on what the more minimal amount of information could be that's printed about the command?

malikkotob:
I'm not sure I follow the suggestion in the first two bullets. As for the latter suggestion, are you essentially saying it would do something like this?

```
    protected function printTaskInfo($text, $context = null)
    {
        if ($this->outputAdapter()->verbosityMeetsThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)) {
            $this->printTaskOutput(LogLevel::INFO, $text, $this->getTaskContext($context));
        }
        else {
            $this->printTaskOutput(LogLevel::NOTICE, $text, $this->getTaskContext($context));
        }
    }
```

If so, does that defeat the purpose of the verbosity threshold feature? In the sense that, we would be invoking `verbosityMeetsThreshold` two times, once here, but then again within `printTaskOutput`. Also, with only the single `printTaskInfo` method to invoke, I'm not sure how we would address printing messages of varying verbosity levels within the context of `printAction`, for example. For the use case I described above, we wouldn't be able to call `printTaskInfo` two times in a row to pass a message that minimally describes the command stack, and then another that spits out the full command (as it currently does). It sounds like what would have to happen for this use case is placing the conditional to check the verbosity level setting within `printAction` itself. If the setting is at least verbose, print a verbose message by calling some new method or `printTaskOutput` directly, else print a more minimal message by calling `printTaskInfo`.

greg-1-anderson:
I was suggesting using the Config object for this selection, not verbosityMeetsThreshold. I agree that would be confusing having two ways to select output verbosity.

malikkotob:
Sorry, I didn't catch that you were talking about the Config object. How does that work? Would Robo house a new setting in config that would allow the verbosity level to be set? (currently, we can set the verbosity threshold, subtle difference)

PR #601, opened 2017-06-29
Add an option to wait an interval of time between parallel processes

gpou:
### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Add an option to parallelExec to wait for a period of time (in seconds) after starting one process and before the next one.

### Description
An example usage would be:
```
$this->taskParallelExec()
    ->process($this->taskCodecept(...))
    ->process($this->taskCodecept(...))
    ->waitInterval(5)
    ->run();
```

### DIFF
--- a/src/Task/Base/ParallelExec.php
+++ b/src/Task/Base/ParallelExec.php
@@ -41,6 +41,11 @@ class ParallelExec extends BaseTask implements CommandInterface, PrintedInterfac
     protected $idleTimeout = null;
 
     /**
+     * @var null|int
+     */
+    protected $waitInterval = null;
+
+    /**
      * @var bool
      */
     protected $isPrinted = false;
@@ -102,6 +107,20 @@ class ParallelExec extends BaseTask implements CommandInterface, PrintedInterfac
     }
 
     /**
+     * Parallel processing will wait `$waitInterval` seconds after launching each process and before
+     * the next one.
+     *
+     * @param int $waitInterval
+     *
+     * @return $this
+     */
+    public function waitInterval($waitInterval)
+    {
+        $this->waitInterval = $waitInterval;
+        return $this;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function getCommand()
@@ -127,6 +146,9 @@ class ParallelExec extends BaseTask implements CommandInterface, PrintedInterfac
             $process->setTimeout($this->timeout);
             $process->start();
             $this->printTaskInfo($process->getCommandLine());
+            if ($this->waitInterval) {
+                sleep($this->waitInterval);
+            }
         }
 
         $this->startProgressIndicator();
--- a/src/Task/Base/ParallelExec.php
+++ b/src/Task/Base/ParallelExec.php
@@ -43,7 +43,7 @@ class ParallelExec extends BaseTask implements CommandInterface, PrintedInterfac
     /**
      * @var null|int
      */
-    protected $waitInterval = null;
+    protected $waitInterval = 0;
 
     /**
      * @var bool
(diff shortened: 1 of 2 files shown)


src/Task/Base/ParallelExec.php
```diff
@@ -127,6 +146,9 @@ public function run()
             $process->setTimeout($this->timeout);
             $process->start();
```
greg-1-anderson: Instead of calling `$process->start()` here, add the process to a queue.

src/Task/Base/ParallelExec.php
```diff
@@ -127,6 +146,9 @@ public function run()
...
             $this->printTaskInfo($process->getCommandLine());
+            if ($this->waitInterval) {
+                sleep($this->waitInterval);
```
greg-1-anderson: Don't sleep here; instead, start the processes inside the `while` loop.

greg-1-anderson: Record the current time at the beginning of the while loop as the 'next start time'. Inside the while loop, if the 'next start time' <= the current time, and the process queue is empty, then pull an item off of the process queue and start it. Add the wait interval to the 'next start time'.

greg-1-anderson: Bonus points: if the queue is empty, sleep for a short period of time, so that we aren't cranking through the `while` loop too hot.

greg-1-anderson: Bonus points pt. 2: encapsulate the enqueue / dequeue / start process into a separate class.

greg-1-anderson:
This is a great enhancement; however, we shouldn't hang up the startup loop with a potentially-lengthy `sleep`. I added some implementation hints as line comments on how this could be done a little smoother.

gpou:
@greg-1-anderson Working on the changes you requested. I have been busy in other things last days, but will come back to this asap as I need those changes :)

greg-1-anderson:
No worries. LMK if you have any questions about my comments. e.g. it's not just `$process->start();`, but everything in the initial `for` loop that should be unrolled and inserted into the `while` loop.

gpou:
@greg-1-anderson Made some changes. Please review again when you have a moment

src/Task/Base/ParallelExec.php
```diff
@@ -122,16 +141,20 @@ public function progressIndicatorSteps()
...
-            $this->printTaskInfo($process->getCommandLine());
-        }
-
         $this->startProgressIndicator();
-        $running = $this->processes;
+        $running = [];
+        $queue = $this->processes;
+        $nextTime = time();
         while (true) {
+            if (($nextTime <= time()) && !empty($queue)) {
```
greg-1-anderson: Maybe `$nextTime >= time()`, if I'm not mistaken.

greg-1-anderson:
Looks great! Perhaps you could add a test w/ `$waitTime` > 0 to prove the logic in the startup loop is correct?

src/Task/Base/ParallelExec.php
greg-1-anderson: Time logic - argh. You have it right here.

gpou:
@greg-1-anderson Ok, I'll add a test. Did you close the PR accidentally? :)

greg-1-anderson:
Yep! Clumsy fingers!

gpou:
@greg-1-anderson  :D Added the test

greg-1-anderson:
Beautiful! Thanks for all of your work on this.

gpou:
@greg-1-anderson Thank you for being so reactive!
Will you publish a new release containing the feature?

greg-1-anderson:
Sure, we've had a string of small releases recently; might as well do one more.

greg-1-anderson:
Tagged 1.1.1.

gpou:
@greg-1-anderson Thanks a lot!

PR #603, opened 2017-07-14
Add noRebuild() method for Codeception run command

maxgorovenko:
### Overview
This pull request:

- [+] Fixes a bug
- [+] Adds a feature
- [-] Breaks backwards compatibility
- [+] Has tests that cover changes

### Summary
This feature is critical for parallel running. Without this, codeception
rebuilds some generated files in each sub-process. This affects to
concurrent writing to the same files the same content. Run fails.

### Description
To this topic there are several bugs in neighbor repos:
First mention at https://github.com/Codeception/Codeception/issues/2054
Then comes here https://github.com/Codeception/robo-paracept/issues/28
And here https://github.com/Codeception/Codeception/pull/4148 but this is not a fix for the problem, just workaround.

My solution adds codeception functionality to Codeception task at Robo project. Using it, we will stop getting errors.

But you should remember, that using flag `--no-rebuild` you should build run `codecept build` manual.

### DIFF
--- a/src/Task/Testing/Codecept.php
+++ b/src/Task/Testing/Codecept.php
@@ -233,6 +233,15 @@ class Codecept extends BaseTask implements CommandInterface, PrintedInterface
     }
 
     /**
+     * @return $this
+     */
+    public function noRebuild()
+    {
+        $this->option("no-rebuild");
+        return $this;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function getCommand()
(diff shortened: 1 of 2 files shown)


ironsmile:
Great! As far as I can tell this would work perfectly. Would give it a try as soon as I can spare some time.

greg-1-anderson:
Looks good -- thanks.

PR #604, opened 2017-07-14
Support simple Composer-based Command plugin mechanism 

greg-1-anderson:
based on the PSR-4 autoload namespace.

### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Simple command plugin mechanism.

### DIFF
--- a/examples/ExampleCommands.php
+++ b/examples/ExampleCommands.php
@@ -1,4 +1,6 @@
 <?php
+namespace Robo\Example\RoboPlugin;
+
 use Robo\Result;
 
 use Consolidation\AnnotatedCommand\CommandData;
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
This implements a plugin scheme based on the idea of @thom8 in https://github.com/acquia/blt/issues/1793.

The examples RoboFile has been repackaged as an example Robo plugin.

greg-1-anderson:
The command plugin commandfile collection code should be factored out of the runner; beyond that, though, this is pretty simple and works well.

greg-1-anderson:
Another alternative would be to use something like Flex, but that would be a lot more heavyweight.

We could also consider using [a "files" autoload entry](https://getcomposer.org/doc/04-schema.md#files) in each plugin, and have each plugin register itself at autoload time. This would require declaring a global function with a static or global variable, or use a singleton to cache the registered classes. The implementation in this PR seems better to me than that, though.

examples/ExampleCommands.php
```diff
@@ -8,19 +10,39 @@
...
+ *             "MyProject\\":"src",
+ *             "MyProject\\RoboPlugin\\":"plugin"
+ *         }
+ *     },
  *
- * To test:
+ * 2. Add command files to the specified directory. The command files must
+ *    have names ending `Commands.php`. They must either be immediately
+ *    inside the directory specified in the psr-4 autoload section, or they
+ *    may be anywhere inside a directory named `Commands`.
```
grasmash: What about hooks?

grasmash:
Testing here: https://github.com/acquia/blt/pull/1821/files

examples/ExampleCommands.php
greg-1-anderson: You can put hooks in any command file. If you prefer, we could also define a default naming convention for files that are intended to have only hooks in them.

Note, however, that there is no difference in behavior between command files and hook files; you could put commands in hook files, or visa-versa.

src/Runner.php
```diff
@@ -178,17 +183,61 @@ public function run($input = null, $output = null, $app = null, $commandFiles =
...
+        $pattern = '#' . $this->namespacePattern . '#';
+        $commandSearchPaths = [];
+        foreach ($this->commandFilePluginPrefixes as $baseNamespace => $paths) {
+            if (preg_match($pattern, $baseNamespace)) {
+                $commandSearchPaths[$baseNamespace] = $paths;
+            }
+        }
+
+        $discovery = new CommandFileDiscovery();
+        $discovery->setSearchLocations(['Commands']);
```
greg-1-anderson: @grasmash Would it make things better for your use cases if the search locations could be altered via a method call that specified the locations? Alternately, we could factor this out into a separate class, and you could call it explicitly with the CommandFileDiscovery object of your choice.

src/Runner.php
```diff
@@ -178,17 +183,61 @@ public function run($input = null, $output = null, $app = null, $commandFiles =
...
+
+    public function setCommandFilePluginPattern($namespacePattern)
+    {
+        $this->namespacePattern = $namespacePattern;
+        return $this;
+    }
+
+    protected function getPluginCommandClasses()
+    {
+        $commandClasses = [];
```
grasmash: Would you mind making this method public? That way, an application that leverages robo can call it and build its own command classes. You may also want to consider making getRoboFileCommands()  public as well.

greg-1-anderson: Yes and no. I don't think it would be good for folks to be calling methods of the runner to do this, but we could provide a plugin discovery class that provided this service. Then Runner could use the new class.

thomscode:
I'm really looking forward to this feature. Is there any further momentum that can be made with this? Looking at the TravisCI report the issue was with a PHP `7.0.7` vs `7.0.8` issue with Symfony `3.3.8`.

greg-1-anderson:
See alternate implementation in #671

Issue #497, opened 2016-11-28
Add command selfupdate - auto update irobo phar like composer can


amenk:
Can we add a selfupdate command just like in composer and magerun?

amenk:
Any progress or thoughts about this?

amenk:
I will make a PR

Issue #234, opened 2015-10-29
Download from Url task


DavertMik:
@gabor-udvari once again, I will ask you to do the thing you proposed )
Looks like Download from Url task would be really useful and can be added to Base namespace. If you can get it implemented (and ImageMinify refactored) that would be great.

PR #615, opened 2017-08-29
#497 Add Basic Self Update Functionality

amenk:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature
- [x] Breaks backwards compatibility (slightly - if anybody had a self:update command in their robo file)

### Summary
Adds self-update functionality like in composer

### Description
... a progress bar would be nice :-)

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -28,6 +28,9 @@ class Application extends SymfonyApplication
             ->addOption(
                 new InputOption('--define', '-D', InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Define a configuration item value.', [])
             );
+
+        $selfUpdateCommand = new SelfUpdateCommand('self:update');
+        $this->add($selfUpdateCommand);
     }
 
     /**
--- a/src/Application.php
+++ b/src/Application.php
@@ -28,8 +28,6 @@ class Application extends SymfonyApplication
             ->addOption(
                 new InputOption('--define', '-D', InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Define a configuration item value.', [])
             );
-
-        $this->addSelfUpdateCommand();
     }
 
     /**
@@ -59,9 +57,17 @@ class Application extends SymfonyApplication
         $this->add($createRoboFile);
     }
 
-    protected function addSelfUpdateCommand()
+    /**
+     * Add self update command, do nothing if null is provided
+     *
+     * @param string $repository GitHub Repository for self update
+     */
+    public function addSelfUpdateCommand($repository = null)
     {
-        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, 'consolidation/robo');
+        if ($repository == null) {
+            return;
+        }
+        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, $repository);
         $this->add($selfUpdateCommand);
     }
 }
--- a/src/Application.php
+++ b/src/Application.php
@@ -59,8 +59,9 @@ class Application extends SymfonyApplication
         $this->add($createRoboFile);
     }
 
-    protected function addSelfUpdateCommand() {
-        $selfUpdateCommand = new SelfUpdateCommand( 'self:update', Robo::VERSION, 'consolidation/robo' );
-        $this->add( $selfUpdateCommand );
+    protected function addSelfUpdateCommand()
+    {
+        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, 'consolidation/robo');
+        $this->add($selfUpdateCommand);
     }
 }
--- a/src/Application.php
+++ b/src/Application.php
@@ -67,7 +67,7 @@ class Application extends SymfonyApplication
         if ($repository == null) {
             return;
         }
-        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, $repository);
+        $selfUpdateCommand = new SelfUpdateCommand('self:update', $this->getName(), $this->getVersion(), $repository);
         $this->add($selfUpdateCommand);
     }
 }
--- a/src/Application.php
+++ b/src/Application.php
@@ -29,8 +29,7 @@ class Application extends SymfonyApplication
                 new InputOption('--define', '-D', InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Define a configuration item value.', [])
             );
 
-        $selfUpdateCommand = new SelfUpdateCommand('self:update');
-        $this->add($selfUpdateCommand);
+        $this->addSelfUpdateCommand();
     }
 
     /**
@@ -58,5 +57,10 @@ class Application extends SymfonyApplication
             $output->writeln("<comment>  Edit this file to add your commands! </comment>");
         });
         $this->add($createRoboFile);
+    }
+
+    protected function addSelfUpdateCommand() {
+        $selfUpdateCommand = new SelfUpdateCommand( 'self:update', Robo::VERSION, 'consolidation/robo' );
+        $this->add( $selfUpdateCommand );
     }
 }
--- a/src/SelfUpdateCommand.php
+++ b/src/SelfUpdateCommand.php
@@ -98,7 +98,7 @@ EOT
 
         if (! is_writable($localFilename)) {
             throw new \Exception(
-                $programName . ' update failed: the "' . $localFilename . '" file could not be written'
+                $programName . ' update failed: the "' . $localFilename . '" file could not be written (execute with sudo)'
             );
         }
 
--- a/src/SelfUpdateCommand.php
+++ b/src/SelfUpdateCommand.php
@@ -29,9 +29,13 @@ class SelfUpdateCommand extends Command
 
     protected $currentVersion;
 
-    public function __construct($name = null, $currentVersion = null, $gitHubRepository = null)
+    protected $applicationName;
+
+    public function __construct($name = null, $applicationName = null, $currentVersion = null, $gitHubRepository = null)
     {
         parent::__construct($name);
+
+        $this->applicationName = $applicationName;
         $this->currentVersion = $currentVersion;
         $this->gitHubRepository = $gitHubRepository;
     }
--- a/src/SelfUpdateCommand.php
+++ b/src/SelfUpdateCommand.php
@@ -48,7 +48,7 @@ EOT
             'http' => [
                 'method' => 'GET',
                 'header' => [
-                    'User-Agent: PHP'
+                    'User-Agent: ' . Robo::APPLICATION_NAME . ' Self-Update (PHP)'
                 ]
             ]
         ];
--- a/src/SelfUpdateCommand.php
+++ b/src/SelfUpdateCommand.php
@@ -29,8 +29,9 @@ class SelfUpdateCommand extends Command
 
     protected $currentVersion;
 
-    public function __construct( $name = null, $currentVersion = null, $gitHubRepository = null) {
-        parent::__construct( $name );
+    public function __construct($name = null, $currentVersion = null, $gitHubRepository = null)
+    {
+        parent::__construct($name);
         $this->currentVersion = $currentVersion;
         $this->gitHubRepository = $gitHubRepository;
     }
@@ -79,7 +80,8 @@ EOT
         return [ $version, $url ];
     }
 
-    public function isEnabled() {
+    public function isEnabled()
+    {
         return !empty(\Phar::running());
     }
 
--- a/src/SelfUpdateCommand.php
+++ b/src/SelfUpdateCommand.php
@@ -79,11 +79,17 @@ EOT
         return [ $version, $url ];
     }
 
+    public function isEnabled() {
+        return !empty(\Phar::running());
+    }
+
+
     /**
      * {@inheritdoc}
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+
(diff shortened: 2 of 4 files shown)


src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,143 @@
...
+            );
+    }
+
+    protected function getLatestReleaseFromGithub($repository)
+    {
+        $opts = [
+            'http' => [
+                'method' => 'GET',
+                'header' => [
+                    'User-Agent: PHP'
```
greg-1-anderson: GitHub requests that the user-agent be filled out for identification purposes. Perhaps this should be a bit more specific.

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,143 @@
...
+        if (! is_writable($localFilename)) {
+            throw new \Exception(
+                $programName . ' update failed: the "' . $localFilename . '" file could not be written'
+            );
+        }
+
+        list( $latest, $downloadUrl ) = $this->getLatestReleaseFromGithub('consolidation/robo');
+
+
+        if (Robo::VERSION == $latest) {
```
greg-1-anderson: This should probably do something rational if running the `dev` version.

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,143 @@
...
+            );
+        }
+
+        if (! is_writable($localFilename)) {
+            throw new \Exception(
+                $programName . ' update failed: the "' . $localFilename . '" file could not be written'
+            );
+        }
+
+        list( $latest, $downloadUrl ) = $this->getLatestReleaseFromGithub('consolidation/robo');
```
greg-1-anderson: Could this be injected into this command class? Ideally, fetch it from a protected method in the Application class, or somehow derive it (e.g. read composer.json). The general idea is, if someone is using [Robo as a Framework](https://robo.li/framework), and they inherit Robo's Application class, then they should also inherit the self:update command.

greg-1-anderson: I should say then they *will* update the self:update command, and it should update their application, not replace their application with Robo.

src/SelfUpdateCommand.php
greg-1-anderson: Also, inject Robo::VERSION.

src/Application.php
```diff
@@ -28,6 +28,9 @@ public function __construct($name, $version)
             ->addOption(
                 new InputOption('--define', '-D', InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Define a configuration item value.', [])
             );
+
+        $selfUpdateCommand = new SelfUpdateCommand('self:update');
+        $this->add($selfUpdateCommand);
```
greg-1-anderson: This doesn't do the right thing unless running as a phar; perhaps phar mode should be detected, and this command should not be added when not running as a phar.

src/SelfUpdateCommand.php
amenk: done

src/SelfUpdateCommand.php
amenk: done

src/SelfUpdateCommand.php
amenk: done

amenk:
Thanks for your review! All fixed ... I hope :-)

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,162 @@
...
+            );
+    }
+
+    protected function getLatestReleaseFromGithub()
+    {
+        $opts = [
+            'http' => [
+                'method' => 'GET',
+                'header' => [
+                    'User-Agent: ' . Robo::APPLICATION_NAME . ' Self-Update (PHP)'
```
greg-1-anderson: Not accurate for derived applications. Maybe use $this->gitHubRepository here?

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,162 @@
...
+
+
+        if ($this->currentVersion == $latest) {
+            $output->writeln('No update available');
+            return;
+        }
+
+        $fs = new sfFilesystem();
+
+        $output->writeln('Downloading ' . Robo::APPLICATION_NAME . ' ' . $latest);
```
greg-1-anderson: Again, maybe $this->gitHubRepository is better here.

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,162 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
```
greg-1-anderson: Was the self:update command copied from Symfony? If so, please rewrite this copyright notice to make it clear that this is a derived work. Explicitly name the Symfony license, and identify the original copyright holder (Fabien Potencier) and the URL where the original code was obtained.

If this command merely used a Symfony command as boilerplate (e.g. from Symfony documentation), then this notice could probably just be removed.

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,162 @@
...
+
+        $version = $releases[0]->tag_name;
+        $url     = $releases[0]->assets[0]->browser_download_url;
+
+        return [ $version, $url ];
+    }
+
+    public function isEnabled()
+    {
+        return !empty(\Phar::running());
```
greg-1-anderson: I don't have time to test this right now, but I presume you did? At one point in the past, `Phar::running()` was returning false regardless of whether Robo was running as a Phar or not.

src/Application.php
```diff
@@ -56,4 +58,10 @@ public function addInitRoboFileCommand($roboFile, $roboClass)
         });
         $this->add($createRoboFile);
     }
+
+    protected function addSelfUpdateCommand()
+    {
+        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, 'consolidation/robo');
```
greg-1-anderson: Could Robo::VERSION be obtained from the Application instance?

Could 'consolidation/robo' be read from the composer.json file?

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,162 @@
...
+        if (! is_writable($localFilename)) {
+            throw new \Exception(
+                $programName . ' update failed: the "' . $localFilename . '" file could not be written (execute with sudo)'
+            );
+        }
+
+        list( $latest, $downloadUrl ) = $this->getLatestReleaseFromGithub();
+
+
+        if ($this->currentVersion == $latest) {
```
greg-1-anderson: Did you decide that handling dev versions was unnecessary since this only works for phars? That's probably sufficient. An explanatory comment here would be helpful, though.

src/Application.php
greg-1-anderson: The 'Robo as a Framework' document should be updated to describe what a derivative app needs to do to get the self:update functionality. If the above enhancements work, though, then it seems that derived apps don't need to do anything, in which case no documentation is necessary (although a note there that this is a feature would be great).

greg-1-anderson:
Looking really good -- just a few more notes. Thanks for your work on this.

src/SelfUpdateCommand.php
amenk: I feel that is okay like it is - gitHub would see the requested repo anyways :)

src/SelfUpdateCommand.php
amenk: it is taken from magerun partially, but rewritten in the biggest parts, so I will rather remove that.

src/Application.php
amenk: Github Repo is not necessarily identical to the composer name .. so we would have to go to packagist and then fetch he github repo ... too much ...

src/SelfUpdateCommand.php
amenk: Yes, I thought it's fine because of  the non-phar-disabling.

src/Application.php
greg-1-anderson: We could document the requirement that the github repo be the same as the Composer package name to use this mechanism, and also explain how to override this method for apps where this is not true.

amenk: I don't like it ... could lead to "false positives". Github repo and composer.json are two very different things, just by convention often the same. What we could / should do is to disable the update from your main repo for apps. How can I detect it's an app?

src/SelfUpdateCommand.php
amenk: I tested it

src/Application.php
greg-1-anderson: Well, you could read composer.json and check to see if the repository is consolidation/robo.

The better thing, I think, would be if it was possible to inject the version and repository into the Application class, and pass the correct values in from the front controller. The Application class is already provided `$version` in its constructor (you may have to cache this separately, because Symfony prefers to protect information like this in private fields); you could add the packagist project [here](https://github.com/consolidation/Robo/blob/1.1.2/robo#L19) via setter injection. Getting this into the application (which I would also do via setter injection) might be a bit inconvenient, as the Robo runtime class initialization is a bit messy (b/c support, mostly). However, the advantage here is that anyone who uses Robo as a framework must provide their own front controller, and this would be a natural place for them to provide their update project also.

amenk: fixed

src/SelfUpdateCommand.php
amenk: added

src/Application.php
```diff
@@ -59,9 +57,17 @@ public function addInitRoboFileCommand($roboFile, $roboClass)
...
+     *
+     * @param string $repository GitHub Repository for self update
+     */
+    public function addSelfUpdateCommand($repository = null)
     {
-        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, 'consolidation/robo');
+        if ($repository == null) {
+            return;
+        }
+        $selfUpdateCommand = new SelfUpdateCommand('self:update', Robo::VERSION, $repository);
```
greg-1-anderson: Use `$version` from the constructor here. (You'll probably need to cache it in a class field.) Also inject the app name from the constructor here.

src/SelfUpdateCommand.php
```diff
@@ -60,7 +60,7 @@ protected function getLatestReleaseFromGithub()
             'http' => [
                 'method' => 'GET',
                 'header' => [
-                    'User-Agent: ' . Robo::APPLICATION_NAME . ' Self-Update (PHP)'
+                    'User-Agent: ' . Robo::APPLICATION_NAME  . ' (' . $this->gitHubRepository . ')' . ' Self-Update (PHP)'
```
greg-1-anderson: Use the app name injected here instead of a constant. Ditto line 123.

greg-1-anderson:
Looking really good -- left just a couple of comments on places where re-usabilty needs to be enhanced a bit.

amenk:
done.

two more things which would be nice to have:

* add support for --simulate (I currently do not know how to fetch that option)
* Progress bar during the download (also no clue how to do this)

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,166 @@
...
+            );
+    }
+
+    protected function getLatestReleaseFromGithub()
+    {
+        $opts = [
+            'http' => [
+                'method' => 'GET',
+                'header' => [
+                    'User-Agent: ' . Robo::APPLICATION_NAME  . ' (' . $this->gitHubRepository . ')' . ' Self-Update (PHP)'
```
greg-1-anderson: This can use the application name variable now.

src/SelfUpdateCommand.php
greg-1-anderson: Don't forget to update the header

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,166 @@
...
+
+
+        if ($this->currentVersion == $latest) {
+            $output->writeln('No update available');
+            return;
+        }
+
+        $fs = new sfFilesystem();
+
+        $output->writeln('Downloading ' . Robo::APPLICATION_NAME . ' (' . $this->gitHubRepository . ') ' . $latest);
```
greg-1-anderson: This can use the application name field now.

src/SelfUpdateCommand.php
```diff
@@ -0,0 +1,166 @@
...
+{
+    private $command;
+
+    protected $gitHubRepository;
+
+    protected $currentVersion;
+
+    protected $applicationName;
+
+    public function __construct($name = null, $applicationName = null, $currentVersion = null, $gitHubRepository = null)
```
greg-1-anderson: `$name` and `$applicaionName` are the same thing, aren't they? Did I miss the place where you changed the [instantiation of the Application instance](https://github.com/consolidation/Robo/blob/1.1.2/src/Robo.php#L270), or did you miss adding a file in your last commit?

greg-1-anderson:
If you implemented the body of the self update command using Robo tasks, then `--simulate` would be automatic. I think this would be preferable to hard-coding simulate & progress bar code just for self:update. This can be a follow-on task, though, if you prefer.

I thought that we had an issue requesting a download from URL task with progress bar. I only found #234.

src/SelfUpdateCommand.php
greg-1-anderson: OIC `$name` is `self:update`.

greg-1-anderson:
Made a few minor updates and merged to master. Thank you for contributing this.

Another follow-on task that would be very helpful, if you felt so inclined, would be to add an automatic update nag message to periodically inform users that a new version is available. [Terminus](https://github.com/pantheon-systems/terminus) already has code to do this. It is important to omit this message in --no-interaction mode (to avoid adding this output in scripts -- requiring that the `CI` environment variable not be defined is also helpful), and it is also important to throttle the calls to the GitHub API, and not check too often, to avoid exceeding the GitHub API rate limit.

PR #622, opened 2017-09-12
Add failGroup() method for Codeception run command

maxgorovenko:
### Overview
This pull request:

- [-] Fixes a bug
- [+] Adds a feature
- [-] Breaks backwards compatibility
- [+] Has tests that cover changes

### Summary
Add failGroup() method for Codeception run command

### Description
This feature is major for parallel running. Without this, codeception
puts failed tasks to the same file, each thread rewrites file. With
this config we can send fails to different files and aggregate them
later manually.

### DIFF
--- a/src/Task/Testing/Codecept.php
+++ b/src/Task/Testing/Codecept.php
@@ -247,7 +247,7 @@ class Codecept extends BaseTask implements CommandInterface, PrintedInterface
      */
     public function failGroup($failGroup)
     {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I usually prefer to see test assertions contain only string constants, not variable references, unless absolutely necessary. Your use of variables here is trivial, though, so I won't ask you to take time to change it.

Thanks for contributing this, and thanks for including tests.

PR #624, opened 2017-09-19
Add overwrite argument to Filesystem\loadShortcuts.php::_rename

asgorobets:
### Overview
This pull request:

- [X] Fixes a bug
- [ ] Adds a feature

### Summary
I've noticed when using _rename shortcut from FS tasks, the overwrite parameter is silently ignored, even thought it's present in the task itself. Adding it as it doesn't break backwards compatibility and is potentially useful to shortcut users.


greg-1-anderson:
LGTM, Thanks

PR #625, opened 2017-09-19
Fix Robo\Task\Docker\Base to implement CommandInterface

asgorobets:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Fix Robo\Task\Docker\Base to implement CommandInterface as it is already an implementation of the above. With the fix you can use Docker commands in SshExec tasks and other tasks expecting a CommandInterface or a string command let's say if you plan to proxy a docker command execution to a remote machine


greg-1-anderson:
Thanks for finding this omission.

PR #631, opened 2017-10-09
Make Changelog task more flexible.

grasmash:
This pull request:

* Adds `setBody()` to permit users to set raw text directly and bypass the default text processing of the log.
* Adds `setHeader()` likewise.
* Splits some logic into separate methods to make it easier to override in child classes.

### DIFF
--- a/docs/tasks/Development.md
+++ b/docs/tasks/Development.md
@@ -5,33 +5,46 @@
 Helps to manage changelog file.
 Creates or updates `changelog.md` file with recent changes in current version.
 
+Specify individual changes and allow Robo to format the changelog entry:
 ``` php
 <?php
 $version = "0.1.0";
-$this->taskChangelog()
+$this->taskChangelog('CHANGELOG.md')
  ->version($version)
- ->change("released to github")
+ ->change("Added .gitignore file.")
+ ->change("Removed tmp directory.")
  ->run();
 ?>
 ```
 
-Changes can be asked from Console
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Could you explain the use-case a little better? If calling both setBody and setHeader, then the changelog task only inserts into the changelog file, and collects no log information?

Some additional class documentation explaining the different supported ways to use this class w/ this PR would be helpful.

grasmash:
> the changelog task only inserts into the changelog file, and collects no log information?

That's correct. If using the both setBody and setHeader methods, the value of the change log task is in creating a file (if it doesn't exist) and appending or updating the relevant portion of that file.

greg-1-anderson:
That sounds good. Could you add some documentation (top of the Changelog task sources would be fine) that shows the two ways of using taskChangelog (explicit header and body or collected log messages)?

grasmash:
Done.

A separate matter:  it appears that the change log task requires the filename as a constructor parameter. This would seem to deprecate the filename() method, correct?

greg-1-anderson:
I'd prefer to resolve that discrepancy by making the constructor parameter optional, so that the setter method may be used if desired. I don't know if this task would ever be used with chained state or the producer / consumer pattern, but it's best to be flexible on that count and avoid required constructor injection for tasks.

greg-1-anderson:
Thanks for this. Further improvements may be done in separate PRs if desired.

PR #632, opened 2017-10-09
Add openBrowser() to GitHubRelease task.

grasmash:


greg-1-anderson:
I'd rather not put cross-task knowledge in GitHub release of the open browser task. Tasks should avoid building other tasks in general.

What I would prefer here would be to ensure that the github release task makes the release URL available to the task collection state, so that one could later call taskOpenBrowser on this result.

grasmash:
That is already possible with:

```
    $data = $result->getData();
    $response = $data['response'];
    $this->taskOpenBrowser($response->html_url)->run();
```

greg-1-anderson:
Use chained state to copy the result from one task into a setter method of a follow-on task:
```
$collection->
  taskGitHubRelease()->
    ...
    ->storeState('release-url', 'response')
  taskOpenBrowser()
    ->deferTaskConfiguration('setUrl', 'release-url')
  ->run();
```
The `storeState` command above copies $data['response'] into the stored state item 'release-url'. The `deferTaskConfiguration` calls `setUrl` just before the task is executed, and provides the stored state from the previous command. In order for this to work, taskOpenBrowser would need a `setUrl` setter injection method, because stored state does not work with constructor injection.

To make this even easier, `taskGitHubRelease` could be declared to be a "url" producer, and `taskOpernBrowser` could be delcared to be a url consumer.  storeState / deferTaskConfiguration might be a bit complex for folks, but using a producer / consumer relationship would allow:
```
$collection->
  taskGitHubRelease()->
    ...
  taskOpenBrowser()
  ->run();
```
Then, taskOpenBrowser would open the URL provided by the most-recently-executed URL producer.

PR #646, opened 2017-11-07
Remove deprecated functionality from documentation

thomscode:
### Overview
This pull request:

- [ ] Fixes a bug
- [ ] Adds a feature

### Summary
Removes documentation for deprecated configuration functionality

### Description
Newer functionality is already documented

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -438,20 +438,6 @@ In a RoboFile, use `Robo::Config()->get('task.TaskGroup.MyOperation.settings.dir
 
 In the implementation of `taskMyOperation()` itself, it is in general not necessary to access configuration values directly, as it is preferable to allow Robo to inject configuration as described above. However, if desired, configuration may be accessed from within the method of any task that extends `\Robo\Task\BaseTask` (or otherwise uses `ConfigAwareTrait`) may do so via `static::getConfigValue('key', 'default');`.
 
-### Providing Default Configuration in Code
-
-RoboFiles that wish to provide default configuration values that can be overridden via robo.yml values or commandline options may do so in the class' constructor method.  The example below demonstrates how to set up a default value for the `task.Ssh.remoteDir` configuration property in code:
-```
-class RoboFile
-{
-    public function __construct()
-    {    
-        Robo\Task\Remote\Ssh::configure('remoteDir', '/srv/www');
-    }
-}
-```
-If `task.Remote.Ssh.remoteDir` is set to some other value in the robo.yml configuration file in the current directory, then the value from the configuration file will take precedence.
-
 ### Loading Configuration From Another Source
 
 Sometimes, a RoboFile might want to define its own private configuration file to use in addition to the standard `robo.yml` file. This can also be done in the constructor.


greg-1-anderson:
This bit of documentation is not deprecated. This is a good way for a RoboFile to provide default values for configuration that may be overridden via a robo.yml file.

greg-1-anderson:
Unless this code is broken now.

thomscode:
This doesn't work properly without [consolidation/config #8](https://github.com/consolidation/config/pull/8), which you said was the old/deprecated method.

greg-1-anderson:
By "doesn't work properly" do you mean that it isn't possible to use anything other than simple literals for default values (no nested arrays)?

Seems the choices are:

1. Rewrite Robo\Config to be a config overlay.
2. Revisit https://github.com/consolidation/config/pull/8, and re-implement with `Dflydev\DotAccessData\Data`
3. Document the limitation and include an example of how to use a config overlay

Unless we do `1.`, using a config overlay is only available to apps that use Robo as a Framework, as RoboFiles must use whatever config object Robo provides. I'm not sure whether switching Robo config to a config overlay would be backwards compatible or not. Both classes implement ConfigInterface, so in theory that would work out.

In the short term, I think I'd prefer to document the limitation rather than entirely remove this documentation section.

thomscode:
When adding a default value it's not added as a PHP array, but as a dot-separated string defining the same path used in `robo.yml` to override the default.

The current implementations of `Consolidation\Config::hasDefault` and `Consolidation\Config::getDefault`  only check for and retrieve an existing key in the `defaults` array and do not check for partial matches within the dot-separated string key.

My current implementation solves this problem by searching for partial matches within the dot-separated string and returns a PHP array containing the appropriate level of information based on the specific key requested.

My implementation is pretty innefficient, but does solve the problem. I think implementing `Dflydev\DotAccessData\Data` is the easiest solution and the best to not break backwards compatibility.

greg-1-anderson:
OK. I could support https://github.com/consolidation/config/pull/8 if it used `Dflydev\DotAccessData\Data`

thomscode:
I left out the key problem as to why this doesn't work. In the documentation example:
```php
Robo\Task\Remote\Ssh::configure('remoteDir', '/srv/www');
```
This value would be stored in the `defaults` array like:
```php
$defaults = array(
    "task.Remote.Ssh.settings.remoteDir" => "/srv/www"
);
```
When Robo checks to see if any default values are stored for `Robo\Task\Remote\Ssh`, it looks for the `task.Remote.Ssh.settings` and expects to receive an array. Since the key stored in defaults is `task.Remote.Ssh.settings.remoteDir` not `task.Remote.Ssh.settings`, nothing is returned and no default is loaded.

As stated above, the pull request does fix this, but implementing `Dflydev\DotAccessData\Data` would be more efficient.

Unless you would prefer option 1 (listed above) I'll re-open the pull request and re-work this to implement `Dflydev\DotAccessData\Data`.

thomscode:
I'll close this PR and rework the [consolidation/config #8](https://github.com/consolidation/config/pull/8)

Issue #670, opened 2018-01-19
Lay the groundwork for Robo plugins


ademarco:
## Motivation

Currently the only way to customize Robo behavior is to built your own standalone application as described
in [Robo as a Framework](http://robo.li/framework/).

This issue wants to explore possible ways for third-parties to manipulate the service container in order to allow advanced customization like: alternative command class discovery methods, custom plugin systems, and so on.

## Possible solution

A possible solution would be to fire events in `\Robo\Robo::configureContainer()` that will allow event subscribers to manipulate the container.

Additionally we should provide a way for subscribers to be discovered at run-time, so that third-parties would be able to manipulate the container by, for example, placing the subscriber class under a predefined namespace (e.g. discovery by naming-convention).

Parts of the solution outlined above can be found in the [European Commission's Task Runner](https://github.com/ec-europa/oe-task-runner) which is an example of "Robo as a framework" project.

The Task Runner above allows for [command classes to be auto-discovered](https://github.com/ec-europa/oe-task-runner#expose-custom-commands-as-php-classes) and commands to be [dynamically defined via YAML](https://github.com/ec-europa/oe-task-runner#expose-custom-commands-as-yaml-configuration) and it can achieve that by just taking control over the Robo Runner container.

The outlined solution would essentially remove the need for a custom application (e.g. "Robo as a Framework").

## Outline

The method described above could be implemented in the following way:

#### Step 1: Define an event for container manipulation

We would ideally create an event for container manipulation, like `\Robo\Events\AlterContainer` or similar, which will include the actual container for manipulation.

Additional events that will fire in other sensitive parts of the Robo Runner can also be created.

#### Step 2: Auto-discovery of event subscribers

In order to auto-discover event subscriber classes we should pass the Composer class loader to the runner, something like:

```php
<?php
...
    } else {
      $classLoader = require __DIR__.'/../vendor/autoload.php';
    }
}

$runner = new \Robo\Runner();
$runner->setSelfUpdateRepository('consolidation/robo');
$runner->setClassLoader($classLoader);
$statusCode = $runner->execute($_SERVER['argv']);
exit($statusCode);
```

Then, after that, we would collect event subscribers based on some sort of naming convention, say we will load all classes in PSR-4 that use the partial namespace `..\Robo\EventSubscribers\..`, something like: 

```php
<?php

    public function registerEventSubscribers(Container $container, ClassLoader $classLoader)
    {
[...]

PR #671, opened 2018-01-21
Lay the groundwork for Robo plugins #670

ademarco:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
PR related to #670, current plan is:

- [x] Add class discovery service + tests
- [x] Add command discovery to Runner:run()
- [x] ~Add event subscribers discovery~ This will be discussed in a follow-up issue.


ademarco:
Still very work in progress, for the moment I provide a `ClassDiscovery` service which can be used to discover classes regardless of their purpose then a Command Collector service can use the class discovery service to collect command classes and an Event Subscriber Collector to collect event subscribers.

I'll now work on the two services above.

ademarco:
I've updated the PR description with a development plan.

src/Runner.php
```diff
@@ -169,6 +181,12 @@ public function run($input = null, $output = null, $app = null, $commandFiles =
                 $commandFiles = [];
             }
         }
+
+        /** @var \Robo\ClassDiscovery\RelativeNamespaceDiscovery $discovery */
+        $discovery = Robo::service('relativeNamespaceDiscovery');
+        $discovery->setRelativeNamespace($this->commandFileRelativeNamespace);
+        $commandFiles = array_merge((array)$commandFiles, $discovery->getClasses());
+
```
ademarco: So this is how we integrate command discovery at the moment, we can/should turn this into a separate service.

src/Robo.php
```diff
@@ -298,6 +306,8 @@ public static function addInflectors($container)
             ->invokeMethod('setHookManager', ['hookManager']);
         $container->inflector(\Robo\Contract\VerbosityThresholdInterface::class)
             ->invokeMethod('setOutputAdapter', ['outputAdapter']);
+        $container->inflector(\Robo\Contract\ClassLoaderAwareInterface::class)
+            ->invokeMethod('setClassLoader', ['classLoader']);
```
greg-1-anderson: Do we really need to inflect this? Inflection allows any CommandFile (et. al.) to have access to the class loader. If we only need global (built-in) classes to do this, typically, then we should probably just inject this reference during container initialization.

src/ClassDiscovery/RelativeNamespaceDiscovery.php
```diff
@@ -0,0 +1,100 @@
...
+
+        foreach ($this->getClassLoader()->getPrefixesPsr4() as $baseNamespace => $directories) {
+            $directories = array_filter(array_map(function ($directory) use ($relativePath) {
+                return $directory.$relativePath;
+            }, $directories), 'is_dir');
+
+            if ($directories) {
+                foreach ($this->search($directories, $this->searchPattern) as $file) {
+                    $relativePathName = $file->getRelativePathname();
+                    $classes[] = $baseNamespace.$this->convertPathToNamespace($relativePath.DIRECTORY_SEPARATOR.$relativePathName);
```
greg-1-anderson: This looks about right on inspection, but the Windows Appveyor tests are failing. You could potentially just use "/" here, I think, as DIRECTORY_SEPARATOR is not necessary in instances where paths are only being used by php apis.

greg-1-anderson: (Any change here will likely require changes in convertPathToNamespace & maybe elsewhere of course. I'm okay with any solution that passes on Appveyor.)

robo
```diff
@@ -1,22 +1,23 @@
...
-        require_once 'phar://robo.phar/vendor/autoload.php';
+        $autoloaderPath = __DIR__ . '/../../autoload.php';
     }
 }
+$classLoader = require $autoloaderPath;
 $runner = new \Robo\Runner();
-$runner->setSelfUpdateRepository('consolidation/robo');
+$runner
+  ->setSelfUpdateRepository('consolidation/robo')
+  ->setClassLoader($classLoader);
```
greg-1-anderson: We should inject the plugin discovery here, to make it really obvious for Robo-as-a-Framework folks how to customize this.

src/Runner.php
```diff
@@ -43,6 +44,16 @@ class Runner implements ContainerAwareInterface
...
 
+    /**
+     * @var \Composer\Autoload\ClassLoader
+     */
+    protected $classLoader = null;
+
+    /**
+     * @var string
+     */
+    protected $commandFileRelativeNamespace = 'Robo\Commands';
```
greg-1-anderson: Remove the auto-initialized constant, and inject it from the Robo script (so folks searching for the default value will find it in only one place).

Perhaps the default should be something like `Robo\Plugin`, so that we will then have `\PLUGIN_NAME\Robo\Plugin\Commands`, \PLUGIN_NAME\Robo\Plugin\EventHandler` or whatever is needed.

src/Runner.php
greg-1-anderson: For now just put it in a separate protected method. Most folks won't need to override this behavior, but it would be better to separate this out anyway.

tests/src/Commands/FirstCustomCommands.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Robo\Commands;
```
greg-1-anderson: This should probably be in some namespace other than the default pattern, so that tests must explicitly set the discovery namespace in order for these to be discovered. We don't want to pollute all of the functional tests with the plugin fixtures.

greg-1-anderson:
Looking pretty good. Thanks for taking this on.

ademarco:
Thank you for the review @greg-1-anderson ! I believe I've tackled all your comments, could you please check if I missed anything? After that I'll start working on event subscribers plugins and delegated container.

src/Runner.php
```diff
@@ -57,7 +68,7 @@ public function __construct($roboClass = null, $roboFile = null)
         $this->dir = getcwd();
     }
 
-    protected function errorCondtion($msg, $errorType)
+    protected function errorCondition($msg, $errorType)
```
greg-1-anderson: Typo's in the public API are a bummer, as fixing them is technically a b/c break. I think it's going to be rare enough for folks to want to override / call this that we can go ahead and fix it. Could you add a note in the CHANGELOG for this, though, so we don't forget?

tests/unit/ClassDiscovery/RelativeNamespaceDiscoveryTest.php
```diff
@@ -0,0 +1,92 @@
...
+class RelativeNamespaceDiscoveryTest extends \Codeception\Test\Unit
+{
+    private $ds = DIRECTORY_SEPARATOR;
+
+    public function testGetClasses()
+    {
+        $classLoader = new ClassLoader();
+        $classLoader->addPsr4('\\Robo\\PluginTest\\', [realpath(__DIR__.'/../../plugins')]);
+        $service = new RelativeNamespaceDiscovery($classLoader);
+        $service->setRelativeNamespace('Robo\Plugin');
```
greg-1-anderson: 👍

greg-1-anderson:
Looking really good; looking forward to seeing it finished up. A few docs on this mechanism would also be helpful, once the code is battened down.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### Unreleased
+
+* Typo in `\Robo\Runner:errorCondtion()` fixed as `\Robo\Runner:errorCondition()`.
```
greg-1-anderson: n.b. *Breaking* label (see 1.0 - it hasn't happened in a while.)

greg-1-anderson:
Please also compare against #604. One thing I see that I did over there is I converted the "examples" directory into a built-in plugin. This made it easier to run the examples, for sure, but maybe we do not want to expose all of the example commands in the `robo` phar.

The docs in the other PR might be a helpful starting point here.

docs/extending.md
```diff
@@ -33,6 +33,65 @@ Once you have done this, all of the tasks defined in the extension you selected
...
+
+Available commands:
+  help                  Displays help for a command
+  list                  Lists commands
+ my-project
+  my-project:command-one       
+  my-project:command-two        
+```
+
+Note: remember to run `composer dump-autoload` after adding new command classes.
```
greg-1-anderson: Is this necessary after adding new command classes, or only after adding new psr-4 namespaces?

greg-1-anderson:
One difference I notice compared to #604 is that this PR searches all namespaces for the relative pattern, whereas #604 only searches those PR namespaces that contain the pattern inside the namespace prefix declared in each composer.json file. Perhaps the technique used here is more flexible and easier for clients to get right. I don't know that the performance improvement gained from doing it the way #604 does it is significant enough to be worthwhile (although I did not measure).

It would be good to get some feedback from @DavertMik @thom8 @grasmash @weitzman (or some subset thereof) as to whether this does or does not meet use cases for plugin commands.

docs/extending.md
ademarco: Yeah, that's necessary after adding a new psr-4 namespace. Anyway, I've removed it as this is falls into knowing how to work with Composer.

ademarco:
@greg-1-anderson I've aligned it more with what was done on #604, e.g. now commands need to end in `Commands.php` in order to be discovered.

docs/extending.md
```diff
@@ -33,6 +33,66 @@ Once you have done this, all of the tasks defined in the extension you selected
...
+    public function commandOne() { }
+
+    /**
+     * @command my-project:command-two
+     */
+    public function commandTwo() { }
+}
+```
+
+Please note: command files classes must be placed under `Robo/Plugin/Commands` relative namepsace and their name
```
jfhovinne: Typo: 'namepsace'

drupol:
Very nice, can't wait for this to be merged!

greg-1-anderson:
Just to be clear, the only backwards-compatibility break here is fixing the spelling of the errorCondition method, right?

drupol:
Do you think it was used with a typo?

greg-1-anderson:
@drupol It's highly unlikely that anyone overrode this method, which is why we're going to accept the b/c "break" as a bugfix w/out bumping the Robo major version. If it causes a problem for a client, we could always roll back the spelling error.

drupol:
Indeed. :+1:

ademarco:
Yes the typo is the b/c break I'm referring to indeed.

greg-1-anderson:
OK, I am going to merge this to master as an experimental feature. Let's get some more feedback and refine as needed before making a stable release.

I'd like to see the `examples` directory implemented as a plugin in `autoload-dev`, so they are available in a dev checkout of Robo, but not in the phar. #604 does this. It would also be handy to do the same thing with the default commands in the RoboFile at the project root that are only useful for building Robo (e.g. `build:phar`).

drupol:
Ooooh yeah !

PR #675, opened 2018-02-02
Add a method Robo::getCommandInstance.

greg-1-anderson:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature
- [x] Has tests that cover changes (indirectly)

### Summary
Adds static method Robo::getCommandInstance

### Description
It is frequently a request by command authors, who, in order to implement some command `my:a` which in turn should execute other commands `my:b` and `my:c`, the desire of the author is to recover the commandfile instance for the subcommands so that the command's implementation methods may be called directly.

In general, this technique has been discouraged, because calling an implementation method directly omits any hook evaluation, so the result might not be the same as running `my:b` and `my:c` e.g. via exec. However, the advantage of having a reference to the command object is that the direct command result is available, whereas using exec or `$application->run()` all you would get is stdout.

### DIFF
--- a/src/Robo.php
+++ b/src/Robo.php
@@ -301,6 +301,49 @@ class Robo
     }
 
     /**
+     * Given the fully qualified namespace and classname of a command instance,
+     * instantiate it and store it in our container.
+     *
+     * @param string $commandClass
+     *
+     * @return null|object
+     */
+    public static function storeCommandInstance($commandClass)
+    {
+        $container = Robo::getContainer();
+
+        if (!class_exists($commandClass)) {
+            return;
+        }
+        $reflectionClass = new \ReflectionClass($commandClass);
+        if ($reflectionClass->isAbstract()) {
+            return;
+        }
+
+        $commandFileName = static::commandInstanceContainerName($commandClass);
+        $container->share($commandFileName, $commandClass);
+        return static::getCommandInstance($commandClass);
+    }
+
+    /**
+     * Given the fully qualified namespace and classname of a command instance,
+     * look it up from our container.
+     *
+     * @param string $commandClass
+     * @return object
+     */
+    public static function getCommandInstance($commandClass)
+    {
+        $commandFileName = static::commandInstanceContainerName($commandClass);
+        return static::service($commandFileName);
+    }
+
+    protected static function commandInstanceContainerName($commandClass)
+    {
+        return "{$commandClass}Commands";
+    }
+
+    /**
      * Retrieves a service from the container.
      *
      * Use this method if the desired service is not one of those with a dedicated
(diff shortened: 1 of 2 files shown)


dafeder:
Any chance to get this merged? Still interested in this!

greg-1-anderson:
Did you test it? Does it satisfy your use case? What is your use case?

pfrenssen:
I tested this out. My use case is to have a single command that sets up a development environment for a Drupal website. This command is intended to run multiple other Robo commands, to set up config files like `settings.php`, `services.yml`, etc.

In my test I tried to create a command that calls 2 other commands directly: `behat:generate-config` (which generates a `behat.yml` configuration file) and `drush:generate-config` (which generates a `drush.yml` file).

My commands live in a Robo extension and are set up using the instructions at https://robo.li/extending/#register-command-files-via-psr-4-autoloading. I am using Robo 2.2.1, and this PR applies cleanly on it. The commit that tests this PR is https://github.com/code-lab-eu/robo-drupal-setup/pull/1/commits/1c549ad89322046183acc402fd51ef51d83ff60a

I am getting an error, it appears that in my setup the command file I am trying to instantiate is not registered in the container.

```
$ ./vendor/bin/robo dev:setup
 [error]  Alias (ConfigFileCommands) is not being managed by the container
➜                                                                                                                                                              
➜   Requested RoboFile `/home/pieter/robo-drupal-setup/RoboFile.php` is invalid, please provide valid absolute path to load Robofile.
➜                                                                                                                                                              
➜
➜   Robo is not initialized here. Please run `robo init` to create a new RoboFile.
```

Also I am getting complaints about a missing `RoboFile` but since I expose commands in an extension I am expecting that these shouldn't require a `RoboFile`?

greg-1-anderson:
@pfrenssen: Could you compare with `$application->run()` in Robo 3? Capturing the output should work better there. Seems like perhaps you do not need the output for your use-case anyway, so perhaps this would also work well for you in Robo 2.

pfrenssen:
I tried with Robo 3 but this gives a similar error:

```
$ ./vendor/bin/robo dev:setup
 [error]  Alias (ConfigFileCommands) is not being managed by the container or delegates
Requested RoboFile `/home/pieter/v/event-venue/vendor/code-lab-eu/robo-drupal-setup/RoboFile.php` is invalid, please provide valid absolute path to load Robofile.
Robo is not initialized here. Please run `robo init` to create a new RoboFile.
```

I'm not clear what you mean by comparing with `$application->run()`, I am just getting started with Robo and I am trying out to build a standalone robo extension which I can include in my Drupal projects using Composer, so that I don't need to have a RoboFile.

PR #676, opened 2018-02-07
Implement the example RoboFile as a Robo plugin.

greg-1-anderson:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Description
Continuation of #671.

Implements Robo examples as a plugin. With this implementation, examples are available from a git checkout of Robo, but are not available in the phar.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
     },
     "autoload-dev":{
         "psr-4":{
-            "Robo\\":"tests/src"
+            "Robo\\":"tests/src",
+            "RoboExample\\":"examples/src"
         }
     },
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
@DavertMik @ademarco @drupol FYI feedback welcome.

I also timed it, and it takes about 3ms to find the commandfile directories in Robo, which I think is acceptable.

ademarco:
Looks good to me!

drupol:
Looks good to me too as well ! Simple and enough to let people understand how it works.
👍

PR #695, opened 2018-03-06
allow metadata to be set even when special is empty by making the . i…

flavius:
Allow metadata to be set even when special is empty by making the . in metadata optional.


greg-1-anderson:
Looks good; could you please add a couple of tests to confirm functionality with ad without the special section?

PR #715, opened 2018-04-07
Fix double escape when specifying a remoteShell with rsync. Add test.

peckrob:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
When specifying remoteShell() option with rsync, the result is double-escaped. For example:

```
            $this->taskRsync()
                ->fromPath('.')
                ->toHost($server)
                ->toUser('user')
                ->toPath('/var/www')
                ->remoteShell('ssh -i /home/user/.ssh/deploy')
                ->run();
```

Generates:

```
Running rsync --rsh ''\''ssh -i /home/user/.ssh/deploy'\''' . 'user@192.168.1.1:/var/www'
```

This changeset fixes the double escape and adds a test:

```
Running rsync --rsh 'ssh -i /home/user/.ssh/deploy' . 'user@192.168.1.1:/var/www'
```

### Description
Did not test on Windows (do not have the ability to do so easily).

### DIFF
--- a/src/Task/Remote/Rsync.php
+++ b/src/Task/Remote/Rsync.php
@@ -415,7 +415,7 @@ class Rsync extends BaseTask implements CommandInterface
      */
     public function remoteShell($command)
     {
-        $this->option('rsh', "'$command'");
+        $this->option('rsh', "$command");
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Thanks for this. I don't think many Windows users install rsync.

PR #716, opened 2018-04-09
Always return exec output (even when printed)

ipeevski:
### Overview
This pull request:

- [X] Fixes a bug
- [ ] Adds a feature

### Summary
Keep a copy of the output when printed so it can be returned as part of the result object

### Description
When printing the Process output and using a buffer function, the Process::getOutput() returns nothing. This patch keeps a copy of the output (via the buffer function) so it can return it in the end.


ipeevski:
boo ... sorry for the failed tests. I don't quite get why the expected output to 'ls' is 'Hello world' ... any insight?

PR #718, opened 2018-04-11
Adding a printResult parameter for Robo\Result, allowing users to fully suppress errors.

jonpugh:
I'm trying to allow Provision to hide error output. I can't seem to figure out how to prevent Result from calling Result::printResult()

This PR doesn't actually work for me, but I'm hoping to get some advice on where to go from here.

### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Result() has an extra parameter, $printResult that defaults to TRUE.

### Description
Provision PR coming soon.

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -22,12 +22,14 @@ class Result extends ResultData
      * @param string $exitCode
      * @param string $message
      * @param array $data
+     * @param bool $printResult
      */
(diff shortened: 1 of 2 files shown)


jonpugh:
This is clearly not desirable, I was just seeing if I could get it to work. 

See #720 for a simpler PR that gives me the desired effect.

PR #720, opened 2018-04-17
Block printing of result inside Result constructor

jonpugh:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

I know we can't just block all output like this, but I figured I'd get it how I need it to be and figure out a way to work backward from there.

### Summary
I can't stop error messages from printing (along with that [CompletionWrapper] class indicator) in Provision ~tasks~ steps. 

When I run `provision verify` and it fails it looks like this:
screenshot from 2018-04-17 15-45-59

If I add Robo with this PR (commenting out printResult()) I get this (which is what I want):
screenshot from 2018-04-17 15-46-49

It's really important for user experience. With this patch, the checkboxes [ ]  turn into :x: as the steps complete, it's really nice.

I tried everything, ->silent(), printOutput(), everything I could find, but I guess since it is hard coded into the constructor it's not affected by these.

Thoughts?

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -27,7 +27,7 @@ class Result extends ResultData
     {
         parent::__construct($exitCode, $message, $data);
         $this->task = $task;
-        $this->printResult();
+        // $this->printResult();
 
         if (self::$stopOnFail) {
             $this->stopOnFail();


greg-1-anderson:
We can't just remove this line from the Result constructor -- and believe me, I'd like to get rid of it -- because we would break existing uses. In your use case, we go from double-printing the error to showing it just once, but in other instances it would go from a single print to no output for the error at all.

Maybe if we migrated all of your output styling code into Robo and made a Robo 2.0, we could break b/c and make this look nice like the second diagram. We could probably do this by moving the result-printing code out of the Result class and into the task collection.

What do you think, @DavertMik? Robo 1.0.0 was tagged on 13 Oct 2016. Is 18 months too soon for a mostly-compatible 2.0.0? Most RoboFiles and tasks would probably still work.  I'm ready to let go of my original goal to only use the PSR-3 logger in TaskInfo and just use SymfonyStyle directly. 😄 I'd also like to clean up the Robo app startup code a bit.

Issue #650, opened 2017-11-29
I want use Add and Delete by watch task.


tomcat0090:
I want use Add and Delete by watch task.
but, I can't change option.
because, lurker option is hard code.

see: https://github.com/consolidation/Robo/blob/master/src/Task/Base/Watch.php#L80

Are there any good ideas?

PR #751, opened 2018-07-15
Events as third parameter in watch function

bbasinski:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Adds ability to set multiple events to file/dirs.

### Description
Robo has hard coded watching for modification event only which is not firing when files are added or removed. To give possibility to react on this kind of changes monitor function now can take third argument which stands for events. Like in paths it can be array or single value. To maintain backward compatibility default event was set to modify event. This PR is related to #650.

Usage:
```php
<?php

$this->taskWatch()
  ->monitor(
    'src', 
    function() {
      $this->taskExec('phpunit')->run();
    },
    \Lurker\Event\FilesystemEvent::CREATE
)->run();

$this->taskWatch()
  ->monitor(
    'src', 
    function() {
      $this->taskExec('phpunit')->run();
    },
    [
      \Lurker\Event\FilesystemEvent::CREATE, 
      \Lurker\Event\FilesystemEvent::DELETE
    ]
)->run();
```

### DIFF
--- a/src/Task/Base/Watch.php
+++ b/src/Task/Base/Watch.php
@@ -55,15 +55,7 @@ class Watch extends BaseTask
      */
     public function monitor($paths, \Closure $callable, $events = FilesystemEvent::MODIFY)
     {
-        if (!is_array($paths)) {
-            $paths = [$paths];
-        }
-
-        if (!is_array($events)) {
-            $events = [$events];
-        }
-
-        $this->monitor[] = [$paths, $callable, $events];
+        $this->monitor[] = [(array)$paths, $callable, (array)$events];
         return $this;
     }
 
--- a/src/Task/Base/Watch.php
+++ b/src/Task/Base/Watch.php
@@ -1,7 +1,6 @@
 <?php
 namespace Robo\Task\Base;
 
-use Lurker\Event\FilesystemEvent;
 use Lurker\ResourceWatcher;
 use Robo\Result;
 use Robo\Task\BaseTask;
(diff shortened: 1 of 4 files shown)


src/Task/Base/Watch.php
```diff
@@ -49,15 +49,21 @@ public function __construct($bindTo)
...
         if (!is_array($paths)) {
             $paths = [$paths];
         }
-        $this->monitor[] = [$paths, $callable];
+
+        if (!is_array($events)) {
+            $events = [$events];
+        }
+
+        $this->monitor[] = [$paths, $callable, $events];
```
greg-1-anderson: Typecase as `(array) $events` and the `if` block above should not be necessary, I think.

greg-1-anderson:
Looks good. Could you provide docs and a test?

src/Task/Base/Watch.php
bbasinski: You mean it should allow only one event?

greg-1-anderson: No; if `$events` is an array, then the statement `(array) $events` is a no-op. If `$events` is a scalar, then `(array) $events` is equivalent to `$events = [$events];`. That's standard but perhaps lesser-known PHP array behavior.

bbasinski: Didn't know about this. Thanks.

tests/unit/Task/WatchTest.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+use AspectMock\Test as test;
+
+class WatchTest extends \Codeception\TestCase\Test
```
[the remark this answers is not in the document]
greg-1-anderson: It's okay to ignore Nitpick-CI advice for things that are unimportant, e.g. test class style

greg-1-anderson:
The Nitpick / Scrutinizer issues are not significant. Thank you for this contribution.

Issue #768, opened 2018-08-27
Issues with env() in Exec


q0rban:
### Steps to reproduce
1. Create a task that executes the `env` command. Note the environment variables.
2. To this task, add a call to `->env('FOO', 'BAR')`. Note that many of the above environment variables are missing.
3. Add an additional call to `->env('BAR', 'BAZ')`. Note that `FOO=BAR` is now missing from the environment variables.

### Expected behavior
1. Using `->env()` should inherit existing environment variables. (This will be default behavior in Symfony 4)
2. Calling `->env()` more than once should not remove previously set environment variables.

### Actual behavior
1. Many of the existing environment variables are removed, including on some environments things like `PATH` which can break the execution of certain scripts. (For example, a script that calls `#!/usr/bin/env php` might fail if PATH is missing with `php not found`)
2. Previously set environment variables in a Robo task are wiped out.

### System Configuration
MacOS, PHP 7.2

greg-1-anderson:
A PR for this would be welcome.

PR #769, opened 2018-08-27
Ensure that environment variables are preserved in Exec #768

q0rban:
### Overview
This pull request:

- [X] Fixes a bug

### Summary
See #768

### DIFF
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -313,7 +313,9 @@ trait ExecTrait
         if (isset($this->env)) {
             // Symfony 4 will inherit environment variables by default, but until
             // then, manually ensure they are inherited.
-            $this->process->inheritEnvironmentVariables();
+            if (method_exists($this->process, 'inheritEnvironmentVariables')) {
+                $this->process->inheritEnvironmentVariables();
+            }
             $this->process->setEnv($this->env);
         }
 
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -311,6 +311,9 @@ trait ExecTrait
         }
 
         if (isset($this->env)) {
+            // Symfony 4 will inherit environment variables by default, but until
+            // then, manually ensure they are inherited.
+            $this->process->inheritEnvironmentVariables();
             $this->process->setEnv($this->env);
         }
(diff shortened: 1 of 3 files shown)


q0rban:
Getting a false positive on this test: https://travis-ci.org/consolidation/Robo/jobs/421167983#L866

Any ideas on a better way to test this? The test fails on my local, but not on Travis.

q0rban:
Ah, looks like appveyor fails, though: https://ci.appveyor.com/project/greg-1-anderson/robo/build/1.0.568#L845

Good! Now to push up the fixes.

greg-1-anderson:
The behavior of inherited environment variables varies depending on the configuration of your php installation. I don't know if you can downgrade this via an `ini_set`.

q0rban:
@greg-1-anderson I changed the test to count the number of environment variables in 6bd0a72, which I think should make that test more consistent across environments. I tested it on my local without the fix and it failed as expected.

q0rban:
Any thoughts on this @greg-1-anderson ?

greg-1-anderson:
This is okay with passing tests. It looks like only the "lowest" tests are failing. It's okay to just bump the minimum required version of all of the components in the composer.json to be the current version that you have in composer.lock, as long as you do not remove support for any major version of any component (e.g. Symfony 2).

greg-1-anderson:
Maybe it would be better to check in code whether `inheritEnvironmentVariables` is available, and not use it if it is not. That would avoid breaking Symfony 2 builds. I don't think we're ready to require symfony/proces:^3 everywhere here yet.

greg-1-anderson:
That last commit looks like the right approach. Could you back out any unnecessary version pins in composer.json? There's no need to get this exactly right, but it'd be better to keep unrelated components at the same minimum version on a best-effort basis.

q0rban:
It probably wasn't obvious from the commit when you looked at it, but I had reverted back to prior to the version changes to composer.json, made changes to check for the existence of the function (after realizing that no version of Symfony 2.x's Process had the function), and force pushed so that none of those version changes would be in the PR.

greg-1-anderson:
Looks great -- thanks for pushing this through to completion.

PR #771, opened 2018-08-28
Don't install curl; just use the one bundled with git.

greg-1-anderson:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Appveyor changed the way they handle curl installs.


greg-1-anderson:
This PR seems to not have build on Appveyor; that may indicate that there is something wrong with the change.

It also appears that the previous Appveyor configuration is building again without this modification. Closing.

PR #776, opened 2018-09-05
Add IO::confirmDefault() method

TravisCarden:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Add a `confirmDefault()` method to `\Robo\Common\IO`.

### Description
I was looking for such a function today. I expected it to be present as a matter of symmetry with `ask()`/`askDefault()`.

### DIFF
--- a/src/Common/IO.php
+++ b/src/Common/IO.php
@@ -134,6 +134,18 @@ trait IO
     }
 
     /**
+     * @param string $question
+     * @param bool $default
+     *
+     * @return string
+     */
+    protected function confirmDefault($question, $default)
+    {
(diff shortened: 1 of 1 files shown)


danepowell:
Robo inherits a lot of this from Symfony, and the upstream version of confirm() already takes a default, which we just ported into Robo: https://github.com/consolidation/Robo/pull/850

I wonder if it would make more sense to just clean up the output of that method rather than creating a new one.

greg-1-anderson:
Yeah, this is either just a duplicate of #850, or needs a re-roll post-#850 if the formatting additions here are still desired.

PR #788, opened 2018-10-16
Support Windows line endings in ".semver" file

### LINKED ISSUE #774: Allow CRLF in SemVer task (i.e. support Windows line endings)
### Steps to reproduce
1. Create a SemVer task : `$this->taskSemVer()->increment('major')->run();`
2. Create a `.semver` file :

```
---
:major: 2
:minor: 0
:patch: 0
:special: ''
:metadata: ''
```

### Expected behavior
No error: the `major` component is incremented in the `.semver` file.

### Actual behavior
The task fails:

```
[error]    in task Robo\Task\Development\SemVer

  Bad semver file.
```

### System Configuration
OS: Windows 10
PHP: v7.2.9
Robo: v1.3.1

cedx:
Fixes #774


.gitattributes
```diff
@@ -9,3 +9,5 @@
 /.travis.yml        export-ignore
 /.scrutinizer.yml   export-ignore
 /appveyor.yml       export-ignore
+
+/tests/_data/semver.crlf   text eol=crlf
```
greg-1-anderson: -1 on this. Rather than require just one file has Windows line endings, it would be better to assume that all of the text fixtures are consistently stored with Unix line endings. Then, in the tests, read in the fixture, convert the line endings, and write it to a tmp file to use in the test.

greg-1-anderson:
Thanks for this. I have just one request related to the tests; see comment on the code.

.gitattributes
cedx: OK, I'll fix it.

greg-1-anderson:
Looks great. Thanks for fixing up the test fixture.

cedx:
Thanks for accepting this code 😃

PR #807, opened 2018-11-27
Revise passing state documentation

justinlevi:
Update the alternative approach to using returned results to be more inline with the suggestions proposed here:
https://github.com/consolidation/Robo/issues/382

Using the Symfony Process component for returning simple system info.

### Overview
This pull request:

Attempts to improve the documentation a little.

### DIFF
--- a/docs/collections.md
+++ b/docs/collections.md
@@ -282,7 +282,19 @@ More complex task configuration may be done via the `defer()` method. `defer()` 
         ->run();
 ?>
 ```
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
The focus of that section of the documentation is how to pass information from one task to another. The example of using PHP's `exec` is just a side point that one should only use `defer` for complex task collections. It's off-topic to diverge into a discussion on how to call the Symfony Process component here.

Optionally, if you wanted to put that example somewhere else, then we could include a link to it from this section.

If you do move this addition, I think you can use `$process->run()` instead of `start()` + `wait()`, and you are missing a `$` on the last line of the code example.

PR #813, opened 2018-12-13
Ensure nested collections with rollbacks preserve correct order of execution

q0rban:
### Overview
This pull request:

### Summary
Adds to the rollback order test with a nested collection with rollbacks.


greg-1-anderson:
Thanks!

PR #815, opened 2018-12-14
Allow for aborting completions or rollbacks.

q0rban:
### Overview
This pull request:

- [X] Adds a feature

### Summary
Adds a new exception `ForcedException` for forcefully aborting a rollback or completion.

### DIFF
--- a/src/Exception/ForcedException.php
+++ b/src/Exception/ForcedException.php
@@ -1,0 +1,7 @@
+<?php
+namespace Robo\Exception;
+
+class ForcedException extends \Exception
+{
+
+}
--- a/src/Exception/ForcedException.php
+++ b/src/Exception/ForcedException.php
@@ -1,7 +1,0 @@
-<?php
-namespace Robo\Exception;
-
-class ForcedException extends \Exception
-{
-
-}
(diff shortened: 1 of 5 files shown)


src/Exception/ForcedException.php
```diff
@@ -0,0 +1,4 @@
+<?php
+namespace Robo\Exception;
+
+class ForcedException extends \Exception {}
```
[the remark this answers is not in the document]
q0rban: Old habits die hard.

src/Exception/ForcedException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+namespace Robo\Exception;
+
+class ForcedException extends \Exception
```
greg-1-anderson: Naming things is hard. "Forced" exception seems unclear to me. "AbortRollbackException" would be more clear, except that this might also happen in a completion function. It's also possible to use `ignoreErrorsTaskWrapper` for a purpose completely unrelated to rollback / completion (although I don't know that this is very common).

How about `AbortTasksException`?

A comment in this class file to complement the documentation already existing elsewhere (in this PR) would also be helpful.

q0rban: @greg-1-anderson that works for me! Updated.

greg-1-anderson:
LGTM. Thanks.

PR #841, opened 2019-03-08
Added the exists() method to FilesystemStack

legovaer:
### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Adds a missing `exists()` method to the taskFilesystemStack.

### Description
For a project I needed to check if a given directory and/or file exists. As this method exists on the [Filesystem component of Symfony](https://github.com/symfony/symfony/blob/4.2/src/Symfony/Component/Filesystem/Filesystem.php), I thought it would be nice to add this method to Robo's FilesystemStack as well.

### DIFF
--- a/src/Task/Filesystem/FilesystemStack.php
+++ b/src/Task/Filesystem/FilesystemStack.php
@@ -36,6 +36,7 @@ use Robo\Common\BuilderAwareTrait;
  * @method $this remove(string|array|\Traversable $file)
  * @method $this rename(string $from, string $to, bool $force = false)
  * @method $this symlink(string $from, string $to, bool $copyOnWindows = false)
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thank you for contributing to Robo; however, this method does not fit the usage model of the Robo filesystem stack. If you need to check if a file exists, you should use the lower-level APIs (Symfony Filesystem, or simply `file_exists`). Robo filesystem stacks are designed to queue up commands to run later. Each method should return a reference to the task itself, so that methods may be chained together. Having a method that returns bool interferes with this pattern.

legovaer:
Gotcha, thanks for pointing that out 👍

PR #850, opened 2019-03-29
Fixes #849: Allow default argument for confirm() questions.

### LINKED ISSUE #849: Allow IO::confirm() to take a default value
It would be nice to be able to easily create confirmation dialogs with a default value other than "no".

It would be trivial to accept a default argument for confirm() here, would you be open to a PR for this?
https://github.com/consolidation/Robo/blob/c2daf4e90b8cd6e896f880fac4110e7bc998a7d5/src/Common/IO.php#L131

danepowell:
### Overview
This pull request:
- [x] Adds a feature

### Summary
Allow a default argument for confirm() prompts so they can be run non-interactively (BLT currently has to override this method to allow a default argument). This also aligns more closely with the super method in Symfony, which takes a default argument.

It's a somewhat trivial change, not sure if it needs dedicated tests.

### DIFF
--- a/src/Common/IO.php
+++ b/src/Common/IO.php
@@ -125,12 +125,13 @@ trait IO
 
     /**
      * @param string $question
+     * @param bool $default
      *
      * @return string
      */
-    protected function confirm($question)
+    protected function confirm($question, $default = false)
     {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
LGTM, thanks.

PR #851, opened 2019-04-02
Rename callCollectionStateFuntion to callCollectionStateFunction

q0rban:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
There was a typo in a method name in the CollectionBuilder class. Thankfully it is a protected method so it should be somewhat safe from API breaks.


greg-1-anderson:
Yeah, it's unlikely that anyone will be calling that method directly (only possible from CollectionBuilder derived classes, which should be rare). If it causes a problem, we can add back the misspelled version of the method.

Thanks for the fix.

Issue #852, opened 2019-04-06
Prevent undefined variable warning in ImageMinify task


pixelbrackets:
### Steps to reproduce
Pass an image folder to ImageMinify task, which contains images not in the default list (png/jpg/gif/svg).

```bash
$ ls assets/images/
favicon.png
favicon.ico
```

```php
$this->taskImageMinify('assets/images/*')
    ->to('dist/images/')
    ->run();
```

### Expected behavior
Show a warning if no minifier is found for a given file extension.

### Actual behavior
Robo shows an error `Minifier method  cannot be found!`

Robo also throws a PHP warning/notification `PHP Notice:  Undefined variable: minifier in /somepath/vendor/consolidation/robo/src/Task/Assets/ImageMinify.php on line 374`

### System Configuration
Ubuntu 16.04
PHP 7.0.33

### Proposed solution

Define the missing variable at the beginning of the method.

greg-1-anderson:
Fixed in master and 1.x. Thanks.

PR #853, opened 2019-04-06
ImageMinify: Prevent undefined variable error

pixelbrackets:
If the given imagefile is not one of the defaults (png/jpg/gif/svg) and no minifier was passed to the ImageMinify task a warning for an undefined variable is show.

Prevent this by setting an empty string to the beginning of the file check.

Refs #852

### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Adds an empty variable string to the beginning of the loop to check for matching minifiers. This will prevent trying to access a non existing variable later on.

### Note

This bugfix is for the master (Development version 2.x).

It may also be cherry-picked to the `1.x` branch as well (I could send a second PR as well).

### DIFF
--- a/src/Task/Assets/ImageMinify.php
+++ b/src/Task/Assets/ImageMinify.php
@@ -339,6 +339,8 @@ class ImageMinify extends BaseTask
 
         // loop through the files
         foreach ($files as $from => $to) {
+            $minifier = '';
+
             if (!isset($this->minifier)) {
                 // check filetype based on the extension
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks. A second PR would be helpful, as I wouldn't forget to cherry-pick this later if you did  that.

pixelbrackets:
Thanks for the fast merges! :)

PR #864, opened 2019-05-08
Updating phpdoc to reflect correct typecast

davidsneighbour:
### Overview
The parameter roboClass of the constructor accepts string and array (as per https://robo.li/framework/), but is commented so that only string is expected.

- [x ] Fixes a bug
- [ ] Adds a feature


davidsneighbour:
No need then [year end cleanup :)]

Issue #871, opened 2019-05-14
Add ability to set RoboFile and RoboClass.


albertski:
My scenario is this:
I have RoboFile.php that has some methods.  I also have MyProjectRoboFile.php that has some methods.  Currently, I have to rename MyProjectRoboFile.php to RoboFile.php in order for it to run.

It would be nice if could pass RoboFile and RoboClass as an option to the robo command.

albertski:
Before I create a pull request would this be something you guys would be okay with:

Current robo file:
```
$runner = new \Robo\Runner();
```

New robo file:
```
// Get Robo Class or Robo File set as an option.
$roboClass = NULL;
$roboFile = NULL;
foreach ($_SERVER['argv'] as $key => $argv) {
  if (substr($argv, 0, 12) == '--robo-class=') {
    $roboClass = substr($argv, 12);
    unset($_SERVER['argv'][$key]);
  }
  elseif (substr($argv, 0, 11) == '--robo-file=') {
    $roboFile = substr($argv, 11);
    unset($_SERVER['argv'][$key]);
  }
}
```

greg-1-anderson:
Try `--load-from`

albertski:
Oh wow. Didn't notice that. That worked great.

greg-1-anderson:
@koppieesq ^^

Issue #870, opened 2019-05-10
"Getting Started" needs to be "Getting Started"-er


koppieesq:
This is a stub ticket for a task to update the "Getting Started" documentation to make it friendlier for users who are just getting started.

greg-1-anderson:
#871

PR #872, opened 2019-05-17
Clearer explanation of --load-from option.

koppieesq:
This is in response to #871 and #870.  Simple modification to the getting started documentation, to make it a little more clear what the `--load-from` option can be used for.

This is my first PR for this project, so please be gentle.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -129,12 +129,19 @@ No other values should be used for the default value. For example, `$options = [
 
 ### Load From Other Robofile
 
-Robo can execute commands from a different RoboFile, eg. located in different directory or with a different filename.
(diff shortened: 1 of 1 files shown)


Sweetchuck:
I think the documentation should expose following facts:

- The file name can be anything, it is not limited to `RoboFile.php`
- The class name inside the file has to be the same name as the file has.
- The class has to be in the root namespace. `Foo.php` => `\Foo`
- The internal current directory (cwd) of the PHP process will be switched to the directory where the provided RoboFile came from.

koppieesq:
Thanks for the input @Sweetchuck - I've taken your recommendations (see new commit).

greg-1-anderson:
LGTM - thanks!

PR #878, opened 2019-06-21
Allow command classes to end in Commands or Command.

Jelle-S:
### Overview
This pull request:

- [x] Adds a feature

### Summary
This PR allows command classes to end in both Commands or Command, and (because of PSR-4) the files in Commands.php or Command.php

### Description
In my head it makes more sense to have a command file/class per command, in stead of one with multiple. You could perfectly do this right now, but the filename wouldn't make sense. It's a bit of a nitpick, but it just seems cleaner in my mind.


greg-1-anderson:
Sure, makes sense. Thanks.

Jelle-S:
Any idea on when this might be added? :)

greg-1-anderson:
Maybe this will go in faster if I actually press that pretty "Squash and merge" button. Let's try that.

:stuck_out_tongue_closed_eyes:

PR #882, opened 2019-07-17
Add $stopOnFail flag to \Robo\Task\Base\ParallelExec

Slamdunk:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
ParallelExec is useful on CI to parallelize steps, but CI bill the time consumed, to it turns out that if one parallel process fails we can save money by stopping other parallel process too


src/Task/Base/ParallelExec.php
```diff
@@ -172,6 +188,12 @@ public function run()
                         }
                     }
                     unset($running[$k]);
+                    if ($this->stopOnFail && $process->getExitCode() !== 0) {
+                        foreach ($running as $otherRunningProcess) {
+                            $otherRunningProcess->stop();
+                        }
+                        $queue = [];
```
Slamdunk, line 208: I don't touch `$running` array because it feel safer to me to let it handle by the main `while` cycle.

src/Task/Base/ParallelExec.php
```diff
@@ -172,6 +188,12 @@ public function run()
                         }
                     }
                     unset($running[$k]);
+                    if ($this->stopOnFail && $process->getExitCode() !== 0) {
+                        foreach ($running as $otherRunningProcess) {
+                            $otherRunningProcess->stop();
```
Slamdunk, line 206: This can't be done without prefixing everything with `exec` command, read:

1. https://symfony-documentations.readthedocs.io/en/2.6/components/process.html#process-signals
1. https://github.com/symfony/symfony/issues/5759
1. https://bugs.php.net/bug.php?id=39992

greg-1-anderson: Hm, I am ambivalent about this. Would it be possible to use `posix_setpgid` as described in the PHP bug report to fix things for those who have the posix commands?

tests/cli/StopOnFailureCest.php
```diff
@@ -0,0 +1,27 @@
...
+use \CliGuy;
+
+class StopOnFailureCest
+{
+    public function _before(CliGuy $I)
+    {
+        $I->amInPath(codecept_data_dir('sandbox'));
+    }
+
+    public function parallelProcessesGetStoppedIfStopOnFailIsSet(CliGuy $I)
```
greg-1-anderson: Nice test. If `posix_setpgid` works, make a second test just like this one that does not use `exec`, and markTestSkipped if the posix commands are not available.

src/Task/Base/ParallelExec.php
Slamdunk: I'll try

tests/cli/StopOnFailureCest.php
Slamdunk: I'm sorry, I just realized I've never read you change request: can you explain me what you are asking for? I don't get it :confused:

greg-1-anderson (changes requested):
This looks like a very useful enhancement.

1. The requirement for using `exec` might not be clear to new users. Try using `posix_setpgid` and see if you can stop the other processes even if `exec` is not used.
2. This feature needs documentation. The requirement for `exec` should be explained even if `posix_setpgid` works, because the code should operate even if the posix commands are not available. In this instance, users will need to employ `exec`
3. Rename `stopOnFail` to something else, as this feature behaves differently than the global stopOnFail flag.
4. The Codeception tests are no longer maintained in Robo. Please convert the test to phpunit (see the tests/integration directory)
5. There should be two tests, one that uses exec (like the existing test), and another that does not. The second test should be skipped if the posix commands are not available.

tests/cli/StopOnFailureCest.php
greg-1-anderson: I added a summary of the requested changes. LMK if anything remains unclear.

Slamdunk:
> 1. The requirement for using `exec` might not be clear to new users. Try using `posix_setpgid` and see if you can stop the other processes even if `exec` is not used.
> 2. This feature needs documentation. The requirement for `exec` should be explained even if `posix_setpgid` works, because the code should operate even if the posix commands are not available. In this instance, users will need to employ `exec`
> 5. There should be two tests, one that uses exec (like the existing test), and another that does not. The second test should be skipped if the posix commands are not available.

Since `v3.3.0` Symfony\Process already does the work, so `exec` here is no longer needed:

https://github.com/symfony/process/commit/4ad0b868fc001769ea4fb8b860d6f56383f530fe#diff-9a01fc0e340da4c3f1e4a16029a63977R296-R299

> 3. Rename `stopOnFail` to something else, as this feature behaves differently than the global stopOnFail flag.
> 4. The Codeception tests are no longer maintained in Robo. Please convert the test to phpunit (see the tests/integration directory)

Done. I've also rebased the PR against `2.x` branch.

greg-1-anderson (changes requested):
Looks good, but PRs need to go into the default branch (3.x) first. Looks like this will probably rebase against 3.x without any trouble.

Slamdunk:
> Looks good, but PRs need to go into the default branch (3.x) first. Looks like this will probably rebase against 3.x without any trouble.

If you are going to backport it into `2.x` after the merge, that's a branching model new to me; I've always seen the contrary: everything BC compatible into stable branches, forward ported after merge, see https://github.com/laminas/automatic-releases/blob/1.9.x/docs/branching-model.svg

If you are not going to backport it into `2.x` after the merge, may I ask you why?

PR #886, opened 2019-07-23
Fix PHPDoc @param and @return type hints

Sweetchuck:
This pull request contains only PHPDoc modifications.
For example `@param $foo` vs `@param string $foo`

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -59,8 +59,9 @@ class RoboFile extends \Robo\Tasks
     /**
      * Generate a new Robo task that wraps an existing utility class.
      *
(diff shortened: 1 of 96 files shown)


greg-1-anderson:
Also contains one spelling error fix, but the misspelled method name in question is protected, so there is no need to maintain a deprecated misspelled version.

drupol:
Nice! Can't wait to see this in a release, it will fix the PHPStan issue that [we have](https://github.com/php-taskman/github/issues/7) on [php-taskman/github](https://github.com/php-taskman/github).

Thanks!

PR #888, opened 2019-07-24
taskPack modified to correctly set file permissions in ZIP files (ins…

alex-scott:
…tead of default 0666)

### Overview
This pull request:
- [x] Adds a feature

### Summary
taskPack modified to correctly set file permissions in ZIP files

### Description
Without these changes, on any Unix/Linux system, created ZIP files extracts with all files having permissions 0666 and there is no way to change it. This fix adds code to automatically set permissions in archive same to permissions of original files.

PR #897, opened 2019-10-05
taskPack modified to correctly set file permissions in ZIP files on Unix and on Windows NTFS

MirazMac:
### Overview
This pull request:

- [x] Adds a feature

### Summary
taskPack modified to correctly set file permissions in ZIP files on Unix and on Windows NTFS.

### Description
Without these changes, on any *nix or Windows system, created ZIP files stores the files with the permission there is no way to change it. This fix adds code to automatically set permissions in archive same to permissions of original files for *nix systems. And since on Windows the file permissions are different and even keeping the original permissions doesn't work there. So, it sets the files permissions to 0644 by default on Windows.

Referenced here but the pull wasn't merged because of the coding style issue. I fixed it and added the Windows system workaround. 
#888

### DIFF
--- a/src/Task/Archive/Pack.php
+++ b/src/Task/Archive/Pack.php
@@ -245,10 +245,41 @@ class Pack extends BaseTask implements PrintedInterface
                     if (!$zip->addFile($file->getRealpath(), "{$placementLocation}/{$relativePathname}")) {
                         return Result::error($this, "Could not add directory $filesystemLocation to the archive; error adding {$file->getRealpath()}.");
                     }
+                    // This will retain the file permissions for *nix systems
+                    // Source: https://github.com/consolidation/Robo/pull/888
+                    if (stripos(PHP_OS, 'WIN') !== 0) {
+                        $zip->setExternalAttributesName(
+                            "{$placementLocation}/{$relativePathname}",
+                            \ZipArchive::OPSYS_UNIX,
+                            $file->getPerms() << 16
+                        );
+                    } else {
+                        // For windows we force 0644 for all the files
+                        // Because it seems when the archive is created on Windows the files by default has
+                        // 0666 permission (can be seen when unpacked on a *nix system)
+                        $zip->setExternalAttributesName(
+                            "{$placementLocation}/{$relativePathname}",
+                            \ZipArchive::OPSYS_WINDOWS_NTFS,
+                            33188 // Is the decimal for 100644
+                        );
+                    }
                 }
             } elseif (is_file($filesystemLocation)) {
                 if (!$zip->addFile($filesystemLocation, $placementLocation)) {
                     return Result::error($this, "Could not add file $filesystemLocation to the archive.");
+                }
+                if (stripos(PHP_OS, 'WIN') !== 0) {
+                    $zip->setExternalAttributesName(
+                        $placementLocation,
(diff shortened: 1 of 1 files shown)


src/Task/Archive/Pack.php
```diff
@@ -245,11 +245,42 @@ protected function addItemsToZip($zip, $items)
...
+                            $file->getPerms() << 16
+                        );
+                    } else {
+                        // For windows we force 0644 for all the files
+                        // Because it seems when the archive is created on Windows the files by default has
+                        // 0666 permission (can be seen when unpacked on a *nix system)
+                        $zip->setExternalAttributesName(
+                            "{$placementLocation}/{$relativePathname}",
+                            \ZipArchive::OPSYS_WINDOWS_NTFS,
+                            33188 // Is the decimal for 100644
```
greg-1-anderson: This would be clearer if expressed in octal (`0100644`).

I guess there is no +x bit on Windows? [is-executable](https://www.php.net/manual/en/function.is-executable.php) returns `true` only for `.exe` files, which is pointless on Linux. I guess manipulating Linux files on a Windows system just has to be an edge case that we don't / can't support.

MirazMac: Using Octal expression doesn't work on Windows, it seems only decimal values work on windows via setExternalAttributesName(). 
And this is not an edge case, many of the PHP devs develop on windows and may create an archive to distribute on linux servers. It causes a huge problem, like if the index/root file has a permission of 0666, it shows an 403 forbidden error when accessed via HTTP and file compressed via Robo in windows has 0666 permission for all the files(except directories, they have 0755).
So, I guess it would be an essential things to implement, I'm using this locally and it's working for me and someone may face the same issue in future that's why I created the pull request, it's your choice if you want to merge or not.
Thanks.

greg-1-anderson: What do you get when you run this on Windows?
```
php -r 'print 0100644;'
```

MirazMac: Running php -r 'print 0100644'; shows a syntax error, but php -r 'print(0100644)'; running this outputs nothing

MirazMac: By saying octal expression doesn't work on windows I meant about the setExternalAttributesName(); method's parameter, not the OS entirely.

greg-1-anderson: That seems impossible to me; regardless of whether you represent a number in octal or decimal in the source, the data passed to a method parameter should be the same. Can you explain why `setExternalAttributesName()` fails with an octal parameter? What is at work here?

greg-1-anderson: Please provide a test.

MirazMac: It seems it's an internal bug of PHP, besides setExternalAttributesName() isn't properly documented, the only example they provided is for the unix system and while on windows the octal values doesn't change the permission as expected, I'm kinda busy to write the tests, and I'm not sure what to tests against, as extracting the zip on windows to check if the has the permission of 0644 or not won't work on Windows because that's not how windows permissions work at all.
I'm closing this, thanks.

greg-1-anderson:
Maybe someone else will want to move this forward later.

The behavior of `setExternalAttributesName()` should not vary based on whether octal or binary is used.

PR #915, opened 2019-11-04
Patch 2

### LINKED ISSUE #909: Static analysis
```bash
composer require --dev phpstan/phpstan:^0.9.0
vendor/bin/phpstan analyze --level=0 -c phpstan.neon src/
```

then raise Level one by one

`phpstan.neon`

```yaml
parameters:
    level: max
    ignoreErrors:
        # Uses func_get_args()
        - '#^Method Robo\\Tasks::task\(\) invoked with [12345] parameters?, 0 required\.$#'
```

Thank you for considering!
See my recent PR-s.

szepeviktor:
Fixes #909

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,11 @@ install:
 
 script:
   - composer test
+  - |
+    if [ "$(phpenv version-name)" == "7.2" ]; then
+      composer require --dev "phpstan/phpstan-shim:^0.11.0"
+      vendor/bin/phpstan analyze
+    fi
 
 after_success:
   - travis_retry php vendor/bin/php-coveralls -v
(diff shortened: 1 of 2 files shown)


szepeviktor:
@greg-1-anderson Pls help me resolve these composer mysteries.

greg-1-anderson:
Odd. It's failing because it's trying to create scenarios for the "highest" case, which isn't necessary and shouldn't be done. I don't know what, if anything, from your PR is causing the failure. Perhaps it's always been creating scenarios during "highest" testing, but it simply works (wasting time but otherwise innocuous) in other tests. The best solution is to make it not create scenarios.

PR #931, opened 2020-03-02
ignore links to tests folder on export

medisun:
### Overview
This pull request:

- [x] Fixes a bug

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,5 @@
 /.travis.yml        export-ignore
 /.scrutinizer.yml   export-ignore
 /appveyor.yml       export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -8,3 +8,5 @@
 /.travis.yml        export-ignore
 /.scrutinizer.yml   export-ignore
 /appveyor.yml       export-ignore
+/scenarios/symfony2/tests export-ignore
+/scenarios/symfony4/tests export-ignore
```
greg-1-anderson: This directory should be `.scenarios.lock`, not `scenarios`, right? Shouldn't we just exclude all of the scenarios from export?

PR #934, opened 2020-03-05
added the --no-suggest option to composer tasks

tkleinhakisa:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Added the ability to use composer install, update and require tasks with the `--no-suggest` option

### Description
```php
$this->taskComposerInstall()->noSuggest()->run(); // composer install --no-suggest
```

Regarding tests, i found some related to composer tasks in https://github.com/consolidation/Robo/blob/master/tests/unit/Task/ComposerTest.php but i haven't been able to run them. I can add some if required but i would need some guidance on how to do it.

Thanks


greg-1-anderson:
LGTM. Thanks.

PR #946, opened 2020-05-12
Add a limit to replace via regexp

davidsneighbour:
# Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary

When using [Replace with Regexp](https://robo.li/tasks/File/#replace) there is no possibility to add a limit to the replaced items. This would be useful for instance when I replace version numbers in the DocBlock of fields and we want to replace only a single occurence.

for consolidation/Robo#945

### DIFF
--- a/src/Task/File/Replace.php
+++ b/src/Task/File/Replace.php
@@ -45,6 +45,11 @@ class Replace extends BaseTask
     protected $from;
 
     /**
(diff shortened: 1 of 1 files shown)


davidsneighbour:
Tests and Documentation changes (if any are required) are missing from this.

PR #954, opened 2020-06-15
Escape SSH commands correctly

prudloff-insite:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
This changes the way SSH commands are escaped in order to work with commands containing `'`.

### Description
Consider the following code:
```php
    $gitTask = $this->taskGitStack()
      ->commit('foo bar');

    $this->taskSshExec()
      ->exec($gitTask)
      ->run();
```

Without this patch, it will produce (which will fail):
```bash
ssh server 'cd "/path/" && git commit -m 'foo bar''
```

And with the patch:
```bash
ssh server 'cd "/path/" && git commit -m '\''foo bar'\'''
```
(Now the Git message is escaped correctly.)

Sorry, we are still stuck on Robo 1.4.12 but I would understand if you didn't want to merge a fix on this old branch.

### DIFF
--- a/src/Task/Remote/Ssh.php
+++ b/src/Task/Remote/Ssh.php
@@ -2,6 +2,7 @@
 
 namespace Robo\Task\Remote;
 
+use Robo\Common\ProcessUtils;
 use Robo\Contract\CommandInterface;
 use Robo\Exception\TaskException;
 use Robo\Task\BaseTask;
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This looks good, but it will need to go in on the default branch first. The code should be about the same there.

prudloff-insite:
There is a PR for the 3.x branch here: https://github.com/consolidation/robo/pull/989

prudloff-insite:
1.x is not supported anymore.
Replaced by https://github.com/consolidation/robo/pull/989.

PR #955, opened 2020-06-17
Fix #935: Upgrade to League Container 3.x

### LINKED ISSUE #935: league/container 2.x is unmaintained and throws warnings on composer install
### Steps to reproduce
Run `composer install` on any project using any version of Robo.

### Expected behavior
No warnings.

### Actual behavior
Deprecation warning:
> Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.

This warning occurs because Robo depends on `"league/container": "^2.4.1",`, and league/container 2.x versions depend on the abandoned container-interop.

Additionally, league/container 2.x is [explicitly no longer supported](https://container.thephpleague.com/) as of over a year ago.

Robo should upgrade league/container to 3.x in order to fix this.

danepowell:
### Overview
This pull request:

- [x] Fixes a bug
- [x] Adds a feature

### Summary
Fix #935, see description there.

League Container 2.x is unmaintained and throws warnings when users install Robo. Robo needs to upgrade to League Container 3.x.

### Description
This work is incomplete, but I thought I'd open a PR so others can build on it. Specifically:
- [ ] League Container changed a bunch of method names from 2.x->3.x, but must have changed something more significant as well because tests are failing with errors such as `No container implementation has been set.`. We need to figure out how to get tests passing and finish the upgrade
- [ ] This doesn't consider backwards compatibility. Because of the new method names, League Container 3.x is not BC with 2.x, so Robo will probably either need a new major branch to support it, or will need a shim/wrapper class/methods to support both 2.x and 3.x, assuming you are concerned about maintaining compatibility with downstream projects (like Drush) that still use League Container 2.x.


danepowell:
Converting to regular PR so tests run and folks can assist with debugging them...

joeparsons:
>  League Container changed a bunch of method names from 2.x->3.x, but must have changed something more significant as well because tests are failing with errors such as No container implementation has been set.. We need to figure out how to get tests passing and finish the upgrade

`No container implementation has been set` is an exception that [League\Container\ContainerAwareTrait::getContainer()](https://github.com/thephpleague/container/blob/3.3.1/src/ContainerAwareTrait.php#L34-L46) now throws if a container has not yet been set.  The [2.x version of this method](https://github.com/thephpleague/container/blob/2.4.1/src/ContainerAwareTrait.php#L25-L33) did not throw exceptions.

I verified locally that [these changes](https://github.com/danepowell/Robo/pull/1/files) to `Robo\Runner` resolve the `No container implementation has been set` errors although there are still other test failures.

greg-1-anderson:
We now have a 3.x-unstable Robo branch. It would be reasonable to bump Robo 3.x up to League/Container 3.x, and leave Robo 1.x and 2.x on the current version. Robo clients would have to "come along" to League/Container 3 to get Robo 3, then, but we wouldn't have to worry about making a shim for multiple League versions.

joeparsons:
I think this can be closed now that #985 is merged.

PR #960, opened 2020-07-23
Create a semver file if using the SemVer task for the first time

pixelbrackets:
### Overview
This pull request:

- [x] Fixes a bug

### Summary
Issue description: 959

Remove the check that a given filename is writeable, since this requires the file to exist beforehand.
    
We dont expect the file to exist however. Therefore this check should not be made, or create a file if failing.

Since this check is included in file_put_contents anyway we may remove the check safely.

I added a test for this issue as well.

### DIFF
--- a/src/Task/Development/SemVer.php
+++ b/src/Task/Development/SemVer.php
@@ -226,6 +226,7 @@ class SemVer implements TaskInterface
         if (empty($this->path)) {
             return true;
         }
+
         $semver = sprintf(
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Looks good, thanks.

And Coveralls, what's up with you? Coverage increased, and you still mark this as a fail? Really? Robots sometimes.

Thanks for adding a test.

Issue #961, opened 2020-07-29
Feature request: Add archive filtering


obriat:
For security reason I need to search and delete files from my final archive (e.g.: *.txt, *.md, *.json, node_module, ...).
Could filtering parameters be added to the taskPack::addDir taskPack::add methods or add a filtering method :
```php
<?php
$this->taskPack(<archiveFile>)
->addFile('dir/file.txt', 'file.txt')  
->filtering('*.txt', '*.md', 'toto/**/node_module')  
->run();`

I don't want to modify the source folder (I still have git actions to perform on it).
I tried to copy my origin folder but its to slow.

obriat:
I made a PR https://github.com/consolidation/Robo/pull/964, comments are welcome

PR #964, opened 2020-08-14
Add ignorelist to pack command.

obriat:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature
- [x ] Has tests that cover changes
- [ x] Adds or fixes documentation

### Summary
Related to #961, allows to filter patterns when creating an archive.

### Description
Files or folders could be excluded by using patterns, it's a good way to clean your package before delivery without touching you project folder tree.

```php
<?php
$this->taskPack(<archiveFile>)
->add('project') 
->exclude(['dir/*.zip', '*.md'])
->run();
```

Good to know :
Zip allows regexp, glob or string, @see https://symfony.com/doc/current/components/finder.html#file-name. 
Tar, gz and bz2 allow only regexp @see https://pear.php.net/package/Archive_Tar/docs/latest/Archive_Tar/Archive_Tar.html#methodsetIgnoreList.

### DIFF
--- a/src/Task/Archive/Pack.php
+++ b/src/Task/Archive/Pack.php
@@ -38,6 +38,12 @@ class Pack extends BaseTask implements PrintedInterface
     private $archiveFile;
 
     /**
+     * A list of patterns to exclude from the archive.
+     *
+     * @var array
+     */
+    private $ignoreList;
+    /**
      * Construct the class.
      *
      * @param string $archiveFile
@@ -48,6 +54,7 @@ class Pack extends BaseTask implements PrintedInterface
     public function __construct($archiveFile)
     {
         $this->archiveFile = $archiveFile;
+        $this->ignoreList = [];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Task/Archive/Pack.php
```diff
@@ -134,6 +141,27 @@ public function add($item)
...
+     *   @see \Archive_Tar::setIgnoreList
+     *
+     * @return $this
+     */
+    public function exclude($ignoreList)
+    {
+        if (!is_array($ignoreList)) {
+            $ignoreList = [$ignoreList];
+        }
+        $this->ignoreList = $ignoreList;
```
greg-1-anderson, line 161: Maybe array_merge here so folks can exclude multiple times.

tests/integration/PackExtractTest.php
```diff
@@ -30,7 +30,7 @@ public function archiveTypeProvider()
         if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
             return [['zip']];
         }
-        return [['zip'], ['tar'], ['tar.gz'], ['tar.bz2'], ['tgz']];
+        return [['tar'], ['tar'], ['tar.gz'], ['tar.bz2'], ['tgz']];
```
greg-1-anderson, line 33: Was this change in error?

obriat: oh crap :(

src/Task/Archive/Pack.php
obriat: good idea, quite obvious

tests/integration/PackExtractTest.php
obriat: Actually it's not just a typo: it raises an issue with the zip and "tar" exclude pattern formats. I had to made some more tests and maybe I'll have to add checks or conversions...

src/Task/Archive/Pack.php
obriat: Done

tests/integration/PackExtractTest.php
obriat: I fix the typo. I had to normalize the exclude pattern and narrow it to regex without enclosing slashes so it's compatible with both Symfony Finder and Archive_tar

obriat:
Ok, I fix the doc by adding the correct comments and running `robo docs`.
I rewrote the git history to have cleaner comments, but I now realise that's maybe not the correct way if you want to review my changes...

greg-1-anderson:
It's usually better to not rewrite history in a PR; I always squash-and-merge when done anyway. This PR is short enough that it wasn't an issue, though.

Thank you for contributing this.

PR #967, opened 2020-08-27
Removed grasmash/yaml-expander dependency

alsar:
### Overview
This pull request:

- [x] Fixes a bug

### Summary
Removed unused dependency `grasmash/yaml-expander`.


greg-1-anderson:
You shouldn't need a Symfony 5 scenario; if the dependencies are all resolving correctly, then the default scenario should be using Symfony 5.

greg-1-anderson:
I pushed a simpler version of this to the main branch.

PR #968, opened 2020-08-27
Add Runner function for retrieving an application to be used with tests

### LINKED ISSUE #947: Recommendations for automated testing of Robo Scripts/Commands?
Hey Greg,

What's the best way to go about testing custom Robo Scripts, or anything extending `\Robo\Tasks`?

For example, say I wanted to assert the `foo` method said `This is a Robo script, $name` in https://github.com/consolidation/Robo/blob/master/examples/robo.script ?

Or do any unit/functional tests exist that I can look at that test https://github.com/consolidation/Robo/blob/master/examples/src/Robo/Plugin/Commands/ExampleCommands.php ?

Thanks!

lpeabody:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Add a formal mechanism for testing commands, including those which are fully interactive.

### Description
Adds a mechanism to retrieve an application for testing purposes. In addition, adds a couple of integration tests that validates the results of the `try:input` and `try:exec` commands.

Fixes #947

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -12,6 +12,7 @@ use Robo\Exception\TaskExitException;
 use League\Container\ContainerAwareInterface;
 use League\Container\ContainerAwareTrait;
 use Consolidation\Config\Util\EnvConfig;
+use Symfony\Component\Console\Output\NullOutput;
 
 class Runner implements ContainerAwareInterface
 {
@@ -154,6 +155,41 @@ class Runner implements ContainerAwareInterface
         }
         $commandFiles = $this->getRoboFileCommands($output);
         return $this->run($argv, $output, $app, $commandFiles, $this->classLoader);
+    }
+
+    /**
+     * Return an initialized application loaded with specified commands and configuration.
+     *
+     * This should ONLY be used for testing purposes. Works well in conjunction with Symfony's CommandTester.
+     *
+     * @see https://symfony.com/doc/current/console.html#testing-commands
+     *
+     * @param string|null $appName
+     *   Name of the application.
+     * @param string|null $appVersion
+     *   Version of the application.
+     * @param string|null $commandFile
+     *   Name of the specific command file that should be included with the application.
+     * @param \Robo\Config\Config|null $config
+     *   Robo configuration to be used with the application.
+     * @param \Composer\Autoload\ClassLoader|null $classLoader
+     *   Class loader to use. If PHPUnit and the application runner rely on the same autoloader, this can be left null.
+     *
+     * @return \Robo\Application
+     *   Initialized application based on passed configuration and command classes.
+     */
+    public function getAppForTesting($appName = null, $appVersion = null, $commandFile = null, $config = null, $classLoader = null) {
+        $app = Robo::createDefaultApplication($appName, $appVersion);
+        $output = new NullOutput();
+        $commandFiles = $this->getRoboFileCommands($output); // $output is just used for printing error messages, it can be a throwaway stream
+        $container = Robo::createDefaultContainer(null, $output, $app, $config, $classLoader);
+        if (is_null($commandFile)) {
+            $this->registerCommandClasses($app, $commandFiles);
+        }
+        else {
+            $this->registerCommandClass($app, $commandFile);
+        }
+        return $app;
     }
 
     /**
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -168,8 +168,8 @@ class Runner implements ContainerAwareInterface
      *   Name of the application.
      * @param string|null $appVersion
      *   Version of the application.
-     * @param string|null $commandFile
-     *   Name of the specific command file that should be included with the application.
+     * @param string|array|null $commandFile
+     *   Name of the specific command file, or array of commands, that should be included with the application.
      * @param \Robo\Config\Config|null $config
      *   Robo configuration to be used with the application.
      * @param \Composer\Autoload\ClassLoader|null $classLoader
@@ -182,12 +182,12 @@ class Runner implements ContainerAwareInterface
     {
         $app = Robo::createDefaultApplication($appName, $appVersion);
         $output = new NullOutput();
-        $commandFiles = $this->getRoboFileCommands($output); // $output is just used for printing error messages, it can be a throwaway stream
         $container = Robo::createDefaultContainer(null, $output, $app, $config, $classLoader);
-        if (is_null($commandFile)) {
-            $this->registerCommandClasses($app, $commandFiles);
-        } else {
-            $this->registerCommandClass($app, $commandFile);
+        if (!is_null($commandFile) && (is_array($commandFile) || is_string($commandFile))) {
+            if (is_string($commandFile)) {
+                $commandFile = [$commandFile];
+            }
+            $this->registerCommandClasses($app, $commandFile);
         }
         return $app;
     }
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -3,6 +3,7 @@
 namespace Robo;
 
 use Composer\Autoload\ClassLoader;
+use Robo\Traits\CommandTesterTrait;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Input\StringInput;
 use Robo\Contract\BuilderAwareInterface;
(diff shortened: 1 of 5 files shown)


src/Runner.php
```diff
@@ -156,6 +157,41 @@ public function execute($argv, $appName = null, $appVersion = null, $output = nu
...
+     *   Class loader to use. If PHPUnit and the application runner rely on the same autoloader, this can be left null.
+     *
+     * @return \Robo\Application
+     *   Initialized application based on passed configuration and command classes.
+     */
+    public function getAppForTesting($appName = null, $appVersion = null, $commandFile = null, $config = null, $classLoader = null)
+    {
+        $app = Robo::createDefaultApplication($appName, $appVersion);
+        $output = new NullOutput();
+        $commandFiles = $this->getRoboFileCommands($output); // $output is just used for printing error messages, it can be a throwaway stream
```
greg-1-anderson, line 185: This only needs to be called if it is going to be registered.

lpeabody: Do you want to leave the responsibility of command registration out of this function altogether?

greg-1-anderson: I think a lot of tests will want to just register the default list of commands. There could be two versions, though, one that took a commandfile / commandfile list, and the other that called getRoboFileCommands and called through to the first.

lpeabody: Removed call to `Runner::getRoboFileCommands` and replaced it with the following check:

If the `$commandFile` parameter is a string or an array, then invoke `Runner::registerCommandClasses`. If `$commandFile` is a string, it gets converted to an array prior to being passed to `Runner::registerCommandClasses`.

lpeabody: Example usage https://bitbucket.org/lpeabody/release-manager/src/fdc3554f553ac75b7fa4fd9e434f2c8278d85e33/tests/CommandTesterTrait.php?at=feature%2Frobo-966-g1a-updates#lines-25:49

greg-1-anderson: Sure, that works; on reflection, I'm not sure that `Runner::getRoboFileCommands()` is often useful in testing, as it usually just passes through what was provided to the Runner constructor.

tests/integration/CommandTesterTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->assertContains("1234", $tryInputOutput);
+    }
+
+    public function testTesterWithOptions()
+    {
+        list($execOutput, $statusCode) = $this->executeCommand('try:exec', []);
+        $this->assertEquals(0, $statusCode);
+        list($execOutput, $statusCode) = $this->executeCommand('try:exec', [], ['--error']);
+        $this->assertNotEquals(0, $statusCode);
+    }
```
lpeabody, line 32: One thing of note, and I don't quite understand it, but this test generates the following output:

```
lpeabody@pop-os:~/Projects/Robo (feature/947-part-deux-build-test-application)$ fin exec /var/www/vendor/bin/phpunit tests/integration/CommandTesterTest.php
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.19
Configuration: /var/www/phpunit.xml.dist
Error:         No code coverage driver is available

.hey
hey
hey
hey
hoy
hoy
hoy
gou
gou
die
.                                                                                                                                                                    2 / 2 (100%)

Time: 10.23 seconds, Memory: 8.00MB

OK (2 tests, 6 assertions)
```

I assume this is because there is some disconnect between the output stream injected into the command class versus the output stream available to each ExecTask being executed?

lpeabody:
@greg-1-anderson this is ready for review/scrutiny now.

tests/integration/CommandTesterTest.php
greg-1-anderson: That output, `hey`, `hoy`, etc. is what one of the test fixture scripts print out. It seems that the fixture is printing its output to stdout instead of having it captured.

I don't see that when I run `composer unit tests/integration/CommandTesterTest.php`; maybe it has something to do with how you're running the tests? Does `composer unit` work for you?

greg-1-anderson:
Regarding the failed test, it looks like perhaps this technique does not work well in Symfony 4. It would be fine if you wanted to just detect Symfony 4 and mark the test skipped. I'm not sure about the detection part; usually such breakage would be caused by a missing class not yet available that you could just test for with `class_exists()`. In this case, though, it looks like the code is running, it's just not supplying the stdin stream to the test command.

src/Runner.php
```diff
@@ -156,6 +157,43 @@ public function execute($argv, $appName = null, $appVersion = null, $output = nu
...
+     * @param string|null $appName
+     *   Name of the application.
+     * @param string|null $appVersion
+     *   Version of the application.
+     * @param string|array|null $commandFile
+     *   Name of the specific command file, or array of commands, that should be included with the application.
+     * @param \Robo\Config\Config|null $config
+     *   Robo configuration to be used with the application.
+     * @param \Composer\Autoload\ClassLoader|null $classLoader
+     *   Class loader to use. If PHPUnit and the application runner rely on the same autoloader, this can be left null.
```
greg-1-anderson, line 178: The class loader is only needed if testing Robo plugins. I think this is only rarely going to be needed, but might as well leave it in for those fringe cases.

tests/src/Traits/CommandTesterTrait.php
```diff
@@ -0,0 +1,58 @@
...
+     * @param array $command_extra
+     * @param string|array|null $commandClasses
+     * @return array
+     */
+    protected function executeCommand($command_string, $inputs = [], $command_extra = [], $commandClasses = null)
+    {
+        if (is_null($commandClasses)) {
+            $commandClasses = $this->commandClasses;
+        }
+        $app = $this->runner->getAppForTesting($this->appName, $this->appVersion, $commandClasses);
```
greg-1-anderson, line 50: I think you could get rid of the if above and use `$commandClasses ?? $this->commandClasses` here. I think ?? was added in 7.0.  We only need 7.1+ here.

src/Runner.php
greg-1-anderson: The comment is not correct, though; it doesn't matter if PHPUnit and the application have the same or different autoloader, the plugin tests won't work unless this reference is provided, and the classloader isn't needed unless loading plugins.

greg-1-anderson:
Hm, locally I ran:
```
$ composer scenario symfony4
$ composer test
```
Everything passed.

greg-1-anderson:
On Travis, there is also this message:
```
stty: standard input: Inappropriate ioctl for device
stty: standard input: Inappropriate ioctl for device
```
Perhaps there's some tty issue on some platforms with Symfony 4 that is fixed in Symfony 5?

Just adding a markTestSkipped to `testInputApis` iff the Symfony version is 4 is fine, though. It's not necessary to figure out if the tty configuration on the local platform will sustain the test. Just add a comment explaining.

lpeabody:
Thanks for the review @greg-1-anderson, I'll tend to it.

tests/integration/CommandTesterTest.php
lpeabody: I ran the test by explicitly calling `/var/www/vendor/bin/phpunit tests/integration/CommandTesterTest.php`. However, if I run `composer run unit` I do not see the above mentioned output.

lpeabody:
I think I've addressed all your feedback. Looks like we're in the green!

tests/integration/CommandTesterTest.php
```diff
@@ -16,6 +16,9 @@ public function setUp()
 
     public function testInputApis()
     {
+        if (getenv('SCENARIO') == 'symfony4' && getenv('DEPENDENCIES') == 'lowest') {
+            $this->markTestSkipped('There is a bug with a lower dependency of symfony4 in how it handles tty.');
```
greg-1-anderson, line 20: Oh if this is only happening in LOWEST then we can just bump the minimum version of Symfony 4 in composer.json and we're all set. It's fine to just use what's in composer.lock for the symfony4 scenario as the lowest viable version. Ideally just bump the dependencies that need it.

`gtenv('SCENARIO')` isn't a reliable way to detect Symfony 4 anyway.

greg-1-anderson:
I'm just going to merge this; I'll deal with fixing the LOWEST pins and removing the markTestSkipped later. Follow-up PR would be appreciated if you feel like it though.

greg-1-anderson:
Thanks for all of your help here.

lpeabody:
Thanks for your help as well!

PR #970, opened 2020-08-29
Support saving and restoring state.

greg-1-anderson:
#966 / #968 were actually fairly dangerous, as they could corrupt the current commandfile instance state. This is an attempt to rectify that.

@lpeabody : Do you have any idea why CommandTesterTest takes 10s to run? None of the other tests take so long. It seems that all of the time is in the `execute` method. Have not traced it beyond that.


lpeabody:
@greg-1-anderson Ah, yeah that's annoying, it didn't occur to me how obnoxious that would become. It's because the PHPUnit test I added for testing the trait is running `try:exec` twice...

lpeabody:
@greg-1-anderson opened https://github.com/consolidation/Robo/pull/971 to address the long-running test annoyance.

PR #971, opened 2020-08-29
Don't use try:exec for testing the CommandTester trait.

lpeabody:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Removes test which greatly reduces time of test completion, and it doesn't really add value.


lpeabody:
@greg-1-anderson ^

lpeabody:
Back to 4.45 seconds https://travis-ci.org/github/consolidation/Robo/jobs/722362391#L893

greg-1-anderson:
OK, thanks. I wasn't so concerned about the total test run time, I was mostly concerned about whether there might be some bottleneck that was slowing down startup that might affect runtime code.

Thanks for explaining and cleaning up.

PR #972, opened 2020-08-31
Prefer passing $io parameter over using cached objects in class fields.

greg-1-anderson:
### Overview
This pull request:

- [x] Improves code design
- [ ] Fixes a bug
- [ ] Adds a feature

### Summary
Prefer passing $io parameter to Robo commands instead of using cached $input, $output and $logger objects from the DI container.

### Description
In #970, it became clear that the practice of caching the OutputInterface (including doing so indirectly, e.g. via a logger) makes it difficult to write testable code. Saving / restoring state adds complexity, and does not solve the underlying problem.

It has been possible for some time for Robo commands to declare that they want an $io object passed to them as a parameter. This means that commands would use `$io->writeln()` in place of `$this->say()`, which is preferable insofar as the later relies on the cached `$output` object injected from the DI container, whereas the former is composed from whatever `$input` and `$output` objects were provided when the Symfony API method (e.g. `Command::run()`) was called. These instances might be different than what is in the DI container when Symfony APIs are used directly, e.g. during tests, or when one command calls another, and other similar situations.

This PR makes the `$io` parameter the preferred way to declare Robo commands, and deprecates methods like `say()` and `yell()`. A new class, `ConsoleIO` is provided to make it more convenient for Robo commands to have access to both the SymfonyStyle I/O methods, and also the InputInterface that contains the commandline arguments and options. Using named arguments in the Robo command method to declare arguments and options is still supported.

Note that creating tasks without a builder is now deprecated, and the `$io` object needs to be passed to the collection builder so that tasks are updated with the right output stream (for printTaskInfo et. al.), so `$this->taskFoo()` will instead need to become `$this->collectionBuilder($io)->taskFoo()`.

Things we could still consider doing in this PR:
- [x] Add `say` and `yell` and c. to ConsoleIO, so calls to `$this->say()` could become `$io->say()`.
- [x] Update documentation to use preferred `$io` technique.
- [x] Update tests to use preferred `$io` technique.

Follow-on work to consider later:
- [ ] Consider using something shorter / more readable than `$this->collectionBuilder($io)`
- [ ] Stop using deprecated I/O objects from the DI container entirely in Robo internal code. Do not inject $input / $output / $logger etc. into DI container in tests, only include them at runtime (for backwards compatibility).
- [ ] Convert TaskIO to use SymfonyStyle instead of logger for output (breaking?).
- [ ] Remove deprecated objects from DI container at runtime (breaking).
- [ ] Demonstrate one command calling another. This is generally discouraged in the Symfony documentation, but ideally should work; presently, cached objects from DI container make this not work well.


greg-1-anderson:
Adding OutputAwareTrait to TaskIO is a breaking change for any Task that added OutputAwareTrait itself. This is perhaps unusual.

greg-1-anderson:
This is a big enough change that I am going to make a 3.x version of Robo for it. Robo 3 will be largely compatible with Robo 2 (similar to the 1 -> 2 transition); the input / output / logger objects will remain in the DI container, but deprecated.

Very few changes will be required; the large changes will all be encouraged.

greg-1-anderson:
I changed the default branch from `main` to `2.x` so that future-unstable can be done on the 3.x branch.

greg-1-anderson:
I made a couple more minor breaking changes:

- The typehints in InputAwareTrait were made more flexible.
- initTestTasksTrait can no longer be provided a container or $input object.

This should not affect many Robo clients.

greg-1-anderson:
Another breaking change:

- Rename `loadTasks` to `Tasks` everywhere it appears.
- Rename `loadShortcuts` to `Shortcuts` everywhere it appears.

These changes should not affect most Robo clients, as most should just be extending \Robo\Tasks, which uses LoadAllTasks, which contains the `loadTasks` / `Tasks` references.

Issue #936, opened 2020-04-01
Github access_token via query parameter is deprecated


danepowell:
### Steps to reproduce
Created a new Github release using [GithubRelease.php](https://github.com/consolidation/Robo/blob/master/src/Task/Development/GitHubRelease.php)

### Expected behavior
Release should be created without deprecation warnings from Github.

### Actual behavior
Moments later I received an email from Github:
> your personal access token (blt release generator) using Robo was used as part of a query parameter to access an endpoint through the GitHub API... Please use the Authorization HTTP header instead, as using the `access_token` query parameter is deprecated

### Details
Instead of rolling its own Github API client, I might suggest that Robo switch to using a client such as this one: https://github.com/KnpLabs/php-github-api

In that case you can just specify `AUTH_HTTP_TOKEN` and everything works fine.

Alternatively, Robo needs to send the token via HTTP Basic Auth rather than as a query parameter in [Github::sendRequest()](https://github.com/consolidation/Robo/blob/master/src/Task/Development/GitHub.php#L120)

Issue #363, opened 2016-06-17
When running a watch task Class not found thrown


t73biz:
When installing robo via composer and installing henrikbjorn/lurker to implement a taskWatch, I recieve the following error.

```
[Watch]  Class ResourceWatcher not found. Please install henrikbjorn/lurker Composer package 
[Watch]  Exit code 1 
```
Looking further into the issue, there is a Namespace mismatch with Lurker\ResourceWatcher and Robo\Task\Base\Watch

daaru00:
I still having this issue using `1.2.1` Robo version as framework, it ask me to require `henrikbjorn/lurker` but there are a Symfony version missmatch:
`henrikbjorn/lurker` does not support Symfony 4 (it require `^2.2|^3.0`), I opened an issue about this: https://github.com/flint/Lurker/issues/30

Anyway watch task cannot be used, I can't install it's requirements.

greg-1-anderson:
See https://github.com/flint/Lurker/pull/29.

In the interim, if you do not strongly require Symfony 4, try pinning to Symfony 3.4 in your application, and see if that allows you to use taskWatch again.

daaru00:
With Symfony 3.4 it continues to not work, `henrikbjorn/lurker` require some Symfony component at 3.0 https://github.com/flint/Lurker/blob/master/composer.json#L25

I fix it requiring in my composer.json:
```json
"require": {
    "symfony/config": "^3.0",
    "symfony/event-dispatcher" : "^3.0"
}
```

Anyway it would be better to make it compatible with Symfony 4, using Robo as a framework I might need other modules that instead require it.

greg-1-anderson:
For Symfony 4 support, please ask at https://github.com/flint/Lurker/pull/29

pretzlaw:
Leaving a small workaround here for all those who are stuck right now and need a working solution immediately:
https://gist.github.com/pretzlaw/f11c26076688c650b05bee279e696669

Does a `class_alias` of `\Lurker\ResourceWatcher` towards a proxy to the "yosymfony/resource-watcher" package.

PR #973, opened 2020-09-01
Use lurkerlite to support watch task in robo.phar (2.x) 

totten:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary

Swap [henrikbjorn/lurker](https://github.com/flint/Lurker/)  with the fork  [totten/lurkerlite](https://github.com/totten/Lurker).

### Description

[henrikbjorn/lurker](https://github.com/flint/Lurker/) is required for `taskWatch()`. The package appears to be unmaintained, and dependency-management issues prevent it from being bundled into `robo.phar`. Consequently, `robo.phar` cannot support `taskWatch()`.

This patch swaps in a fork, [totten/lurkerlite](https://github.com/totten/Lurker), with a smaller dependency-graph - and bundles that into `robo.phar`. So `taskWatch()` works by default.

Under the hood, `lurkerlite` removes the dependencies on `symfony/config` and `symfony/event-dispatcher`. Therefore, it can coexist with Symfony v2, v3, v4, v5, v6, v7, v-8, and v-7.

### Comments

See also: #936, #363

I'm not planning any heavy development on `lurkerlite` -- I just needed an installable copy, and I figured I'd post it back to `Robo.git` since I first stumbled upon `lurker` while browsing here. Hopefully, having fewer dependencies and a passing test-suite will keep the administrative low.

### DIFF
--- a/.scenarios.lock/symfony4/composer.json
+++ b/.scenarios.lock/symfony4/composer.json
@@ -36,7 +36,8 @@
         "consolidation/annotated-command": "^4.2.1",
         "consolidation/output-formatters": "^4.1.1",
         "consolidation/self-update": "^1.2",
-        "league/container": "^2.4.1"
+        "league/container": "^2.4.1",
+        "totten/lurkerlite": "^1.3"
     },
     "require-dev": {
         "g1a/composer-test-scenarios": "^3",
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
This is great, thanks.

lurkerlite should be listed in `suggest`, though, not `require`.

Optionally, it would be good if there were a test. Currently, none of the tests in `tests/unit` are running, because I could not support Codeception tests. Tests are being ported to `tests/integration`. I'll accept this PR without tests if you don't have time, but adding a simple test would help ensure that the watch task continues to work with Robo.

The test would only really serve this purpose, though, if it actually called the lurkerlite library. The current test mocks away the library. A functional test would have to make use of [threads](https://www.php.net/manual/en/class.thread.php) or something similar so that the main test thread could modify a file and then observe whether the watcher reacted appropriately. If you don't want to get into this, I understand; no one got around to it so far with the original lurker library.

If you do write a test, then lurkerlite should be listed in both `suggest` and `require-dev`.

PR #975, opened 2020-09-04
Do not allow tests to pass a container in to initTestTasksTrait.

greg-1-anderson:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Clean up initTestTasksTrait. This PR breaks b/c only with the unstable 2.x branch; we'd rather not allow tests to specify the container or $input object that should go into the container.

@lpeabody Robo core does not need either of those items injected. Can you fix up your tests to not depend on passing the container or input to initTestTasksTrait?

### DIFF
--- a/tests/src/Traits/TestTasksTrait.php
+++ b/tests/src/Traits/TestTasksTrait.php
@@ -15,20 +15,13 @@ trait TestTasksTrait
     protected $capturedOutput;
     protected $logger;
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Oh wait I thought this was part of your command tester PR, but I see from git blame that this was me, back in February.

nevermind, I'll just make this part of the Robo 3.x in #972.

PR #979, opened 2020-09-08
Fixes #978: Do not call ignorelist on the archive if it is empty.

### LINKED ISSUE #978: Tar: setIgnoreList causes empty archives
Our builds started failing because the archive we create & upload to our environments were empty. After a lot of debugging I found that https://github.com/consolidation/Robo/pull/964 caused the issue. 

https://github.com/consolidation/Robo/pull/964/files#diff-b826726cd9ecd307766875aaa192e6c5R58 sets the ignoreList to an empty array. Then it calls the `setIgnoreList` method on the tar object: https://github.com/consolidation/Robo/pull/964/files#diff-b826726cd9ecd307766875aaa192e6c5R209

Which causes the ignore regexp to result in `#/#` in https://github.com/pear/Archive_Tar/blob/master/Archive/Tar.php#L732

This regexp matches all our files, resulting in empty archives :(

greg-1-anderson:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Apparently archives with no ignore list are now empty. I guess that all the tests set an ignore list, but have not fully investigated in order to get a quick release out.

### DIFF
--- a/src/Task/Archive/Pack.php
+++ b/src/Task/Archive/Pack.php
@@ -206,7 +206,9 @@ class Pack extends BaseTask implements PrintedInterface
         }
 
         $tar_object = new \Archive_Tar($archiveFile);
-        $tar_object->setIgnoreList($this->ignoreList);
+        if (!empty($this->ignoreList)) {
+            $tar_object->setIgnoreList($this->ignoreList);
+        }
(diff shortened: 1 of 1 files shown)


Jelle-S:
Yep, that seems to do the trick

PR #985, opened 2020-09-14
Fix #935: Upgrade to League Container 3.x. (Robo 3.x version)

### LINKED ISSUE #935: league/container 2.x is unmaintained and throws warnings on composer install

joeparsons:
### Overview
This pull request:

- [x] Fixes a bug
- [x] Adds a feature

### Summary
League Container 2.x is unmaintained and throws warnings when users install Robo. Robo needs to upgrade to League Container 3.x.

This includes @danepowell's changes from #955 (modified so they apply cleanly to the Robo 3.x branch) and an additional change to `Robo\Runner` to handle the exceptions now thrown by [League\Container\ContainerAwareTrait::getContainer()](https://github.com/thephpleague/container/blob/3.3.1/src/ContainerAwareTrait.php#L34-L46)

### Description
3.x tests pass locally for me with these changes.


src/Collection/CollectionBuilder.php
```diff
@@ -83,7 +83,7 @@ public function __construct($commandFile)
     }
 
     /**
-     * @param \League\Container\ContainerInterface $container
+     * @param \League\Container\Container $container
```
greg-1-anderson: Shouldn't all of these references to ContainerInterface become Psr\Container\ContainerInterface?

joeparsons: I was thinking that and even made that change in a local branch based on 2.x at one point but wasn't totally sure since Robo uses some `League\Container` specific features.  I'll try making that change to this branch now.

src/Robo.php
```diff
@@ -139,7 +139,7 @@ public static function loadConfiguration($paths, $config = null)
      * @param null|\Consolidation\Config\ConfigInterface $config
      * @param null|\Composer\Autoload\ClassLoader $classLoader
      *
-     * @return \Psr\Container\ContainerInterface
+     * @return \League\Container\Container|\Psr\Container\ContainerInterface
```
greg-1-anderson: \League\Container\Container **isa** \Psr\Container\ContainerInterface, so it seems unnecessary to explicitly list it as a return type.

greg-1-anderson:
LGTM, thanks for the upgrade.

PR #987, opened 2020-09-30
Avoid public methods from Robo\Common\IO to be interpreted as commands.

rodrigoaguilera:
### Overview
This pull request:

- [x] Fixes a bug

### Summary
There is two new commands in the list when running `robo list`: `current:state` and `restore:state`.
Running them doesn't yield any output.

### Description
Digging into the code I discovered that they come from `Robo\Common\IO` as they are used as a trait in `Robo\Tasks`.
This change was introduced in the following commit:
https://github.com/consolidation/Robo/commit/cf58ea643e77e80f79f3b1993b071d0a52202d5f
So is included in 2.2.1 and 2.2.0.

This solution is a workaround to be able to fix this ASAP in stable branches by defining those methods as a hook that alters a high unlikely command name. 

I think the proper solution is to rename those two methods to `getCurrentState` and `setRestoredState` in the annotated-command package but since is an interface change that would require a major version release.
Additionally, handling an `@ignore-command` annotation can be quite handy for this kind of situations when we want class/trait public methods to have a good name that is not interpreted as a robo command and is not prefixed by get/set.

Do you want me to open the PRs in the annotated-command package?

For the test I used the ApplicationTest from the now-disabled codeception suite, so it can be a base to transform the tests from that class into phpunit.

What do you think?


greg-1-anderson:
I don't like the `@hook` workaround.

This functionality is so new, it is unlikely that many external clients refer to these methods. I am inclined to call them "experimental" in the versions they appear, and fix the name as an exceptional b/c breaking bugfix.

While these features were not advertised as experimental, neither have they been publicized (blog, slack, twitter), so this is perhaps okay.

The alternative would be to enhance the capabilities of AnnotatedCommand and allow Robo to declare currentState / restoreState is ignored method names, just like get*.  This is a more robust solution, but also more work.

If you'd like to make a PR for one of these fixes in annotated-commands, I'd appreciate it.

rodrigoaguilera:
The latest commit should fail the tests as it requires
https://github.com/consolidation/annotated-command/pull/211
but illustrates what I am trying to accomplish. I think is a nice feature to have a @ignored-command annotation and I can leverage the the dynamic annotations to be able to ignore certain methods from Robo.

greg-1-anderson:
Released consolidation/annotated-command 4.2.2 with @ignored-command annotation.

src/AnnotatedCommand/CommandInfoAlterer.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
greg-1-anderson: This isn't how I would inject ignored method names into Annotated Command. Instead, I'd make a field of ignored names or ignored regex's in the AnnotatedCommandFactory, add a setter / appender for that, and call it in the DI constructor.

Since the currentState and restoreState methods are part of a Robo trait, though, you can just annotate them with @ignored-command and remove this.

If you don't want to use the annotation, then please make a PR in annotated-commands to add an ignored name / ignored regex facility to AnnotatedCommandFactory.

greg-1-anderson (changes requested):
Let's use an alternate solution than alterCommandInfo.

src/AnnotatedCommand/CommandInfoAlterer.php
rodrigoaguilera: Thank you for the input.
I'll bring this conversation to the annotated-command repo with a PR, just a few notes here so I write better code.

I will go with the ignored regex facility but for that I also need to convert to non-static the `isCommandMethod` in the factory.
I am not familiar with configuring the container but I'll follow the examples I see around these lines:
https://github.com/consolidation/Robo/blob/3.x/src/Robo.php#L324

Am I correct?

greg-1-anderson:
I forgot that isCommandMethod is static. It need not have been. It also need not have been public.

I guess you can leave the existing isCommandMethod public / static and mark it deprecated, and then make a new protected non-static method with a similar name.

rodrigoaguilera:
Waiting for a release that includes 
https://github.com/consolidation/annotated-command/pull/212
before continuing work on this.

greg-1-anderson:
It's possible to work off of a branch of a dependency, but in this case doing so would have required temporarily adding an entry for your fork of annotated-command to the repositories section.

In any event, I released annotated command 4.2.3 with your contribution this morning, so you don't need to do that.

rodrigoaguilera:
mmm :thinking: the update to pear/archive_tar to 1.4.10 seems to be breaking the php 7.1 tests

rodrigoaguilera:
Definitely this change
https://github.com/pear/Archive_Tar/commit/8407ee4e572d3580d2f80f09c122a6ee7a84d6e0

rodrigoaguilera:
It was not related with php 7.1. I guess the other builds use a cached version of pear/archive_tar.
I hope you don't mind changing that part of the test in this PR.
By looking at the docblock it seems that function was never meant to accept a regex, or at least a regex with the special characters `'#', '.', '^', '$'`.
https://github.com/pear/Archive_Tar/blob/8dd8b31cbbadd11bf88157f415cd5dc08b063043/Archive/Tar.php#L728

I don't know what the coveralls checks are complaining about. Anything I can do to help with that?

greg-1-anderson:
Yeah, that looks like a breaking change in pear/archive_tar, but it was a bug fix, and in any event isn't our fault. Thanks for fixing up the test failure.

Coveralls is perpetually reporting a failure now because our coverage is so low. If you look at the report, you will see that coverage for Robo core, is really high; however, Robo came with a lot of example tasks with no tests. I also made the coverage a bit lower still when I had to drop support for testing Robo with Codeception, and did not port all of the tests.

If you would like to submit some PRs to increase coverage, it would be appreciated; however, that's definitely out of scope for this PR.

PR #989, opened 2020-10-05
Escape SSH commands correctly

prudloff-insite:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Same as https://github.com/consolidation/Robo/pull/954 but for 3.0.

### Description
The fix is exactly the same.


prudloff-insite:
3.x is not maintained anymore so now targeting 5.x.
We have a few projects using Robo 5 with this patch and it seems to work correctly.

What would be needed to have this committed?

PR #994, opened 2020-10-16
New Logfile tasks for Robo

christingruber:
### Overview

This pull request:
- [ ] Fixes a bug
- [x] Adds a feature

### Summary
New Logfile tasks for Robo

### Description
Adds new task group Logfile with two tasks:
* taskRotateLog() - Will rotate logfiles
* taskTruncateLog() - Will truncate or create an empty logfiles

### DIFF
--- a/src/Task/Logfile/RotateLog.php
+++ b/src/Task/Logfile/RotateLog.php
@@ -10,9 +10,9 @@ use Robo\Task\Logfile\Exception\InvalidArgumentException;
  *
  * ``` php
  * <?php
- * $this->taskRotateLog('logfile.log')->run();
+ * $this->taskRotateLog(['logfile.log'])->run();
  * // or use shortcut
- * $this->_rotateLog('logfile.log');
+ * $this->_rotateLog(['logfile.log']);
  *
  * ?>
  * ```
@@ -35,8 +35,6 @@ class RotateLog extends BaseLogfile
     public function __construct($logfiles)
     {
         parent::__construct($logfiles);
-        // @todo allow multiple logfile rotations
-        $this->logfile = new \SplFileInfo(reset($this->logfiles));
     }
 
     /**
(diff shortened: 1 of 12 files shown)


greg-1-anderson:
This looks pretty good; thanks for the contribution. Would take this with some tests. It's not necessary to convince coveralls that the coverage increased, but some coverage of the basic functionality is needed.

src/Task/Logfile/RotateLog.php
```diff
@@ -40,12 +39,12 @@ public function __construct($logfiles)
     /**
      * @param int $keep
      * @return RotateLog
-     * @throws InvalidArgumentException
+     * @throws \Exception
      */
     public function keep(int $keep): self
     {
         if ($keep < 1) {
-            throw new InvalidArgumentException(
```
greg-1-anderson: What was the adverse effect of throwing a more specific exception?

christingruber: The side effect was that the `robo docs` command generates sub folders in the auto generated docs for the sub folders in the task dir `Task\Logfile`. In my opinion we can hold the custom exception, it is more clean and we can better handle the exception.

What do you think?

greg-1-anderson: More specific exceptions are preferable to more generic ones. I am generally not in favor of creating custom exceptions unless there is a specific use-case where the client is going to want to catch it explicitly, but in cases where there is an existing exception that is suitable, it is helpful to use it.

Perhaps the docs tool was generating extraneous subfolders because of a typo, `InvalidArgumentException` instead of `\ InvalidArgumentException`.

christingruber: You're right. I've decided to change the global `Exception` to a global `InvalidArgumentException`. A custom exception isn't really needed here. But feel free to change this in future, for me both are OK.

greg-1-anderson:
LGTM thanks for the contribution.

christingruber:
Thanks. Is there a second pull request needed for version 2.x or is it done by you?

PR #1001, opened 2020-11-16
Add a way to access the original message in TaskExitException

prudloff-insite:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Add a new `TaskExitException::getOriginalMessage()` method that returns the original error message returned by a task.

### Description
`TaskExitException` prepends the class name to the message returned by the task.
However, it is sometimes useful to access the original message in a catch clause.

Our use case is that we have a CLI tool that returns a JSON error when failing and we want to parse it.

(Sorry, we are still stuck on 1.x because of https://github.com/boedah/robo-drush/issues/18. But the same patch could probably be applied to 2.x.)

### DIFF
--- a/src/Exception/TaskExitException.php
+++ b/src/Exception/TaskExitException.php
@@ -30,7 +30,8 @@ class TaskExitException extends \Exception
     /**
      * @return string
      */
-    public function getOriginalMessage() {
+    public function getOriginalMessage()
+    {
         return $this->originalMessage;
     }
 }
--- a/src/Exception/TaskExitException.php
+++ b/src/Exception/TaskExitException.php
@@ -6,6 +6,11 @@ class TaskExitException extends \Exception
 {
 
     /**
+     * @var string
+     */
+    private $originalMessage;
+
+    /**
      * TaskExitException constructor.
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I'm -1 on the general concept of json-parsing exception messages, and don't think that it is a good idea to duplicate the message in the exception simply to support this use-case.

If you wanted to add some feature that would make the results from an exec process more readily available to a rollback method, e.g. so that you could do your json parsing there, then I would accept a PR to enable that.

prudloff-insite:
I tried using a rollback but could not find how to access the task output.

I ended up doing something basic like this:
```php
    $this->stopOnFail(FALSE);

    $result = $this->taskSshExec()
      ->exec('foo')
      ->run();

    if (in_array($result->getExitCode(), [0, 3])) {
      $this->stopOnFail();

      return json_decode($result->getMessage());
    } else {
      throw new TaskExitException($drushTask, $result->getMessage(), $result->getExitCode());
    }
```

But switching `stopOnFail` for this does not feel very clean.

Issue #1009, opened 2020-12-18
Yaml config loader missing in 2.x


pixelbrackets:
### Steps to reproduce

- Load a configuration YAML file with `Robo::loadConfiguration()`

### Expected behavior

- Parse the YAML file and add the configuration to the Robo configuration scope

### Actual behavior

- Fatal Error `PHP Fatal error:  Uncaught Error: Class 'Symfony\Component\Yaml\Yaml' not found in /home/xxx/vendor/consolidation/config/src/Loader/YamlConfigLoader.php:23`

### System Configuration

Ubuntu 20.4, PHP 7.4

### Additional notes

YAML configuration files are documented in the `getting started` guide - https://robo.li/getting-started/#configuration

It therefore should work automatically to load YAML files. Or at least throw the usual „Please install package X/Y“ warning.

pixelbrackets:
Ah, reason is the remove of package "grasmash/yaml-expander"  in commit 91b9343db9ccb739f067870eb5904807e11c2be5.
This package used to require "symfony/yaml".

The removal of the package was a result of the PR discussion #952. Which aimed at support of Symfony 5 originally.

Suggested solution: 
- require "symfony/yaml" with dependency "^4.0 || ^5.0 ".

This line is added in 3.x branch already, for another reason though (#1005, support PHP 8).

Why was this not detected by a test? There is a test for the configuration loader, which uses the symfony YAML parser. Well, the development dependency "php-coveralls/php-coveralls" requires "symfony/yaml". Thats why the test stays green in development and the exception is not detected until running in production.

pixelbrackets:
merged into 2.x, released in 2.2.2

PR #1010, opened 2020-12-18
[BUGFIX] Restore symfony/yaml dependency

pixelbrackets:
### Overview
This pull request:

- [x] Fixes a bug (Fatal error)
- [ ] Adds a feature

### Summary

See issue #1009 

Robos configuration loader uses the symfony/yaml package 
to parse YAML configuration files. This dependency was removed unintentionally
when package "grasmash/yaml-expander" was removed in commit 91b9343db9ccb739f067870eb5904807e11c2be5. 
The tests stayed green because another development package still requires the yaml parser.

Restore the dependency to avoid fatal errors in production.

Support version 4 and 5 of the package, as described in the robo branch overview.

### Description

Without the package dependency Robo 2.x currently throws fatal errors whenever using YAML configuration files in production.

The fix was added in 3.x branch already, for another reason though (#1005, support PHP 8).

### DIFF
--- a/.scenarios.lock/symfony4/composer.json
+++ b/.scenarios.lock/symfony4/composer.json
@@ -31,12 +31,13 @@
         "phpunit/phpunit": "^6",
         "nikic/php-parser": "^2",
         "php": ">=7.1.3",
+        "consolidation/annotated-command": "^4.2.1",
         "consolidation/config": "^1.2.1|^2",
         "consolidation/log": "^1.1.1|^2.0.1",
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
It was essentially for the same reason. Robo used to get its symfony/yaml dependency from one of its dependencies, but things changed. It should have always stipulated that symfony/yaml as a top-level dependency since it is only suggested in consolidation/config.

Thanks for the update.

pixelbrackets:
Thank you for the quick release of a new version :) :100: :100:

PR #1011, opened 2021-01-03
[FEATURE] Allow phar path in --dir

saitho:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Allows setting phar paths (e.g. `phar://file.phar`) as --dir option.

### Description
I'm building an own phar file that contains my RoboFile as well as the Robo dependencies.
I'd like to be able to load the RoboFile from my phar file. Right now I overwrite the method in Runner.

### DIFF
--- a/src/Runner.php
+++ b/src/Runner.php
@@ -524,13 +524,16 @@ class Runner implements ContainerAwareInterface
                 $this->roboClass = $className;
             }
         }
-        // Convert directory to a real path, but only if the
-        // path exists. We do not want to lose the original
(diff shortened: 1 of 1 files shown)


saitho:
Closing as this approach still doesn't work for me, as I can't get the arguments into the `Runner->execute` method.
Including the RoboFile and using `run` method is sufficient to achieve the goal.

```php
require_once 'RoboFile.php';
$statusCode = $runner->run($argv, null, null, 'RoboFile', $classLoader);
```

PR #1014, opened 2021-01-18
Add --no-exit option

ccsuperstar:
Add --no-exit option for 1.x version (to work with robo-paracept 0.4)

### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Add --no-exit option for 1.x version (to work with robo-paracept 0.4)

### Description
Add --no-exit option for 1.x version (to work with robo-paracept 0.4)

### DIFF
--- a/src/Task/Testing/Codecept.php
+++ b/src/Task/Testing/Codecept.php
@@ -243,6 +243,15 @@ class Codecept extends BaseTask implements CommandInterface, PrintedInterface
     }
 
     /**
+     * @return $this
+     */
+    public function noExit()
+    {
+        $this->option("no-exit");
+        return $this;
+    }
+    
+    /**
      * @param string $failGroup
      * @return $this
      */


greg-1-anderson:
If we're going to backport, we should do the 2.x branch too

ccsuperstar:
This is for use with robe-paracept
This project is no longer maintained and uses the consolidation/robo version 1.0.0. 
https://github.com/Codeception/robo-paracept/blob/master/composer.json

greg-1-anderson:
I'm happy to accept a backport of the `--no-exit` PR from Robo 3.x on the Robo 1.x branch, but only if we also have the same feature on the 2.x branch. There are likely other clients of the codeception task, and it wouldn't be good to have a discontinuity.

ccsuperstar:
Thanks a lot @greg-1-anderson!
In robo-paracept, the version of consolidation/Robo is 1.0.0

How can I benefit from this update without having to fork the project (which is no longer updated since 2019)?
Do you need to make a new release?

Thank you in advance for your help

PR #1025, opened 2021-03-25
Use proper use imports.

drupol:
### Summary

I'm trying to create a binary phar of a custom project with [humbug/box](https://packagist.org/packages/humbug/box). My project depends on Robo.

During the compile process, Humbug will rename namespaces in order to avoid collisions. The problem is that when we import a `trait` with an absolute path, it fails.

In order to fix this, you have to import it properly using the proper `use` statement.


greg-1-anderson:
Never realized that there was an instance where `use` produced different results than using the fully qualified namespace + classname. In any event, looks fine to me.

drupol:
Thanks :)

PR #1027, opened 2021-03-26
Add more types here and there.

drupol:
### Overview
This pull request:

- [x] Adds types here and there


greg-1-anderson:
I'm guessing at least some of these changes break backwards compatibility. I'd be okay with adding typehints in instances where we know there is no b/c issue (e.g. classes that cannot be overridden), but the rest will have to wait for 4.x.

There's no 4.x branch yet.

drupol:
No, I took great care to not break BC. I just inferred and added the return types, that's it.

PR #1028, opened 2021-03-29
Fix mkdir error in taskExtract on PHP 8

wxa:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
If you use `taskExtract` on PHP 8 with `error_reporting = E_ALL`, the task will always produce this error:
```
ERROR: mkdir(): File exists
in /vendor/consolidation/robo/src/Task/Archive/Extract.php:105
``` 
It happens because the line mentioned in the error message uses error suppression on `mkdir()` when trying to create a parent directory for the extraction destination folder:
```
@mkdir(dirname($this->to));
```
The parent folder will almost always exist, so an error will be emitted, and on PHP 8 the `@` operator does not work anymore for certain cases, such as the `E_ALL` level of error reporting.

This PR fixes the error.

### DIFF
--- a/src/Task/Archive/Extract.php
+++ b/src/Task/Archive/Extract.php
@@ -105,7 +105,7 @@ class Extract extends BaseTask implements BuilderAwareInterface
 
         $destinationParentDir = dirname($this->to);
         if (!file_exists($destinationParentDir)) {
-          @mkdir($destinationParentDir);
+            @mkdir($destinationParentDir);
         }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks for this. It would be best if we removed all of the `@` suppression markers and used wrapper functions consistently throughout the code, but I won't block this functional fix on that aspirational goal.

PR #1029, opened 2021-04-01
Added new 'composer show' task.

markdorison:
### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature

### Summary
Added new 'composer show' task.

### Description
I am attempting to add some logic to my Robo commands based on the installed version of a dependency. The output from `composer show` would be useful in determining that.

I imagine I am likely missing some things before these changes would be accepted and I am happy to iterate with some feedback.

I created this branch/PR against `2.x` as that is what I am currently using on my projects due to [drush's dependency requirements](https://github.com/drush-ops/drush/blob/1af5fe0881809bd8baa37386748a0276747554d2/composer.json#L39), but if there is interest in accepting these submissions, I would be happy to recreate it against `3.x`.

### DIFF
--- a/docs/tasks/Composer.md
+++ b/docs/tasks/Composer.md
@@ -328,6 +328,25 @@ $this->taskComposerRequire()->dependency('foo/bar', '^.2.4.8')->run();
 * `options(array $options, $separator = null)`  Pass multiple options to executable. The associative array contains
 * `optionList($option, $value = null, $separator = null)`  Pass an option with multiple values to executable. Value can be a string or array.
 
+## Show
+
+
+Composer Show
+
+``` php
+<?php
+// simple execution
+$this->taskComposerRequire()->dependency('foo/bar')->run();
+
+// inspect output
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
I'm not opposed to a `composer show` task, but for your use case, it might be better to detect your dependencies' capabilities through introspection, e.g. `class_exists`, `defined` et. al.

Also, it is fine / preferable to just use `exec` or the process component to get info to decide which tasks to create. This is a matter of style, of course, but the code is usually easier to read. Tasks are not well suited to returning output; they are better for things that make changes.

Regarding Drush, now that Robo 3 is stable, we could consider supporting ^3 in Drush. However, since Robo 3 uses a different version of league/container, it might be inconvenient to support both Robo 2 and Robo 3 at the same time. Might be worth doing if Drush needs the league/container upgrade for some other reason.

In any event, new features to Robo should go in 3.x first, and then be back-ported as needed.

markdorison:
Thanks for the feedback @greg-1-anderson. I appreciate it.

PR #1030, opened 2021-04-11
Call methods with 'self::' instead of 'static::' in Robo class.

rodrigoaguilera:
### Overview
This pull request:

- Adds a feature

### Summary
Change call to methods in the same class from `static` to `self`

### Description
In order to be able to override some methods of the Robo class with another class that extends from it I would like to change the way the methods are called. 
There is a demo of the `static::` vs. `self::` call in this comment
https://www.php.net/manual/en/language.oop5.static.php#104823

My goal is to add more services to the container in the `configureContainer` method

### DIFF
--- a/src/Robo.php
+++ b/src/Robo.php
@@ -104,7 +104,7 @@ class Robo
     public static function createConfiguration($paths)
     {
         $config = new \Robo\Config\Config();
-        static::loadConfiguration($paths, $config);
+        self::loadConfiguration($paths, $config);
         return $config;
     }
 
@@ -117,7 +117,7 @@ class Robo
     public static function loadConfiguration($paths, $config = null)
     {
         if ($config == null) {
-            $config = static::config();
+            $config = self::config();
         }
         $loader = new YamlConfigLoader();
         $processor = new ConfigProcessor();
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Overriding the Robo class feels wrong to me. Can you follow the template suggested in the guide, [Robo as a Framework](https://robo.li/framework/)?

The application configuration code in Robo is a little messy and I am interested in cleaning it up to make it more understandable and straightforward for clients. (This would be a breaking change, e.g. for Robo 4.) If you have suggestions on how to do that / what a better initialization sequence would look like, I'd be glad to hear your opinions.

If you think that overriding the DI container factory is really the best way to start up an app, maybe we should factor that out of the Robo class, which should really be a final collection of static methods.

rodrigoaguilera:
Thank you for taking the time to review and suggest a solution.
As you might have guessed this PR was related to
https://github.com/consolidation/Robo/pull/1031
and I ended up figuring out how to use `createContainer`.

I am using the "robo as a framework docs" as a guide to develop my own application and the part about using my own container felt like overriding a lot of code and since I was already subclassing the runner I decided to modify the container in the `execute()` method of the runner, mimicking how the container is initialized inside the `run()` method of the Runner.
https://gitlab.com/upstreamable/basin/-/blob/0.x/src/php/Runner.php

Right now I wouldn't know how to untangle the mess of the Robo class, I am just trying to lean on Robo by modifying as little as possible and learning a lot along the way. It is a great piece of sofware.

This PR is not useful anymore so I will close it.

PR #1031, opened 2021-04-11
Fix null return in createContainer.

rodrigoaguilera:
### Overview
This pull request:

- Fixes a bug

### Summary
Fix the createContainer method always returning null instead of the container.

### Description
I believe is just some unintended return.


greg-1-anderson:
Thanks

PR #1034, opened 2021-05-19
Add setProcessInput and deprecate setInput.

rodrigoaguilera:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature
- [x] Adds documentation

### Summary
Use a different method name to set the input for a process as the current one is overridden by the `InputAwareTrait` used by the `CollectionBuilder`

### Description
I was trying to provide a docker-compose.yml file via stdin by doing
```php
$this
  ->taskExec('docker-compose')
  ->setInput($YamlFileContents)
  ->interactive(false)
  ->option('file', '/dev/stdin')
  ->arg('up')
  ->option('detach')
  ->run();
```
but the `setInput()` that ends up being called is not the one on the `ExecTrait` so I propose to deprecate setInput as it wasn't working anyway and add a new method name that is more descriptive.

I am not sure that `interactive` should also be set to false automatically when you provide input for the process but that can be a discussion for another issue.

### DIFF
--- a/docs/tasks/Base.md
+++ b/docs/tasks/Base.md
@@ -21,6 +21,7 @@ if ($this->taskExec('phpunit .')->run()->wasSuccessful()) {
 
 * `simulate($context)`  {@inheritdoc}
 * `setOutput($output)`  Sets the Console Output.
+* `setProcessInput($output)`  Sets the input for the command. Similar to a pipe like `echo "input" | cat`. Usually followed by a call to `->interactive(false)` to avoid other inputs.
 * `dir($dir)`  Changes working directory of command
 * `arg($arg)`  Pass argument to executable. Its value will be automatically escaped.
 * `args($args)`  Pass methods parameters as arguments to executable. Argument values
@@ -49,7 +50,7 @@ $this->taskExecStack()
 * `executable($executable)`   * `param string` $executable
 * `exec($command)`   * `param string|string[]|CommandInterface` $command
 * `stopOnFail($stopOnFail = null)`   * `param bool` $stopOnFail
-* `result($result)` 
+* `result($result)`
 * `setOutput($output)`  Sets the Console Output.
 * `dir($dir)`  Changes working directory of command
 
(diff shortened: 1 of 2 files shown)


docs/tasks/Base.md
```diff
@@ -21,6 +21,7 @@ if ($this->taskExec('phpunit .')->run()->wasSuccessful()) {
 
 * `simulate($context)`  {@inheritdoc}
 * `setOutput($output)`  Sets the Console Output.
+* `setProcessInput($output)`  Sets the input for the command. Similar to a pipe like `echo "input" | cat`. Usually followed by a call to `->interactive(false)` to avoid other inputs.
```
greg-1-anderson: Typo, `$output` instead of $input.

greg-1-anderson:
> I am not sure that interactive should also be set to false automatically when you provide input for the process but that can be a discussion for another issue.

This does not appear to be done, currently. Are you proposing that it should be? If there are no use cases where you need a tty with stdin redirection (seems reasonable to me), then the right time to make this behavior change would be now, when the new method name is introduced. The deprecated method should maintain the old behavior.

rodrigoaguilera:
Thanks for having a look.
It doesn't make sense to provide an input and have a tty allocated so yes, let's make that change. If there is a corner case when interactive is needed it can be set later. Also fixed the typo.

src/Common/ExecTrait.php
```diff
@@ -215,12 +215,29 @@ public function envVars(array $env)
...
+     * @param resource|string $input
+     *
+     * @return $this
+     *
+     * @deprecated
+     */
+    public function setInput($input)
+    {
+        trigger_error('setInput() is deprecated. Please use setProcessInput(().', E_USER_DEPRECATED);
+        return $this->setProcessInput($input);
```
greg-1-anderson: `setInput` should maintain old behavior and not set interactive `false`

greg-1-anderson: This is probably not necessary but I can't prove that it won't cause a problem for someone.

rodrigoaguilera: But you are right, better keep the same behavior for the deprecated method.

greg-1-anderson:
Thanks!

Issue #1023, opened 2021-02-26
escapeArgument deprecation error


JKingweb:
### Steps to reproduce
`$this->taskExec("php")->option("-d", "zend.assertions=1")->arg("example.php")->run();`

### Expected behavior
Robo should run the task without issue

### Actual behavior
Robo runs the task and then prints this message:

> ERROR: The Robo\Common\ProcessUtils::escapeArgument() method is a copy of a method that was deprecated by Symfony 3.3 and removed in Symfony 4; it will be removed in Robo 2.0. 
in ./vendor/consolidation/robo/src/Common/ProcessUtils.php:39

This is despite running Robo 3.0. :P Using `->option()` is the source of the problem; I including an argument for clarity because one must otherwise quit PHP with Ctrl+D to make the error appear.

### System Configuration
Which O.S. and PHP version are you using?

PHP 8.0.2, Linux

greg-1-anderson:
Looks like that deprecated method is overdue for being removed. Run without `E_USER_DEPRECATED` to avoid seeing deprecated warnings.

PRs to clean up taskExec welcome.

markdorison:
I see a PR was created for this against 3.x which is fantastic. What should the approach to 2.x be? Should this be backported if committed or should the error message be updated to note that it will be removed in 3.x (instead of 2.0)?

greg-1-anderson:
If that PR does land on the 3.x branch, it could probably be backported to 2.x.  If it ends up with bc breaks, then it will have to be 4.x, in which case there might be the possibility of making a simpler version for 2.x and 3.x.

greg-1-anderson:
Oh, I was thinking of https://github.com/consolidation/Robo/pull/1038, not https://github.com/consolidation/Robo/pull/1036.

greg-1-anderson:
#1038 does not address the problem with escape in a way that solves this issue. Maybe the best solution is to simply un-deprecate this method in 2.x and 3.x (replace the trigger_error with a comment), and re-deprecate it once there is an actual replacement for escapeArgument. I think I was wrong in my Feb comment; this deprecation was premature; it therefore could be removed as a bugfix.

greg-1-anderson:
Also, for all who visit here, I will also mention that it is also advisable to just use Symfony Process directly if you want to exec a process and get some result that you use to build your task list. TaskExec is good for adding operations in the middle of a task list where you do not need to capture the result, or as the last operation in a task list.

PR #1038, opened 2021-07-06
Changed taskExec to use symfony process placeholders

Alenore:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Changed taskExec to use symfony process placeholders

### Description
taskExec still used some deprecated functions that should have been deleted in Robo 2.0.
Since it didn't, and was throwing E_USER_DEPRECATED when trying to escape some specifics args/options, I've switched the feature to instead use placeholders and let Symfony Process handle the escaping itself.

Please note that :
- It now escaped arguments, simple or not : `a` is escaped to `'a'`, so it's technically not a perfect replacement. However, Symfony Process handles these use case itself, and probably already replaced them anyway even if Robo didn't. Some tests have been dropped as a result, since there's no longer a "no escape" case ;
- I've had to copy two functions from Symfony Process : mostly the `escapeArgument()`, and `replacePlaceholders()`. The former is still used in other part of the core (notably `taskOpenBrowser()` and git commits), while the latter is used for the `getCommandDescription()` to rebuild the executed command to output in the console, and in tests ;
- I've chosen to generate a hash of the arg using md5, since it's quick, will handle duplicate argument values, and is very unlikely to generate duplicates for strings the length of an argument ;
- While running the test suite on Windows, I've had to fix the environment variable set test case : `env` command isn't available on Win, so the test was always failing ;
- I've updated most comments mentioning the old arguments escaping behaviour, but I may have missed some.

Feel free to ask if you need any clarifications.

### DIFF
--- a/src/Common/CommandArguments.php
+++ b/src/Common/CommandArguments.php
@@ -43,6 +43,19 @@ trait CommandArguments
     }
 
     /**
+     * Escape a command line argument.
+     *
+     * @param string $argument
+     * @return string
+     *
+     * @deprecated Use \Robo\Common\ProcessUtils::escapeArgument() instead.
+     */
+    public function escape($argument)
+    {
+        return ProcessUtils::escapeArgument($argument);
+    }
+
+    /**
      * Pass argument to executable. It will be changed to a placeholder.
      *
      * @param string $arg
--- a/src/Common/CommandArguments.php
+++ b/src/Common/CommandArguments.php
@@ -50,7 +50,7 @@ trait CommandArguments
      *
      * @deprecated Use \Robo\Common\ProcessUtils::escapeArgument() instead.
      */
-    public function escape($argument)
+    public static function escape($argument)
     {
         return ProcessUtils::escapeArgument($argument);
     }
--- a/src/Common/CommandArguments.php
+++ b/src/Common/CommandArguments.php
@@ -29,7 +29,7 @@ trait CommandArguments
      * @param string|null $separator
      * @return void
      */
-    public function addArgument($argument, $key = null, $prefix = null, $separator = ' ')
+    protected function addArgument($argument, $key = null, $prefix = null, $separator = ' ')
     {
         if (is_null($key)) {
             $key = "arg" . md5($argument);
--- a/src/Common/CommandArguments.php
+++ b/src/Common/CommandArguments.php
@@ -15,7 +15,35 @@ trait CommandArguments
     protected $arguments = '';
 
     /**
-     * Pass argument to executable. Its value will be automatically escaped.
+     * @var array
+     */
+    protected $argumentsEnv = [];
+
+
+    /**
+     * Adds an argument and its placeholder value, to be executed
+     *
+     * @param string $argument
+     * @param string|null $key
+     * @param string|null $prefix
+     * @param string|null $separator
+     * @return void
+     */
+    public function addArgument($argument, $key = null, $prefix = null, $separator = ' ')
+    {
+        if (is_null($key)) {
+            $key = "arg" . md5($argument);
+        }
+
+        $this->arguments .= null == $prefix ? '' : $separator . $prefix;
+        if (!is_null($argument)) {
+            $this->arguments .= ' "${:' . $key . '}"';
+            $this->argumentsEnv[$key] = $argument;
+        }
+    }
+
+    /**
+     * Pass argument to executable. It will be changed to a placeholder.
      *
      * @param string $arg
      *
@@ -27,8 +55,8 @@ trait CommandArguments
     }
 
     /**
-     * Pass methods parameters as arguments to executable. Argument values
-     * are automatically escaped.
+     * Pass methods parameters as arguments to executable. Argument are
+     * automatically passed as placeholders
      *
      * @param string|string[] $args
      *
@@ -40,7 +68,10 @@ trait CommandArguments
         if (!is_array($args)) {
             $args = $func_args;
         }
-        $this->arguments .= ' ' . implode(' ', array_map('static::escape', $args));
+
+        foreach ($args as $arg) {
+            $this->addArgument($arg);
+        }
         return $this;
     }
 
@@ -59,24 +90,8 @@ trait CommandArguments
     }
 
     /**
-     * Escape the provided value, unless it contains only alphanumeric
-     * plus a few other basic characters.
-     *
-     * @param string $value
-     *
-     * @return string
-     */
-    public static function escape($value)
-    {
-        if (preg_match('/^[a-zA-Z0-9\/\.@~_-]+$/', $value)) {
-            return $value;
-        }
-        return ProcessUtils::escapeArgument($value);
-    }
-
-    /**
      * Pass option to executable. Options are prefixed with `--` , value can be provided in second parameter.
-     * Option values are automatically escaped.
+     * Option values are automatically passed as placeholders.
      *
      * @param string $option
      * @param string $value
@@ -89,15 +104,15 @@ trait CommandArguments
         if ($option !== null and strpos($option, '-') !== 0) {
             $option = "--$option";
         }
-        $this->arguments .= null == $option ? '' : " " . $option;
-        $this->arguments .= null == $value ? '' : $separator . static::escape($value);
+
+        $this->addArgument($value, null, $option, $separator);
         return $this;
     }
 
     /**
      * Pass multiple options to executable. The associative array contains
      * the key:value pairs that become `--key value`, for each item in the array.
-     * Values are automatically escaped.
+     * Values are passed as placeholders.
      *
      * @param array $options
      * @param string $separator
--- a/src/Common/CommandArguments.php
+++ b/src/Common/CommandArguments.php
@@ -52,6 +52,9 @@ trait CommandArguments
      */
     public static function escape($argument)
     {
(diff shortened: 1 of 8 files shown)


src/Common/CommandArguments.php
```diff
@@ -15,7 +15,35 @@ trait CommandArguments
...
+    /**
+     * Adds an argument and its placeholder value, to be executed
+     *
+     * @param string $argument
+     * @param string|null $key
+     * @param string|null $prefix
+     * @param string|null $separator
+     * @return void
+     */
+    public function addArgument($argument, $key = null, $prefix = null, $separator = ' ')
```
greg-1-anderson: Is there a use-case for calling this directly, or could it be made protected?

src/Common/CommandArguments.php
```diff
@@ -15,7 +15,35 @@ trait CommandArguments
     protected $arguments = '';
 
     /**
-     * Pass argument to executable. Its value will be automatically escaped.
+     * @var array
+     */
+    protected $argumentsEnv = [];
```
greg-1-anderson, line 20: Does `$argumentsEnv` have anything to do with environment variables? If this is just the key:value pairs for substitution, then it would probably be better to give it a different name.

greg-1-anderson:
What are the b/c breaks here? If variations in argument escaping are not considered a b/c break (as long as the same taskExec calls run the same way), then are there any b/c breaks?

src/Common/CommandArguments.php
Alenore: It's a mistake, it should be protected. I'll push a fix.

src/Common/CommandArguments.php
Alenore: Env is the name of the argument used by Symfony Process : they pass the placeholders either at exec time, or via environment variables. It's used as environment in this case, since it'd require further changes to the process exec workflow to be passed at the execution step.

Alenore:
There's no breaking change per se on the execution level, but there is to the format of the commands themselves: if someone implemented a custom check of the command (for instance via unit testing, which I had to update in this case), it'll break.

src/Common/CommandArguments.php
Alenore: Fix pushed

src/Task/Base/Exec.php
```diff
@@ -132,7 +134,7 @@ public function run()
     {
         $this->hideProgressIndicator();
         // TODO: Symfony 4 requires that we supply the working directory.
-        $result_data = $this->execute(Process::fromShellCommandline($this->getCommand(), getcwd()));
+        $result_data = $this->execute(Process::fromShellCommandline($this->getCommand(), getcwd(), $this->argumentsEnv));
```
greg-1-anderson, line 137: OK, so this line is using `$argumentsEnv` to set environment variables -- but didn't you already insert the argument values into the command line via `replacePlaceholders`, or is that not in the execution path here?

How do these environment variables interact with `$tastExec->env()`? It looks like there are still test showing that this works.

src/Common/CommandArguments.php
```diff
@@ -58,25 +89,9 @@ public function rawArg($arg)
...
 
-    /**
-     * Escape the provided value, unless it contains only alphanumeric
-     * plus a few other basic characters.
-     *
-     * @param string $value
-     *
-     * @return string
-     */
-    public static function escape($value)
```
greg-1-anderson, line 69: Removing a public method is a bc break -- didn't notice this before. To merge this PR into 3.x, we'd need to restore this method.

src/Task/Base/Exec.php
Alenore: The `replacePlaceholders()` used in `getCommandDescription()` is there to display the command line that will be run by Symfony. The only reason I had to copy that method from SymfonyProcess into ProcessUtils is to rebuild the command line to display it. If there was another way to get a command translated from placeholders to actual value before executing it, it would be preferable, alas Process doesn't seem to provide that feature.
However, it's not meant to rebuild the command line before executing it : let's let Symfony handle that part.

These environment variable are passed as `proc_open()` 5th argument by Process, so it's entirely self-contained into the process we mean to execute, and and won't affect environment variables you can access while in Robo in any means.

src/Common/CommandArguments.php
Alenore: Restored that method.

Alenore:
You mentionned in #1023 that this PR didn't fix that specific issue, however the whole point of that PR was to delegate the arguments escaping to Process, while still providing a way to escape arguments the same way Process does if people still want to escape it beforehand.
To be clear, #1023 is fixed with these changes, it was actually the first case I've tested against.

greg-1-anderson:
Yes, you're right, taskExec is fixed, but the escape method needs to remain because it's public, and other code might be calling it directly. You'll need to rebase to get the change that removed the call to trigger error.

Alenore:
That method was updated with the latest from Process, so there's still a public static `ProcessUtils::escapeArgument()` method.
Since the method has been updated, I cannot just include the other PR changes. Besides, `escapeArgument()` now has a reason to exist and should not be removed at all, since it's used to recompose the command description.

greg-1-anderson:
You're right, I was confused by the diff. I am also not used to the new GitHub feature where I have to approve the tests to run. I'm going to do that now.

greg-1-anderson:
OK this should be about ready to be merged now; I just want to give it one more review to make sure I didn't miss anything else or make any other mistakes.

Alenore:
The escape() function was incorrectly set as non static, just fixed that.

src/Common/CommandArguments.php
```diff
@@ -15,7 +15,48 @@ trait CommandArguments
...
+     * Escape a command line argument.
+     *
+     * @param string $argument
+     * @return string
+     *
+     * @deprecated Use \Robo\Common\ProcessUtils::escapeArgument() instead.
+     */
+    public static function escape($argument)
+    {
+        return ProcessUtils::escapeArgument($argument);
```
greg-1-anderson, line 58: Could you also restore the conditional that returns the unescaped string when it is a simple alphanumeric? This code won't be used much once this PR lands, but it's best to maintain its current behavior.

src/Common/ProcessUtils.php
```diff
@@ -24,58 +24,47 @@ private function __construct()
...
 
-    private static function isSurroundedBy($arg, $char)
+    /**
+     * Used by Symfony Process to form the final command line, with escaped parameters.
+     * Robo uses placeholders to handle the process arguments without escaping the whole string.
+     *
+     * @param string|null $argument
+     * @return string
+     */
+    public static function escapeArgument(?string $argument): string
```
greg-1-anderson, line 52: Since escapeArgument is only used for display purposes in taskExec, maybe we should retain the existing implementation. The new version might have some benefits for Windows, but since it will only be used by code that calls `escape` directly, it would make an odd edge case if the changed version caused problems.

greg-1-anderson: I'm undecided about this; if we don't update the implementation, then the printed string won't exactly match the string that Symfony is actually exec'ing, which could be odd. Then again, you could also get into that situation if you ended up with an older version of Symfony that has a different implementation of escape. I'm not sure if our Composer version constraints on our dependencies allows that, though.

src/Task/Base/Exec.php
```diff
@@ -98,8 +99,9 @@ public function background($arg = true)
      */
     protected function getCommandDescription()
     {
-        return $this->getCommand();
```
greg-1-anderson, line 101: Now I am wondering if this PR is in fact breaking, since you have changed the contract on `getCommand()`. This isn't just relevant for unit tests; Robo uses `getCommand()` for a number of things, and if this method starts returning a string that contains placeholder instead of a string with escaped arguments, then code that called getCommand and passed it to some process execution function would fail.

Would it work out to make `getCommandDescription()` continue to call / return `getCommand()`, have `getCommand()` call `replacePlaceholders`, and add a new method `getCommandTemplate()` that returns the command string with placeholders, and use that in `Task\Exec::run()`?

src/Common/ProcessUtils.php
Alenore: Symfony Process is set at 4.4.11 min, and escapeArgument has been changed to this implementation since at least 4.1.7.
Besides, that method was set as deprecated and meant to be removed in Robo 2.0, with a E_USER_DEPRECATED triggered. I don't think anybody would be really surprised it finally gets changed.

src/Task/Base/Exec.php
Alenore: So I've just tested outside of taskExec, and it does break on other commands : most command include a custom `run()` function, so placeholders don't get replaced.
Something I've noticed on other commands is that `getCommand()` is often used to display the command to be executed, when `getCommandDescription()` should be used instead.

In any case, there's more than `Exec::run()` that needs replacing, but every run() method that uses the `ExecOnCommand` trait, since they all get their arguments from `CommandArguments` and need to have their placeholders replaced.
As far as I know, only `Exec` has a custom process creation, every other calling the `ExecCommand::executeCommand()` method. It would simply be a matter of overwriting it in an `ExecOneCommand::executeCommand()` method that would handle placeholders.

So to sum things up :
- Add a `getCommandTemplate()` method ;
- Change `ExecOneCommand::executeOneCommand()` to replace placeholders.

Sounds good?

greg-1-anderson: Yes, that sounds good.

PR #1039, opened 2021-07-07
Undeprecate escapeArgument().

### LINKED ISSUE #1023: escapeArgument deprecation error

markdorison:
### Overview
This pull request:

- [ ] Fixes a bug
- [ ] Adds a feature

### Summary
Undeprecate `ProcessUtils::escapeArgument()` and replaces error with updated method
documentation per https://github.com/consolidation/Robo/issues/1023#issuecomment-875198155.

Should hopefully be able to be backported to 2.x.

Closes #1023.

### DIFF
--- a/src/Common/ProcessUtils.php
+++ b/src/Common/ProcessUtils.php
@@ -26,18 +26,18 @@ class ProcessUtils
     /**
      * Escapes a string to be used as a shell argument.
      *
+     * This method is a copy of a method that was deprecated by Symfony 3.3 and
+     * removed in Symfony 4; it will be removed once there is an actual
+     * replacement for escapeArgument.
+     *
      * @param string $argument
      *   The argument that will be escaped.
      *
      * @return string
      *   The escaped argument.
-     *
-     * @deprecated since version 3.3, to be removed in 4.0. Use a command line array or give env vars to the `Process::start/run()` method instead.
      */
     public static function escapeArgument($argument)
     {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Looks good, thanks for the PR. Backport welcome.

greg-1-anderson:
Actually running the tests before merging. Must pass, right?

Alenore:
Symfony Process does provide a new `escapeArgument()` method, however in the form of a private method. Shouldn't it be used instead of the old one?

greg-1-anderson:
If the new escapeArguement works better than the one that's copied here, we could pull it in as follow-on work. Since your other PR makes the escape method mostly unused, it might not be worth the effort. Up to you if you want to address this in your PR or not.

PR #1043, opened 2021-07-27
add support for compileString() method in scssphp 1.5 and above

stklcode:
### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Add support for new `compileString()` method which returns a result object instead of string.

### Description
The SCSS task uses `Compiler::compile()` which is deprecated in current scssphp. 
Check, if `compileString()` is available and use it, if so.

This change does not break compatibility with `scssphp/scssphp ~1.0.0`, but eliminates the deprecation warning with 1.5 and above.

Could probably ease extension for source map, because the `CompilationResult` has a `getSourceMap()` method to access the output from the very same result object.

### DIFF
--- a/src/Task/Assets/Scss.php
+++ b/src/Task/Assets/Scss.php
@@ -64,7 +64,12 @@ class Scss extends CssPreprocessor
             $scss->setFormatter($this->compilerOptions['formatter']);
         }
 
-        return $scss->compile($scssCode);
+        if (method_exists($scss, 'compileString')) {
+            // "compileString()" is available since scssphp v1.5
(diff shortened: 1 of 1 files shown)


clubdeuce:
What is the status of this? scssphp 2.0 has removed the `compile` method completely.

stklcode:
Friendly ping @greg-1-anderson 

I know this PR is quite old, back from the 3.x days, and the SCSS task is probably none of the top used features, but I guess the compatibility fix is still valid.

If there’s something blocking or additional improvement in the wishlist, I’d offer do dig into it again. Otherwise feel free to either merge or close it as is.

PR #1055, opened 2021-11-09
[WIP] Homebrew integration

stovak:
### Overview
This pull request:

- [ ] Fixes a bug
- [X] Adds a feature

### Summary
Adding building of Phar and Home-brew formula

### Description

Proposed Workflow


src/Task/Archive/Pack.php
```diff
@@ -201,16 +201,39 @@ public function run()
      */
     protected function archiveTar($archiveFile, $items)
```
stovak: @kporras07 would love a second set of eyes on this rewrite. The Archive_Tar has security issues so we need to replace it with PHP's PharData class. Tests on the archives are failing ATM.

kporras07: It looks good. I think $p_remove_dir and $p_add_dir are now dangling over there unused.

Other than that, I think it looks good but I haven't tested it.

Let me know if you need me to test it somehow or do anything else with this

composer.json
```diff
@@ -113,7 +116,6 @@
     }
   },
   "suggest": {
-    "pear/archive_tar": "Allows tar archives to be created and extracted in taskPack and taskExtract, respectively.",
```
greg-1-anderson: Cool.

src/Task/Archive/Pack.php
greg-1-anderson: If there is a general need to remove archive_tar, perhaps that should happen in a separate PR.

stovak: Its more a need to get tests to pass so the workflows work. But archive_tar is on the list of security issues, but that's incidental to making the Pack and Extract tests pass.

stovak: Anyway, found there was a patched version and updated to the patched version and restored the original code.

PR #1057, opened 2021-11-10
Image Minify: Skip files without minifier & replace faulty message

pixelbrackets:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Fixes a faulty error message in case the image minifier processes all files,
and does not crash before, but some images can not be minified.

Also skip the minification process for all files which have no minifier set.
This allows to continue the task and not stop the minification process
for all files.

### Description
In case of an error the message contains placeholders which
are not replaced anymore downstream
(`Minified {filecount} out of {filetotal} images into {destination}`).
Print the result message as info and then return a short task message
without placeholders instead.

Early return for all files which have no minifier set,
to avoid fatal error with ambigous error message
“Minifier method  cannot be found!”.
For example when webp images exist in the image directory.
The files are added to the error list right away,
to that the task may continue to minify other images and
show a result message (“Minified X of Y files”) at the end.

### DIFF
--- a/src/Task/Assets/ImageMinify.php
+++ b/src/Task/Assets/ImageMinify.php
@@ -216,15 +216,15 @@ class ImageMinify extends BaseTask
         }
 
         $amount = (count($files) == 1 ? 'image' : 'images');
-        $message = "Minified {filecount} out of {filetotal} $amount into {destination}";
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I would take this, but it needs a forward-port to the 4.x branch.

PR #1062, opened 2021-11-22
update LICENSE year

### LINKED ISSUE #1061: update LICENSE year
https://github.com/consolidation/robo/blob/3.x/LICENSE#L3

Arhell:
fix issue https://github.com/consolidation/robo/issues/1061

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014-2020 Codegyre Developers Team, Consolidation Team
+Copyright (c) 2014-2021 Codegyre Developers Team, Consolidation Team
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson (approved):
This used to be automatically updated on every `composer update` run by composer-test-scenarios. We no longer use that Composer plugin; it would be useful, perhaps, to break out the dependency license script so that it could be used standalone where desired.

Arhell:
> This used to be automatically updated on every `composer update` run by composer-test-scenarios. We no longer use that Composer plugin; it would be useful, perhaps, to break out the dependency license script so that it could be used standalone where desired.

Can I merge?

greg-1-anderson:
Yes, please go ahead and merge.

PR #1083, opened 2021-12-17
League container 4

nicosp:
### Overview
This pull request:

- [ ] Fixes a bug
- [ ] Adds a feature
- [ x] Breaks backwards compatibility

### Summary
Updates league/container to 4.x

### Description
Required for compatibility with cake 4.3


greg-1-anderson:
We can't bump our minimum php version without also bumping the Robo major version. I'm not sure about the timeframe for a new major version of Robo. I'm not inclined to do that soon.

This PR would be viable for Robo 3.x if it kept the same minimum php version and used `"league/container": "^4.0",`. We'd need a wrapper method that called either `$container->addShared()` or `$container->share()` as needed for the current version of league/container.

nicosp:
Done. I added support for both versions since container 4.x is php 7.2+

composer.json
```diff
@@ -59,7 +59,7 @@
         "optimize-autoloader": true,
         "sort-packages": true,
         "platform": {
-            "php": "7.2.28"
+            "php": "7.2"
```
greg-1-anderson: Past versions of Composer have interpreted "7.2" as "7.2.0". If this hasn't changed, using this limit would prevent dependencies requiring e.g. ^7.2.5 from being used. Unless there is a specific reason why the shorter version is now better, I'd prefer to specify the patch version here.

greg-1-anderson (approved):
LGTM; this can be merged once the platform.php configuration is explained or fixed.

composer.json
nicosp: The change was left accidentally. I wanted to check symfony 4.

PR #1084, opened 2021-12-24
PackPhar: Don't strip PHP8 attributes

Naktibalda:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary

PackPhar::addStripped removes all comments from file, including PHP8 attributes, when used on PHP7.  
This change makes addStripped keep the attributes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,10 +92,12 @@ jobs:
           - os: ubuntu-latest
             php-version: "7.1"
             dependencies: lowest
+            php-ini-values: assert.exception=1, zend.assertions=1, phar.readonly=false
 
           - os: ubuntu-latest
             php-version: "7.4"
             dependencies: highest
+            php-ini-values: assert.exception=1, zend.assertions=1, phar.readonly=false
 
           - os: ubuntu-latest
             php-version: "8.0"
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
Looks good; would be fine with passing tests. Either change the CI script to set up php.ini, or perhaps use `php -d ... phpunit`

Naktibalda:
I updated workflow script.

.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           - "7.4"
 
         php-ini-values:
-          - assert.exception=1, zend.assertions=1
+          - assert.exception=1, zend.assertions=1, phar.readonly=false
```
greg-1-anderson, line 86: Ah, right, GitHub actions makes this really easy. Awesome.

Naktibalda: Not so easy as I expected. Have you got any idea why it had no effect on PHP 7.1 and 7.4?

greg-1-anderson:
Workflow not running, maybe a syntax error in config.yml?

greg-1-anderson:
I updated the project settings, so hopefully your future commits will run as soon as they are pushed.

Naktibalda:
CI passed

greg-1-anderson:
Thanks for the fix!

Issue #1087, opened 2021-12-27
taskExtract - avoid cross volume rename 


Sweetchuck:
### Steps to reproduce
Use $this->taskExtract() to extract a *.bz2 file when the cwd() and the $target point to different volumes.
This situation leads to [FilesystemStack::_rename()](https://github.com/consolidation/robo/blob/0e4d27bc1bdd76cfdd57ccbb8b4046d134a0d5d7/src/Task/Filesystem/FilesystemStack.php#L121)

### Expected behavior
The temporary directory should be based on the `$this-to` instead of the `getcwd()`.
See [Task\Archive\Extract::getTmpDir()](https://github.com/consolidation/robo/blob/0e4d27bc1bdd76cfdd57ccbb8b4046d134a0d5d7/src/Task/Archive/Extract.php#L279)

With the following patch I was able to solve this problem:
```patch
diff --git a/src/Task/Archive/Extract.php b/src/Task/Archive/Extract.php
index 817608a..f2eee84 100644
--- a/src/Task/Archive/Extract.php
+++ b/src/Task/Archive/Extract.php
@@ -100,12 +100,12 @@ class Extract extends BaseTask implements BuilderAwareInterface
         }
 
         // We will first extract to $extractLocation and then move to $this->to
-        $extractLocation = static::getTmpDir();
+        $extractLocation = static::getTmpDir($this->to);
         @mkdir($extractLocation);
 
         $destinationParentDir = dirname($this->to);
         if (!file_exists($destinationParentDir)) {
-            @mkdir($destinationParentDir);
+            @mkdir($destinationParentDir, 0777 - umask(), true);
         }
 
         $this->startTimer();
@@ -276,8 +276,8 @@ class Extract extends BaseTask implements BuilderAwareInterface
     /**
      * @return string
      */
-    protected static function getTmpDir()
+    protected static function getTmpDir($to)
     {
-        return getcwd() . '/tmp' . rand() . time();
+        return "$to-tmp" . rand() . time();
     }
 }
```
And I think ::getTmpDir() method should not be static.

PR #1090, opened 2021-12-27
Issue #1087 - Avoid cross volume rename in Task\Archive\Extract

Sweetchuck:
Issue #1087

### Overview
This pull request:

- [ ] Fixes a bug
- [x] Adds a feature 

### Adds a feature
I think in some cases it will be faster.

### Breaks backwards compatibility

`protected static function getTmpDir()`
vs
`protected function getTmpDir()`

### DIFF
--- a/src/Task/Archive/Extract.php
+++ b/src/Task/Archive/Extract.php
@@ -99,13 +99,15 @@ class Extract extends BaseTask implements BuilderAwareInterface
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
We could leave the old `getTmp()` in place with a `@deprectated` tag to preserve backwards compatibility.

greg-1-anderson (approved):
LGTM

Issue #1097, opened 2022-01-18
Progressbar vs logger vs stdOutput


Sweetchuck:
### Steps to reproduce
```php
<?php

declare(strict_types = 1);

use Consolidation\AnnotatedCommand\CommandResult;
use Psr\Log\LoggerAwareInterface;
use Psr\Log\LoggerAwareTrait;
use Robo\Common\ConfigAwareTrait;
use Robo\Contract\ConfigAwareInterface;
use Robo\Tasks;

class RoboFileDummy extends Tasks implements LoggerAwareInterface, ConfigAwareInterface
{
    use LoggerAwareTrait;
    use ConfigAwareTrait;

    /**
     * @command dummy:progressbar
     */
    public function cmdDummyProgressbarExecute(
        array $options = [
            'format' => 'list',
        ]
    ) {
        $result = $this
            ->collectionBuilder()
            ->addCode(function (\Robo\State\Data $state): int {
                $state['items'] = [
                    'item-01',
                    'item-02',
                    'item-03',
                ];

                $this->logger->info('a');
                sleep(2);
                $this->logger->info('b');

                return 0;
            })
            ->addCode(function (\Robo\State\Data $state): int {
                $this->logger->info('c');
                sleep(2);
                $this->logger->info('d');

                return 0;
            })
            ->addCode(function (\Robo\State\Data $state): int {
                sleep(2);
                $this->logger->info('e');
                sleep(2);
                $this->logger->info('f');

                return 0;
            })
            ->addCode(function (\Robo\State\Data $state): int {
                sleep(2);
                $this->output()->writeln('stdOutput line 1');
                sleep(2);
                $this->output()->writeln('stdOutput line 2');

                return 0;
            })
            ->run();

        $items = $result['items'];

        return CommandResult::data($items);
    }
}
```

-----------------

**Command:**
`./vendor/bin/robo --load-from='RoboFileDummy.php' -vv dummy:progressbar`

**Actual output:**
```
 [info] a
 [info] b
 0/4 [>---------------------------]   0% < 1 sec/< 1 sec [info] c
 [info] d
 2/4 [==============>-------------]  50% 2 secs/4 secs [info] e
 [info] f
 3/4 [=====================>------]  75% 6 secs/8 secsstdOutput line 1
stdOutput line 2
 4/4 [============================] 100% 10 secs/10 secs
item-01
item-02
item-03
```

**Expected:**
1. The problem is that, the progressbar and a log entry are in the same line.
[...]

PR #1101, opened 2022-02-14
Issue #1097 - ProgressIndicator uses the stdError

Sweetchuck:
Issue #1097 - ProgressIndicator uses the stdError

### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

One step closer, but this doesn't solve the problem when a progressBar and a log entry in the same line.
`2/4 [==============>-------------]  50% 2 secs/4 secs [info] my log entry`

But the stdOutput is clean (or cleaner), so it can be used in pipes.
`robo foo | other`

This modification removes the leading `0D 0A` characters from the stdOutput

### DIFF
--- a/src/Common/ProgressIndicator.php
+++ b/src/Common/ProgressIndicator.php
@@ -2,6 +2,8 @@
 
 namespace Robo\Common;
 
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+
 /**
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Yes, the progress bar definitely should be on stderr, not stdout

Issue #1102, opened 2022-02-14
Config - incompatible function signatures


Sweetchuck:
### Steps to reproduce
| package | version | commit |
|---------|---------|--------|
| consolidation/robo | 3.x-dev |  5d0c99575a85fd8f5e2b2fb8c49974905b3a249a |
| consolidation/config | 2.x-dev | fde7dfee1885ecca9afa3adb8442711d92a5c8ce |
| dflydev/dot-access-data | v3.0.1 | |
| grasmash/expander | dev-master | 84c256112763e00d09b437d269897339ca1ceaf4 |

\Robo\Robo::createConfiguration()
calls the `new \Robo\Config\Config();` constructor without argument.
`$data = null`
But the `\Dflydev\DotAccessData\Data::__construct` does not accepet NULL.

### Expected behavior
No fatal error

### Actual behavior
> PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Dflydev\DotAccessData\Data::__construct() must be of the type array, null given, called in /project/vendor/consolidation/config/src/Config.php on line 80 and defined in /project/vendor/dflydev/dot-access-data/src/Data.php:40
Stack trace:
0 /project/vendor/consolidation/config/src/Config.php(80): Dflydev\DotAccessData\Data->__construct()
1 /project/vendor/consolidation/robo/src/Config/Config.php(43): Consolidation\Config\Config->replace()
2 /project/vendor/consolidation/robo/src/Config/Config.php(35): Robo\Config\Config->replace()
3 /project/vendor/consolidation/robo/src/Config/Config.php(26): Robo\Config\Config->import()
4 /project/vendor/consolidation/robo/src/Robo.ph in /project/vendor/dflydev/dot-access-data/src/Data.php on line 40

greg-1-anderson:
This is fixed in dev-main of consolidation/config. I think you need to be using a non-stable version of grasmash/expander to be affected.

Sweetchuck:
I am using the latest commit from the consolidation/config. Which is the [fde7dfe](https://github.com/consolidation/config/commit/fde7dfee1885ecca9afa3adb8442711d92a5c8ce)

2.x-dev is a branch alias in the composer.json
https://github.com/consolidation/config/blob/fde7dfee1885ecca9afa3adb8442711d92a5c8ce/composer.json#L63

Sweetchuck:
Before the upgrade everything worked fine.

**consolidation/config**
[...]

PR #1103, opened 2022-02-14
Issue #1102 - Makes \Robo\Config\Config compatible with \Dflydev\DotAccessData\Data

Sweetchuck:
Issue #1102

### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -19,7 +19,7 @@ class Config extends ConfigOverlay implements GlobalOptionDefaultValuesInterface
      * Create a new configuration object, and initialize it with
      * the provided nested array containing configuration data.
      */
-    public function __construct(array $data = null)
+    public function __construct(array $data = [])
     {
         parent::__construct();
 


greg-1-anderson:
We don't want to change the function signature of the constructor until the next major version, to preserve backwards compatibility. Could you fix up the value in the body, as done in  https://github.com/consolidation/config/commit/fde7dfee1885ecca9afa3adb8442711d92a5c8ce?

Sweetchuck:
My first try was to change this in \Robo\Robo::createConfiguration()
from
`$config = new \Robo\Config\Config();`
to
`$config = new \Robo\Config\Config([]);`

Actually function like this is almost pointless `function foo(array $x = null)`, because it can't be called like this `foo(null);` only without argument `foo()` or with a valid argument `foo([]);`

On my project I reverted every dependency back to a working state. I can wait for an official solution.

Thanks anyway

greg-1-anderson:
Actually we never made a stable release of 4.x, so I'll take this as is.

greg-1-anderson:
Whoops. Reverted on the 3.x branch, applied to the 4.x branch.

PR #1106, opened 2022-02-24
Test with dev version of consolidation/log that allows psr-log ^2

greg-1-anderson:


greg-1-anderson:
No changes required other than dependencies in composer.lock, so no release necessary.

PR #1108, opened 2022-04-05
Allow consolidation/log ^3

greg-1-anderson:
Supporting consolidation/log ^3 required a couple of trivial b/c breaks. Robo 4 should be mostly the same as Robo 3 for most clients. Taking advantage of the version bump to drop support for some old components; use Robo ^3 || ^4 if you need compatibility with the full range of dependencies.


alexpott:
I think we're going to need https://github.com/grasmash/expander/pull/17

PR #1112, opened 2022-04-21
Fix `Process::inheritEnvironmentVariables()` deprecation error

JanJakes:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
We're running into the following error when using env():

> ERROR: The "Symfony\Component\Process\Process::inheritEnvironmentVariables()" method is deprecated since Symfony 4.4, env variables are always inherited.

The method call can be safely removed as the current required version of `symfony/process` is "^4.4.9 || ^5 || ^6".

### DIFF
--- a/src/Common/ExecTrait.php
+++ b/src/Common/ExecTrait.php
@@ -374,11 +374,6 @@ trait ExecTrait
         }
 
         if (isset($this->env)) {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Thanks!

PR #1120, opened 2022-07-31
[GenerateMarkdownDoc] Use var_export for rendering default param values

### LINKED ISSUE #1119: GenerateMarkdownDoc renders Boolean default values incorrectly
### Steps to reproduce

I generated markdown of `public static function addNamespace(string $prefix, string $baseDir, bool $prepend = false): void`

### Expected behavior

`*public static* addNamespace($prefix, $baseDir, $prepend = false)`

### Actual behavior

`*public static* addNamespace($prefix, $baseDir, $prepend = )`

### System Configuration
Ubuntu 20.04, PHP 8.0.21

The problem is that GenerateMarkdown uses `print_r` for rendering default values. `print_r` is completely unsuitable for printing Boolean values because it renders `false` as empty string and `true` as `1`.

I think that it would be better to use `var_export` instead.

Naktibalda:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
`print_r` is completely unsuitable for printing Boolean values because it renders false as empty string and true as 1. So I replaced it with `var_export`.

### Description
Fixes #1119

### DIFF
--- a/src/Task/Development/GenerateMarkdownDoc.php
+++ b/src/Task/Development/GenerateMarkdownDoc.php
@@ -685,7 +685,7 @@ class GenerateMarkdownDoc extends BaseTask implements BuilderAwareInterface
             if ($param->allowsNull()) {
                 $text .= ' = null';
             } else {
-                $text .= ' = ' . str_replace("\n", ' ', print_r($param->getDefaultValue(), true));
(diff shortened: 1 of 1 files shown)


DavertMik:
Done!

PR #1122, opened 2022-08-21
Document method parameters and return types

Naktibalda:
### Overview
This pull request:

- [x] Adds a feature

### Summary
A lot of redundant `@param` and `@return` annotation had been removed after declaring correct types in Codeception.

This change documents declared types of method parameters and return types.

### Description

Effect of this change can be seen at https://github.com/Codeception/codeception.github.com/pull/646

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "consolidation/output-formatters": "^4.1.2",
         "consolidation/self-update": "^2.0",
         "league/container": "^3.3.1 || ^4.0",
+        "phpowermove/docblock": "^4.0",
         "symfony/console": "^6",
         "symfony/event-dispatcher": "^6",
         "symfony/filesystem": "^6",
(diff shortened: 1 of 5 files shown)


DavertMik (approved):

DavertMik:
This PR changes only GenMarkdownTask which we use internally in Codeception so I'm for this to be merged
@greg-1-anderson are you ok with that?

.gitignore
```diff
@@ -1,3 +1,4 @@
+composer.lock
```
greg-1-anderson: Robo does highest / lowest testing, which requires composer.lock be committed to the repository. Committing the composer.lock also allows the phar build to be reproducible.

greg-1-anderson (changes requested):
Looks good for me, except composer.lock should not be added to the .gitignore.

DavertMik:
@greg-1-anderson can we merge it?

DavertMik:
@greg-1-anderson I'm merging it?

Naktibalda:
I made a small change to document parameters in correct order, when some parameters have `@param` annotations and others don't.

Effect can be seen in this commit: https://github.com/Codeception/codeception.github.com/commit/36b6d9891080afab38dcb14284b988c9614a3793

greg-1-anderson:
LGTM. Thanks.

PR #1133, opened 2022-12-11
Fix php 8.2 compatibility

chapterjason:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
- Updates dependency to which is 8.2 incompatible
- Fix callable which is 8.2 incompatible


greg-1-anderson:
Thanks!

Issue #1135, opened 2022-12-20
PHP 8.2 support


Sweetchuck:
According to the README.md non of the 3.x or 4.x branches support PHP 8.2,
but I think sooner or later PHP 8.2 should be supported, at least by the 4.x branch.

consolidation/robo:3.0.11 with PHP 8.2 the Exec task throws the following deprecation message:

> Deprecated: Use of "self" in callables is deprecated in [/.../src/Task/Base/Exec.php on line 66](https://github.com/consolidation/robo/blob/e1cc6411bf3634bea920bfc838b28ca8e612a749/src/Task/Base/Exec.php#L66)

There might be other problems as well.

chapterjason:
Has already been fixed in #1131 unfortunately you have to build it yourself locally.

```
git clone git@github.com:consolidation/robo.git
cd robo
git checkout 4.x
composer install
./robo phar:build
mv robo.phar /usr/local/bin/robo
```

Should do it.

Sweetchuck:
Thank you for you reply.

Yes, I saw the fix in the 4.x branch, but I think the Robo-3.x branch also should support PHP 8.2, because it is widely used and if the fix is easy like that.

The README.md should be clear about that which branches accept PHP-8.2 related merge requests.

I have several (~18) reusable Robo-3.x tasks, and I already make them compatible with PHP 8.2. (I noticed this issue during the update).
I would like to know that if I have to open a new branch and set the dependency to `consolidation/robo^4.0` in order to support PHP-8.2.


I might be wrong, but with the current situation a Drupal-9.x project which uses Robo-3.x can't upgrade to PHP 8.2. (I know that there is Drupal-10, but still)

greg-1-anderson:
I'll accept PRs that backport PHP 8.2 support to the 3.x branch.

Note that although test coverage of the core Robo classes is very good, there are a lot of standard Robo tasks that have little to no coverage. This means that things like "supports PHP 8.2" PRs often have to happen iteratively, because it's easy to miss needed changes in uncovered classes. PRs with more tests would be particularly welcome.

PR #1136, opened 2022-12-28
Issue #1135 - PHP 8.2 - Fix Use of "self" in callables is deprecated

Sweetchuck:
### Overview

This pull request:

Issue #1135 

- [x] Fixes a bug
- [ ] Adds a feature

### Summary

The problem is that the `convertDeprecationsToExceptions` is disabled by default in the `phpunit.dist.xml`. I have to enable it manually, but when I do so, a lot of deprecation messages come from the dependencies as well.


greg-1-anderson:
Thanks!

PR #1152, opened 2024-03-11
Ssh should implement PrintedInterface

### LINKED ISSUE #919: Duplicate output in non-interactive mode
### Steps to reproduce
I am using an `Ssh` task with a global `stopOnFail`, with Robo 1.4.11 called in a non-interactive terminal.

Here is a basic example that reproduces the issue:
```php
  public function test() {
    $this->stopOnFail();

    $sshTask = $this->taskSshExec();
    $sshTask
      ->interactive(FALSE)
      ->exec('echo "hello"')
      ->exec('invalid_cmd')
      ->run();
  }
```

### Expected behavior
The "hello" output should only be printed once.

### Actual behavior
The output is printed three times:
```
 echo "hello" &&
 invalid_cmd'
hello
bash: invalid_cmd: command not found
 [Remote\Ssh]  hello
  Time 0.367s
 [Remote\Ssh]  Exit code 127  Time 0.367s
 [notice] Stopping on fail. Exiting....
 [error]  Exit Code: 127 
 [error]    in task Robo\Task\Remote\Ssh 

  hello
```

I think one duplicate is caused by the global `stopOnFail` and another one is caused by the SSH class.

This does not happen in interactive mode.

### System Configuration
PHP 7.2.24 on Linux Mint 19.3.

prudloff-insite:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
`Ssh` uses `ExecTrait` so it has the `getPrinted()` method. It would make sense for this class to implement `PrintedInterface`.

### Description
Fixes https://github.com/consolidation/Robo/issues/919.
The logic in `Result::printError()` does not work correctly because it does not known that `$task->getPrinted()` is available so it always reprints the error even if already printed.

### DIFF
--- a/src/Task/Remote/Ssh.php
+++ b/src/Task/Remote/Ssh.php
@@ -3,6 +3,7 @@
 namespace Robo\Task\Remote;
 
 use Robo\Contract\CommandInterface;
+use Robo\Contract\PrintedInterface;
 use Robo\Exception\TaskException;
 use Robo\Task\BaseTask;
 use Robo\Contract\SimulatedInterface;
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks for the fix.

PR #1154, opened 2024-03-11
Clarify which branches are supported

Rudloff:
### Overview
This pull request:

- [ ] Fixes a bug
- [ ] Adds a feature

### Summary
Are all 4 branches supported?

### Description
It seems the 4.x branch was added to the table but the sentence was not updated.


greg-1-anderson:
That section is really old. Since PHP 7 is EOL, I don't think there's much motivation to continue supporting 1.x through 3.x, so I'd say likely that only 4.x is supported now.

Rudloff:
I updated the table based on your answer.

greg-1-anderson:
Thanks!

PR #1156, opened 2024-04-03
Update to symfony7

larowlan:
Hold off on merging this until its dependencies are merged into their mainline.

### DIFF
--- a/src/Common/IO.php
+++ b/src/Common/IO.php
@@ -54,14 +54,15 @@ trait IO
         return $this;
     }
 
-    public function setInput(InputInterface $input)
+    public function setInput(InputInterface $input): void
     {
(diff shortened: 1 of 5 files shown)


src/Common/IO.php
```diff
@@ -54,14 +54,12 @@ public function restoreState(InputInterface $input = null, OutputInterface $outp
...
 
-    public function setInput(InputInterface $input)
+    public function setInput(InputInterface $input): void
     {
         if ($input != $this->input) {
             $this->io = null;
         }
         $this->input = $input;
-
-        return $this;
```
greg-1-anderson, line 64: This is a b/c breaking change. Is it necessary?

greg-1-anderson:
Continued in #1158

PR #1169, opened 2024-10-21
Don't require exact event-dispatcher version

prudloff-insite:
I am assuming this is a typo.

### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Allow installing any 6.x version of symfony/event-dispatcher.

### Description
Drupal 10.3 requires event-dispatcher 6.4 so we can't install Robo 5.


greg-1-anderson:
Yep, looks like a typo -- thanks for catching that.

PR #1172, opened 2025-02-12
Nullable types must be explicit in 4.x.

ghost:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Nullable types must be explicit in 4.x branch to provide compatibility with PHP 8.4.

### DIFF
--- a/src/Common/IO.php
+++ b/src/Common/IO.php
@@ -45,7 +45,7 @@ trait IO
     }
 
     // This should typically only be called by State::restore()
-    public function restoreState(InputInterface $input = null, OutputInterface $output = null, SymfonyStyle $io = null)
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
The 5.x branch is already compatible with PHP 8.4.  If we merged this here, I'm not sure that this alone would make the 4.x branch compatible. I'm not sure that the 4.x branch can be made compatible in a backwards-compatible way. You should upgrade to the 5.x branch.

ghost:
I have to use 4.x because I use Drupal 10.4 on PHP 8.4 and I can't use 5.x version there because of dependency restrictions. I just added diff of this PR as a patch and now I don't see any deprecation messages. It's enough for my goals, anyway thanks for the response!

Issue #1174, opened 2025-04-16
Deprecation issues, "Error - Process Helper" &tc


MarkRS-UK:
Is this project dead?  Owner (@greg-1-anderson) still active on GitHub but not here.  Last update nearly two years ago.  Mutiple versions in process but not released.  Just wondering.

Having just updated to Fedora 42, when running Robo, even just looking for its version (which it shows me I have 4.0.4) gives the problems listed below.

It does produce the result, as does running one of my own robo files, but...

```
PHP Deprecated:  PHP Startup: session.sid_length INI setting is deprecated in Unknown on line 0

Deprecated: PHP Startup: session.sid_length INI setting is deprecated in Unknown on line 0
PHP Deprecated:  PHP Startup: session.sid_bits_per_character INI setting is deprecated in Unknown on line 0

Deprecated: PHP Startup: session.sid_bits_per_character INI setting is deprecated in Unknown on line 0
Xdebug: [Log Files] File '/var/log/httpd/xdebug_log' could not be opened.
Xdebug: [Step Debug] Could not connect to debugging client. Tried: localhost:9003 (through xdebug.client_host/xdebug.client_port).
PHP Deprecated:  Robo\Common\IO::restoreState(): Implicitly marking parameter $input as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/bin/robo/src/Common/IO.php on line 48

Deprecated: Robo\Common\IO::restoreState(): Implicitly marking parameter $input as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/bin/robo/src/Common/IO.php on line 48
PHP Deprecated:  Robo\Common\IO::restoreState(): Implicitly marking parameter $output as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/bin/robo/src/Common/IO.php on line 48

Deprecated: Robo\Common\IO::restoreState(): Implicitly marking parameter $output as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/bin/robo/src/Common/IO.php on line 48
PHP Deprecated:  Robo\Common\IO::restoreState(): Implicitly marking parameter $io as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/bin/robo/src/Common/IO.php on line 48

Deprecated: Robo\Common\IO::restoreState(): Implicitly marking parameter $io as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/bin/robo/src/Common/IO.php on line 48
[...]

PR #1176, opened 2025-07-23
fix deprecated errors: update dependencies in composer.json

koppieesq:
### Overview

Addresses a litany of warnings that occur when a user has a newer version of PHP, but installs older versions of the required libraries.  This is the same series of warnings reported in #1174.

This PR doesn't fix the helper libraries to specific versions.  Rather, it sets a "floor," allowing other projects (ie. Drupal) to upgrade those packages in the future.

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Short overview of what changed.

### Description
Any additional information.


koppieesq:
Lots of failing tests.  Yikes!  I'm reverting this PR to "draft" mode while I work on it some more.

koppieesq:
_All_ the tests are failing, for an unrelated reason.  I'm opening a new PR to address that: #1177

PR #1184, opened 2025-12-08
Fix deprecation issue with Symfony Console

rsebboc:
### Overview
This pull request make sure users don't get a deprecated message if they are using a newer version of Symfony Console:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary

Added a check to see if the new `addCommand` method exists, otherwise fallback to the `add` method

### Description
The `add` method will be removed in symfony/console 8.0

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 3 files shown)


pixelbrackets:
> Added a check to see if the new addCommand method exists, otherwise fallback to the add method

God catch, Symfony 6 does not have a addCommand method yet, Symfony 7 triggers the deprecation warning to use it tough.

The CI backwards check is red for some other reason, the code is fine.

rsebboc:
CI Fixed in #1185

PR #1185, opened 2026-01-12
Fix: Add Symfony Console 7.4+ addCommand() compatibility

laetus007:
### Overview
This pull request:

- [x] Fixes a bug
- [ ] Adds a feature

### Summary
Proposed Solution:
Added a compatibility wrapper addCommand() in Robo\Application that:

- Calls parent::addCommand() if available (Symfony 7.4+)
- Falls back to parent::add() for Symfony 6
- Override add() in Robo\Application to redirect to addCommand()

Update all internal calls to use addCommand() instead of add()

This ensures Robo works without deprecation warnings on Symfony 7.4+ and remains compatible with Symfony 6.

### Description

When using Robo with Symfony Console 7.4+, a deprecation warning is triggered:
```
ERROR: Since symfony/console 7.4: The "Symfony\Component\Console\Application::add()" method is deprecated and will be removed in Symfony 8.0, use "Symfony\Component\Console\Application::addCommand()" instead.
```

This occurs because Robo and its consumers call add() to register commands, but Symfony 7.4+ expects addCommand() instead. However, Symfony 6 does not have addCommand(), so a compatibility layer is needed.

***Note: We heavily use robo to orchestrate our local development and CI/CD for a single Drupal codebase using a custom upstream in Pantheon for over 200+ sites.  We recently updated to Drupal 11.3 and we started getting a number of errors when running the custom robo commands.  AI was used in the analysis of this issue and the solution proposed.  We have tested against our own codebase and validated the errors no longer appear for us.  I'll watch this to see if there are any test failures and will try to gracefully address, if possible and as time permits.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,85 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "356d965c2355967efaa4c71b48fe813e",
+    "content-hash": "557cebdc403e62c68a828bc13114b10e",
     "packages": [
+        {
+            "name": "composer/semver",
+            "version": "3.4.4",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/composer/semver.git",
+                "reference": "198166618906cb2de69b95d7d47e5fa8aa1b2b95"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/composer/semver/zipball/198166618906cb2de69b95d7d47e5fa8aa1b2b95",
+                "reference": "198166618906cb2de69b95d7d47e5fa8aa1b2b95",
+                "shasum": ""
+            },
+            "require": {
+                "php": "^5.3.2 || ^7.0 || ^8.0"
+            },
+            "require-dev": {
+                "phpstan/phpstan": "^1.11",
+                "symfony/phpunit-bridge": "^3 || ^7"
+            },
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-main": "3.x-dev"
+                }
+            },
+            "autoload": {
+                "psr-4": {
+                    "Composer\\Semver\\": "src"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "Nils Adermann",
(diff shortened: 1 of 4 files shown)


pixelbrackets:
I tested this in issue [1183](https://github.com/consolidation/robo/issues/1183), works like a charm. All breaking messages gone. A merge would be appreciated since this affects every run of a robo script.

vworldat:
@greg-1-anderson any chance to get this merged and released soon? It's a real bummer ATM 🫤

apotek (approved):
This looks complete, and I like the compatibility wrapper with a clear instruction as to how to fix the invalid argument. 

This solves a current issue with Robo. I see no reason to hold it back. Thank you [aetus007](https://github.com/laetus007)

Update: Tested on several projects using Robo and works flawlessly.

andypost:
it looks like better to add the shim to annotated command repo so robo and config and drush can reuse it

faced working on Symfon7 8+ compatibility where the `add()` method is removed https://github.com/consolidation/robo/pull/1187

laetus007:
Yes - I agree that this should not be merged here and it is more appropriate to address - https://github.com/search?q=repo%3Aconsolidation%2Fannotated-command+-%3Eadd%28&type=code

What is best practice in this scenario?  Convert this to draft while the issue is reported and addressed in the annotated-command?

jfremer:
Please note that this patch makes robo incompatible with Symfony <7.4.  There also needs to be an `::add()` fallback on line ~561 of Robo.php.  Running this patch in Symfony 6.x yields the following error on any drush command:

`Fatal error: Uncaught Error: Call to undefined method Drush\Application::addCommand() in /var/www/html/vendor/consolidation/robo/src/Robo.php:561`

tlo450:
What is the latest here? I use the `consolidation/robo` package in my site now, and I'm seeing this deprecation error when attempting to upgrade to Drupal 11.3.x.

I see I can just use the changes from this PR [as a patch](https://github.com/consolidation/robo/issues/1183#issuecomment-3851525535) , but are we really expecting that many changes to `composer.lock` here? Something just doesn't seem right to me... I don't like the idea of pulling in a patch that has that many changes to a composer lock file.

FlorentTorregrosa:
Hi,

Encountered the same PHP error.

Thanks for the MR it fixes the issue.

Here is a patch to apply with Composer to avoid using a moving target URL.

[1185.patch](https://github.com/user-attachments/files/32622363/1185.patch)

PR #1187, opened 2026-03-19
Allow Symfony 8

andypost:
Widen symfony/* constraints to allow ^8.

No code changes needed — verified that the APIs used by this package are unchanged in Symfony 8.

Note: check if Application::add() is used — it was removed in SF8, replaced by addCommand().

Ref: https://github.com/drush-ops/drush/issues/6536

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,12 +31,12 @@
         "consolidation/self-update": "*",
         "league/container": "^3.3.1 || ^4.0",
         "phpowermove/docblock": "^4.0",
-        "symfony/console": "^6 || ^7",
-        "symfony/event-dispatcher": "^6 || ^7",
-        "symfony/filesystem": "^6 || ^7",
-        "symfony/finder": "^6 || ^7",
-        "symfony/process": "^6 || ^7",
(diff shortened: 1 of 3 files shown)


andypost:
@weitzman please take a look as the removed `add()` method probably needs major release and cut off PHP 7

greg-1-anderson:
Do any of these tests run with Symfony 7 now? We'll want to either add that, or drop Symfony 7 support and make a new major release. I think it's safe to add the return typehint and still override an older version that has no return typehint. Might need a returntypemaychange annotation.

I don't specifically know about Applicatoin::add(), but with all of the tests green, looks like we might not need it.

Davidshb:
Hello @andypost, Do you know when symfony 8 will be allowed ?

andypost:
It's already but what for?

Davidshb:
@andypost I am doing an upgrade to symfony 8. I use this package [edgedesign/phpqa](https://packagist.org/packages/edgedesign/phpqa) which have a dependancy to this package but consolidation/robo doesn't support symfony 8 on their lastest version so now the package use 0.5 version which doesn't work

